PR #9, opened 2025-03-17
Update `examples/simple_rag/README.md` to verify the installation of `lxml`

yczhang-nv:
`lxml` might be missing from users' local environment if they don't do the whole `uv install` of `AgentIQ` (i.e. installing `AgentIQ` via pypi package). Add an additional step to confirm that `lxml` is installed before moving forward with the `milvus` setup.


yczhang-nv:
/merge

> This may not be needed given users will use installation from source to run examples but I think it does not hurt to have it

Yes, users installing from source won't have this issue. Just trying to reduce the confusion for those who are running the examples  based on `pypi` packages.

PR #10, opened 2025-03-17
Use a separate README for pypi

AnuradhaKaruppiah:
Change for testing


pypi.md
dagardner-nv, line 1: Does this need to be at the root of the repo, or can it be moved into the `ci` dir?

dagardner-nv (approved):
LGTM, just one question

pypi.md
AnuradhaKaruppiah: I will move it to the ci dir

Issue #21, opened 2025-03-19
Require github ssh key to clone sub-module


quanvuhust:
When I run the command ```git submodule update --init --recursive```
A error raise:
```
Submodule 'external/agentiq-opensource-ui' (git@github.com:NVIDIA/AgentIQ-UI.git) registered for path 'external/agentiq-opensource-ui'
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```

AnuradhaKaruppiah:
Hi @quanvuhust,
You need to add your ssh key to github. Please use the instructions here -  https://github.com/settings/keys

quanvuhust:
I fix this bug by modifying the .gitmodules file to:

![Image](https://github.com/user-attachments/assets/273bcef5-9da8-45ff-ae79-23be54c5c2bd)

PR #53, opened 2025-04-01
Switch UI submodule to https

AnuradhaKaruppiah:
This is to allow pulling down the submodule without the need for SSH keys.


mdemoret-nv (approved):

mdemoret-nv:
Addresses #21

PR #75, opened 2025-04-03
Implement ReWOO Agent

yczhang-nv:
## Description

This PR adds an implementation of a ReWOO agent.

- A ReWOO agent works as follows:

  1. The `planner` node generates the entire chain of tools ("plan") in one go.
  2. The `executor` node processes the "plan" by iterating through each step and invoking the specified tools. Each tool’s output replaces the corresponding placeholder in the plan, and that updated plan is passed to the subsequent tool.
  3. The `solver` node generates the final results based on the complete "plan" and the tool outputs from the previous steps.

- Compared to the ReAct agent, the ReWOO pattern eliminates the need to include the system prompt and all previous steps for every reasoning iteration, thereby reducing token usage and boosting performance.

- This PR also introduces a minor refactor to the `BaseAgent` interface to accommodate the ReWOO pattern, which utilizes three nodes instead of the two nodes used by ReAct.

Closes https://jirasw.nvidia.com/browse/AIQ-849


mdemoret-nv (changes requested):
Looks good but we need unit tests for this

mdemoret-nv (approved):
Great addition!

PR #98, opened 2025-04-08
Add function tracking decorator and update IntermediateStep

dnandakumar-nv:
Introduce a `track_function` decorator to wrap sync, async, and generator functions for tracking execution spans. Added new SPAN_START, SPAN_CHUNK, and SPAN_END types to IntermediateStep. Updated tests to ensure tracking functionality is properly validated.

## Description
Feature: Function Tracking Decorator

This PR adds a straightforward decorator, `@track_function`, that makes it easy to log details about function calls. It works on synchronous and asynchronous functions, generators (sync and async), and class methods—all with a consistent interface.

Key Points:
- Versatile Support: Works with regular functions, async functions, generators, and even class methods.
- Optional Metadata: You can add extra context by passing a dictionary to the decorator. Metadata keys are validated to ensure they’re strings.
- Data Serialization: Automatically serializes input arguments and outputs (or each yield) into a JSON-friendly format. It even handles Pydantic models by calling model_dump() if needed.
- Intermediate Event Logging: The decorator sends tracking events to a reactive stream in three stages:
-- SPAN_START: When the function starts.
-- SPAN_CHUNK: For each yielded item (for generators).
-- SPAN_END: When the function returns (or ends in generators).

Usage Example:

```python
from aiq.profiler.decorators.function_tracking import track_function

@track_function(metadata={"action": "compute", "source": "api"})
def add_numbers(a, b):
    return a + b
```
Just apply `@track_function` to any function (or method) where you need a bit of extra insight into the inputs and outputs. Subscribe to the reactive stream to see the published events and get a better sense of the function’s flow.


src/aiq/profiler/decroators/function_tracking.py
```diff
@@ -0,0 +1,252 @@
...
+    Decorator that can wrap any type of function (sync, async, generator,
+    async generator) and executes "tracking logic" around it.
+
+    - If the function is async, it will be wrapped in an async function.
+    - If the function is a generator, it will be wrapped in a generator function.
+    - If the function is an async generator, it will be wrapped in an async generator function.
+    - If the function is sync, it will be wrapped in a sync function.
+    """
+
+    step_manager: IntermediateStepManager = AIQContext.get().intermediate_step_manager
```
mdemoret-nv: This is too early to be getting the step manager from the context. You need to get it in the decorated function. Its possible the step manager isnt set yet.

mdemoret-nv (approved):

src/aiq/profiler/decroators/function_tracking.py
dnandakumar-nv: done

PR #108, opened 2025-04-09
Add custom function entry point to local workflows

dnandakumar-nv:
## Description
This PR adds support for setting custom entrypoints when running `aiq run`  on local workflows. This was needed when multiple functions were defined in a configuration file, but functions that were not the `workflow` function could not be invoked via command line  without changing the config. 

Now, to run with a custom entrypoint when using `aiq run`, a simple addition to the command can enable it as follows:

```bash
aiq run examples/simple_calculator/configs/config.yml --input "4 and 5" --entry_fn "calculator_inequality"
```

The command will then send the input directly to the `calculator_inequality` function. 

### DIFF
--- a/src/aiq/runtime/session.py
+++ b/src/aiq/runtime/session.py
@@ -103,7 +103,7 @@ class AIQSessionManager:
                 self._context_state.user_input_callback.reset(token_user_input)
 
     @asynccontextmanager
-    async def run(self, message):
+    async def run(self, message, entry_fn: str | None = None):
         """
         Start a workflow run
         """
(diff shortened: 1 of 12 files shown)


AnuradhaKaruppiah:
LGTM. Needs to documented in some README.

dnandakumar-nv:
> LGTM. Needs to documented in some README.

Done

AnuradhaKaruppiah (approved):

src/aiq/runtime/session.py
```diff
@@ -103,7 +103,7 @@ async def session(self,
                 self._context_state.user_input_callback.reset(token_user_input)
 
     @asynccontextmanager
-    async def run(self, message):
+    async def run(self, message, entry_fn: str | None = None):
```
mdemoret-nv, line 106: We dont want to add this here. It adds an extra burden to every `run` call. There is another method in the workflow builder that will create a new workflow from a different entrypoint. Use that instead. Look at the fastapi front end for more info

mdemoret-nv (approved):

PR #130, opened 2025-04-14
Support user-defined HTTP request metadata in workflow tools.

ericevans-nv:
## Description

This pull request introduces the ability to extract and utilize user-defined attributes from incoming HTTP requests. These attributes can be accessed via the UserManagerAttributes class within the AIQContext and used in workflow tools.


src/aiq/builder/function_base.py
```diff
@@ -368,5 +368,8 @@ def _convert_input(self, value: typing.Any):
 
             return getattr(value, first_key)
 
+        if value is None:
```
mdemoret-nv, line 372: Add a comment for why this is needed

src/aiq/data_models/api_server.py
```diff
@@ -37,6 +37,24 @@ class Message(BaseModel):
...
+    method: str | None = None
+    url_path: str | None = None
+    url_port: int | None = None
+    url_scheme: str | None = None
+    headers: typing.Any | None = None
+    query_params: typing.Any | None = None
+    path_params: dict[str, str] | None = None
+    client_host: str | None = None
+    client_port: int | None = None
+    cookies: dict[str, str] | None = None
```
mdemoret-nv: Need descriptions on each of these properties.

src/aiq/runtime/user_manager.py
```diff
@@ -0,0 +1,119 @@
...
+from starlette.datastructures import QueryParams
+
+from aiq.data_models.api_server import Request
+
+
+class UserManagerBase:
+    pass
+
+
+class UserAttributes:
```
mdemoret-nv: These feel like `RequestAttributes` and not `UserAttributes`

src/aiq/front_ends/fastapi/fastapi_front_end_plugin_worker.py
```diff
@@ -136,6 +137,21 @@ def set_cors_config(self, aiq_app: FastAPI) -> None:
             **cors_kwargs,
         )
 
+    def set_user_attributes(self, request: Request, session_manager: AIQSessionManager) -> None:
+        """
+        Extracts and sets user request attributes from an HTTP request.
+        """
+        session_manager.context.user_attributes._request.method = request.method
```
mdemoret-nv: We do not want to set private methods like this. This should get set in the `AIQSessionManager.session()` function. Add a new argument `request` to this method.

mdemoret-nv (changes requested):

src/aiq/runtime/user_manager.py
ericevans-nv: They do and I initially thought of that pattern as well, but since we are exposing this public interface to users to use across workflow tools we want tight control over what data they can access and how. The UserAttributes class is a Facade that will be extended to encapsulate other data classes users might need in the future.  Classes such as OAuth2.0 Attributes, API_Key Attributes etc...

src/aiq/builder/context.py
```diff
@@ -62,6 +63,7 @@ class AIQContextState(metaclass=Singleton):
     def __init__(self):
         self.input_message: ContextVar[typing.Any] = ContextVar("input_message", default=None)
         self.user_manager: ContextVar[typing.Any] = ContextVar("user_manager", default=None)
+        self.user_attributes: ContextVar[UserAttributes] = ContextVar("user_attributes", default=UserAttributes())
```
mdemoret-nv: Can we just call this metadata?

src/aiq/builder/context.py
```diff
@@ -110,6 +112,20 @@ def user_manager(self):
         """
         return self._context_state.user_manager.get()
 
+    @property
+    def user_attributes(self):
```
mdemoret-nv: Call this metadata too

src/aiq/runtime/user_manager.py
mdemoret-nv: I was referring more to the name. Should we call this `RequestAttributes` instead of `UserAttributes`?

mdemoret-nv (approved):

PR #134, opened 2025-04-15
Rename Otelcollector to otlp

divyanshiynv:
## Description
This PR is to rename 'otelcollector' object to otlp which is more relevant for the use case. This object can be used with any telemetry exporter like datadog, zipkin etc. OtelCollector has a different meaning and functionality and that needs its own configuration file to be specified. Using otlp directly is more simpler.


mdemoret-nv (commented):
Is there any way to change the name but maintain backwards compatibility until the next major release?

PR #135, opened 2025-04-17
Add W&B Weave tracing

ayulockin:
## Description

Closes

This PR adds support for exporting the OTEL spans to Weights and Biases [Weave](https://weave-docs.wandb.ai) platform for better interactivity and off-the-shelf LLMOps life-cycle.

As a user all you need to do is update the `config.yaml` with the following:

```
general:
  use_uvloop: true
  telemetry:
    tracing:
      weave:
        _type: weave
        entity: <your-wandb-entity>
        project: my-awesome-genai-app
        log_otel_only: true
```

The `log_otel_only` is a special config which allows you to:

- If `true`, export OTEL spans generated by AgentIQ to Weave.
- If `false`, do `weave.init(..)` under the hood to initialize a Weave client. Doing so will allow you to trace the underlying frameworks, llm vendors, etc., as we have [auto patched integrations](https://weave-docs.wandb.ai/guides/integrations/) with them. This give a more granular trace captured automatically.

An example OTEL trace timeline exported to Weave:

image

View more traces in this project: https://wandb.ai/ayut/simple/weave/traces

This is a draft PR:

- [x] Refactor the integration when https://github.com/wandb/weave/pull/4143 is shipped.
- [ ] Add documentation and example.

cc: @dnandakumar-nv


src/aiq/observability/register.py
```diff
@@ -26,6 +33,81 @@
...
+            return None
+
+    api_key = get_wandb_api_key(config.api_key)
+    if not api_key:
+        logger.error("W&B API key not found. Please provide it in the config or set WANDB_API_KEY environment variable.")
+        yield NoOpSpanExporter()
+        return
+
+    try:
+        auth = base64.b64encode(f"api:{api_key}".encode()).decode()
```
mdemoret-nv: Out of curiosity, whats the purpose for the encode/decode?

mdemoret-nv (approved):

mdemoret-nv:
@ayulockin Thanks for your PR. For outside contributors, we require the DCO signoff. You can find more info [here](https://github.com/NVIDIA/AgentIQ/pull/135/checks?check_run_id=40958500038).

/ok to test

mdemoret-nv:
Thanks for the additional updates. Let me take a look. I want to test this out with a weave instance which I am still trying to get access to.

pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
   "ragas~=0.2.14",
   "rich~=13.9",
   "uvicorn[standard]~=0.32.0",
+  "weave>=0.51.43",
```
ayulockin: I have added weave as a dependency for now so that testing this PR would be easier. 

If this is okay, we can leave it as it is. The reason I made this a requirement is because of changes made to `async_otel_listner.py` file.

src/aiq/observability/register.py
ayulockin: not needed anymore.

pyproject.toml
mdemoret-nv: One option we have is to create a new sub-package which would be `agentiq-weave` to contain all of the weave integration pieces. It would only be this singular plugin for now but it would solve the dependency issue.

Generally, we are trying to keep the core library dependencies light. What do you think of this option?

ayulockin: Oh yes for sure. I can do that. :)

ayulockin:
I think I messed up rebasing. I am closing this PR to quickly open a new one.

PR #148, opened 2025-04-25
Support Python `3.11` for AgentIQ

yczhang-nv:
## Description

- Downgrades Python requirement from `3.12` to `3.11` across all packages
- Adds compatibility layer for `typing.override` decorator `type_utils.py`
- Replaces all `@typing.override` instances with the compatible version


README.md
```diff
@@ -86,6 +87,12 @@ Before you begin using AgentIQ, ensure that you meet the following software prer
     uv venv --seed .venv
     source .venv/bin/activate
     ```
+    Make sure the environment is built with Python version `3.11` or above. If you have multiple Python versions installed,
+    you can specify the desired version using the `--python` flag. For example, to use Python 3.11:
+    ```bash
+    uv venv --seed .venv --python $(which python3.11)
```
mdemoret-nv: I would just use `--python 3.11` here which is easier

src/aiq/utils/type_utils.py
```diff
@@ -45,6 +46,23 @@ def is_valid_json(string):
         return False
 
 
+# Create a compatibility layer for typing.override
+# In Python >= 3.12, use the built-in typing.override
+# Otherwise, create a no-op decorator
+def _create_override_decorator():
```
mdemoret-nv: Is there a reason why this isnt just called "override"? Would be better to make it public and have a proper docstring on it.

mdemoret-nv (approved):

src/aiq/utils/type_utils.py
yczhang-nv: Seems like implementing this compatibility layer as a customized decorator like this does not work: 
```python
def override(func):
    if sys.version_info >= (3, 12):
        from typing import override as _override
        return _override(func)
    return func
```
Dynamically choosing between the built-in decorator and a no-op at runtime prevents static analyzers from seeing the real typing.override symbol, so they won’t flag incorrect overrides. To make it simple, I'll go with just use a if/else with version check.

PR #157, opened 2025-04-29
Add support for environment variable interpolation in config files

mpenn:
## Description

Closes: https://github.com/NVIDIA/AgentIQ/issues/156

This PR adds support for environment variable interpolation in YAML configuration files using the format `${VAR:-default_value}`. This allows developers to:

1. Reference environment variables in configuration
2. Provide default values if the environment variable is not set
3. Use empty strings as default values if needed

To illustrate this concept, an example from the `llms` section of the configuration file is provided below.

```yaml
llms:
  nim_llm:
    _type: nim
    base_url: ${NIM_BASE_URL:-"http://default.com"}  # Optional with default value
    api_key: ${NIM_API_KEY}  # Will use empty string if `NIM_API_KEY` not set
    model_name: ${MODEL_NAME:-}  # Will use empty string if `MODEL_NAME` not set
    temperature: 0.0
```

The environment variable interpolation process follow the rules enumerated below.

- `${VAR}` - Uses the value of environment variable `VAR`, or empty string if not set
- `${VAR:-default}` - Uses the value of environment variable `VAR`, or `default` if not set
- `${VAR:-}` - Uses the value of environment variable `VAR`, or empty string if not set


src/aiq/utils/io/yaml_tools.py
```diff
@@ -14,37 +14,128 @@
...
+    if not isinstance(value, str):
+        return value
+
+    def replace_var(match):
+        var_name = match.group(1)
+        default_value = match.group(2) if match.group(2) else ""
+        return os.environ.get(var_name, default_value)
+
+    pattern = r'\${([^:}]+)(?::-([^}]*))?}'
+    return re.sub(pattern, replace_var, value)
```
mdemoret-nv: What about using something like [`expandvars`](https://pypi.org/project/expandvars/) so we dont have to maintain this logic ourselves? There are more edge cases which need to be dealt with (escape sequences, default values with quotes, etc).

src/aiq/utils/io/yaml_tools.py
```diff
@@ -14,37 +14,128 @@
...
+        dict: The processed configuration dictionary.
+    """
 
     # Read YAML file
     with open(config_path, 'r', encoding="utf-8") as stream:
         config_data = yaml.safe_load(stream)
 
+    # Process the configuration to interpolate variables
+    config_data = _process_config(config_data)
+    assert isinstance(config_data, dict)
```
mdemoret-nv: Loading the file as a YAML and then processing the environment variables will limit environment variables to only work in strings.

Instead, what if we processed the environment variables on the string, and then loaded the YAML file. We could then have things like:
```yaml
functions:
   ${FUNCTION_LIST}
```

I also worry about data types:

```yaml
should_be_an_int: ${MAX_THREADS}
```

Would load into the dictionary as `dict[str, str]` even though it needs to be an integer.

src/aiq/utils/io/yaml_tools.py
```diff
@@ -14,37 +14,128 @@
...
+
+    Returns:
+        dict: The processed configuration dictionary.
+    """
 
     stream = io.StringIO(config)
     stream.seek(0)
 
-    return yaml.safe_load(stream)
+    config_data = yaml.safe_load(stream)
```
mdemoret-nv: We need some way to display the bad YAML to the user if it fails to parse correctly to help with interpolation errors

mdemoret-nv (changes requested):
Good functionality, but I fear we are missing edge cases

src/aiq/utils/io/yaml_tools.py
mpenn: Regarding your second comment here, because we are using Pydantic models to translate these YAML files into aiq configuration objects, data types should be cast to the appropriate type during workflow instantiation.

src/aiq/utils/io/yaml_tools.py
```diff
@@ -14,37 +14,128 @@
...
+        str | int | float | bool | None: The value with variables interpolated.
+    """
+
+    if not isinstance(value, str):
+        return value
+
+    return expandvars.expandvars(value)
+
+
+def _process_config(
```
mdemoret-nv: Is this function still used?

mdemoret-nv (approved):
Looks good

PR #167, opened 2025-04-30
Make `opentelemetry` and `phoenix` as optional dependencies

yczhang-nv:
## Description

This PR makes `opentelemetry` and `phoenix` as optional dependencies for Omniverse integration.

- Moved the telemetry dependencies from AIQ core dependency to a separate group of optional dependency, which can be installed by `uv pip install -e '.[telemetry]'`.
- Created `src/aiq/utils/optional_imports.py` to include optional import functionalities and dummy implementations. This module can be reused if we want to add more optional dependencies in the future.
- When the configuration file contains `tracing` section but the telemetry dependencies are not installed, `TelemetryOptionalImportError` will be raised
- Updated files that imports the optional dependencies to gracefully handle the case when those dependencies are not installed
- Updated logging and documentation

Closes [AIQ-1064](https://jirasw.nvidia.com/browse/AIQ-1064)

- I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/AIQToolkit/blob/develop/docs/source/advanced/contributing.md).


src/aiq/builder/workflow_builder.py
```diff
@@ -59,8 +54,27 @@
 from aiq.memory.interfaces import MemoryEditor
 from aiq.profiler.decorators.framework_wrapper import chain_wrapped_build_fn
 from aiq.profiler.utils import detect_llm_frameworks_in_build_fn
+from aiq.utils.optional_imports import OptionalImportError
+from aiq.utils.optional_imports import try_import_opentelemetry
 from aiq.utils.type_utils import override
 
+# Try to import OpenTelemetry modules
+try:
```
mdemoret-nv, line 62: Can we perform these imports locally and only if an exporter is configured? If an exporter is configured and open telemetry is not installed, this should generate an error. Otherwise people will be confused why nothing is being exported

src/aiq/observability/register.py
```diff
@@ -56,10 +63,16 @@ class OtelCollectorTelemetryExporter(TelemetryExporterBaseConfig, name="otelcoll
 
 @register_telemetry_exporter(config_type=OtelCollectorTelemetryExporter)
 async def otel_telemetry_exporter(config: OtelCollectorTelemetryExporter, builder: Builder):
-
-    from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
-
-    yield OTLPSpanExporter(endpoint=config.endpoint)
+    """Create an OpenTelemetry telemetry exporter."""
```
mdemoret-nv, line 61: This should error if requested. I dont think we should support attempting to use telemetry if its not available

pyproject.toml
```diff
@@ -64,6 +61,11 @@ mem0ai = ["aiqtoolkit-mem0ai"]
 semantic-kernel = ["aiqtoolkit-semantic-kernel"]
 zep-cloud = ["aiqtoolkit-zep-cloud"]
 agno = ["aiqtoolkit-agno"]
+telemetry = [
+  "arize-phoenix~=6.1",
+  "opentelemetry-api~=1.2",
+  "opentelemetry-sdk~=1.3",
+]
```
mdemoret-nv, line 68: Can you put these in alphabetical order?

mdemoret-nv (approved):
Main improvement this PR needs is to localize the imports and to throw errors in invalid configurations (i.e. when users request telemetry but its not installed)

src/aiq/builder/workflow_builder.py
yczhang-nv: We need to use a `DummySpanExporter` here to define the `ConfiguredExporter` dataclass. The case an exporter is configured and open telemetry is not installed is handled locally.

PR #196, opened 2025-05-02
Store structure tool inputs

1ytic:
Store original tool inputs for TOOL_END event in LangchainProfilerHandler.


elvinagam (changes requested):
should this be closed?

PR #302, opened 2025-05-19
Support additional_instructions

### LINKED ISSUE #301: [BUG]: React agent does not use additional_instructions input
### Version

0.1.0

PyPi

### Describe the bug.

1. Expect additional_instructions to be appended to system prompt.   
2. Expect additional_instructions to be used when custom system is configured and a default system prompt is used.  

```shell

```

### Relevant log output

_No response_

### Other/Misc.

logic error in https://github.com/NVIDIA/AIQToolkit/blob/develop/src/aiq/agent/react_agent/register.py#L85

1. additional_instructions only examined when custom system_prompt defined. 
2. The variable _prompt_str is discarded.    

creating a PR to fix.  

### Code of Conduct

- [x] I have searched the [open bugs](https://github.com/NVIDIA/AIQToolkit/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

gfreeman-nvidia:
## Description

React agent supports additional_instructions configuration option.  
Add test cases for custom system prompt and adding additional_instructions.

Closes #301

### DIFF
--- a/src/aiq/agent/react_agent/register.py
+++ b/src/aiq/agent/react_agent/register.py
@@ -15,6 +15,7 @@
 
 import logging
 
+from langchain_core.prompts import ChatPromptTemplate
 from pydantic import Field
 
 from aiq.agent.base import AGENT_LOG_PREFIX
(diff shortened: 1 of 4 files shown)


src/aiq/agent/react_agent/register.py
```diff
@@ -15,6 +15,7 @@
 
 import logging
 
+from langchain_core.prompts import ChatPromptTemplate
```
dagardner-nv: langchain is an optional dependency for us, as such we need to be able to safely import this, without calling any of them methods if langchain is nost installed. 

Move the import to the `create_react_agent_prompt` method.

gfreeman-nvidia: I moved create_react_agent_prompt  to the agent.py file.  since the type was used as an output type hint in the signature, it couldn't be defined in the function body.   This will allow register.py to be importable without langchain.  moving to prompt.py added a circular dependency.

PR #307, opened 2025-05-21
MCP enhancements: improves server config and adds support for all transport types (stdio, streamable-http)

### LINKED ISSUE #304: [FEA]: Add support for stdio MCP Servers

New Feature

High

The current MCP client tool wrapper requires a MCP server that operates over http with SES. Many MCP servers only provide documentation for running the stdio variant. 

To alleviate this, the toolkit provides some documentation on using mcp_proxy to proxy requests from a stdio mcp server to a ses endpoint. Unfortunately, this documentation does not work for many MCP servers.

The default documentation creates a mcp_proxy run script that uses `mcp-proxy uvx run service-name`, but many MCP servers are not compatible with the rum command `uvx run`. Some require docker run commands, which are seemingly not compatible with mcp_proxy. 

As an example, the new github mcp server recommends a docker run command: 

https://github.com/github/github-mcp-server

It would be a lot more straightforward if the mcp_tool_wrapper could accept and interact with a stdio running service, or even better would be if you could drop in the VS code / Cursor MCP configuration verbatim somewhere into AIQ and have it "just work", as most of these servers provide that documentation and little else.

### Additional context

_No response_

### Code of Conduct

- [x] I have searched the [open feature requests](https://github.com/NVIDIA/AIQToolkit/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22%2Cimprovement%2Cenhancement) and have found no duplicates for this feature request

AnuradhaKaruppiah:
## Description

Closes #304, #314

This PR 
1. Allows MCP servers to be configured without wrapping each tool on the server as a separate static function. It does that by introducing a new `mcp_client` function. This function connects to the server, discovers tools and adds them as dynamic functions (`mcp_single_tool`) to the AIQ workflow builder.
2.  Adds support for local MCP services using stdio

Sample usage (see examples/simple_calculator/configs/config-mcp-date-stdio.yml) -
```
functions:
  mcp_time:
    _type: mcp_client
    server:
      transport: stdio
      command: "python"
      args: ["-m", "mcp_server_time", "--local-timezone=America/Los_Angeles"]
    # load 1 of 2 tools and override tool attributes
    tool_filter:
      get_current_time:
        alias: get_current_time_mcp_tool
        description: "Returns the current date and time from the MCP server"

  mcp_math:
    _type: mcp_client
    server:
      transport: sse
      url: "http://localhost:9901/sse"
    # load all tools and use the names and descriptions from the server
```


src/aiq/tool/mcp/mcp_client.py
```diff
@@ -83,49 +87,122 @@ def _generate_field(field_name: str, field_properties: dict[str, Any]) -> tuple:
...
+        Establish a session with an MCP server via stdio within an async context
+        """
+        server_params = StdioServerParameters(command=self.url, args=self._args or [], env=self._env)
+        async with stdio_client(server_params) as (read, write):
+            async with ClientSession(read, write) as session:
+                await session.initialize()
+                yield session
+
+
+class MCPBuilder(MCPBaseClient):
```
mdemoret-nv: We should get rid of `MCPBuilder` and just use `MCPBaseClient` in its place. It's current purpose is just to create clients based on the client type and then forward arguments to the client. The rest of the logic in this class can be put into `MCPBaseClient`

src/aiq/tool/mcp/mcp_client.py
```diff
@@ -83,49 +87,122 @@ def _generate_field(field_name: str, field_properties: dict[str, Any]) -> tuple:
...
 
 
-class MCPSSEClient:
+class MCPBaseClient(ABC):
+    """
+    Base client for creating a session and connecting to an MCP server
+    """
+
+    def __init__(self, url: str):
+        self.url = url
```
mdemoret-nv: Is there a need for `url` here? Its a bit odd that we use this for both the command and the url. Would be better to have them be separate properties so they dont get mixed up

src/aiq/tool/mcp/mcp_client.py
```diff
@@ -151,12 +228,12 @@ async def get_tool(self, tool_name: str) -> MCPToolClient:
         return tool
 
     async def call_tool(self, tool_name: str, tool_args: dict | None):
-        async with self.connect_to_sse_server() as session:
+        async with self.connect_to_server() as session:
```
mdemoret-nv: On line 225, `self._tools.get(tool_name)` should be `self._tools.get(tool_name, None)` to raise the `ValueError`

src/aiq/tool/mcp/mcp_tool.py
```diff
@@ -33,8 +33,12 @@ class MCPToolConfig(FunctionBaseConfig, name="mcp_tool_wrapper"):
     function.
     """
     # Add your custom configuration parameters here
-    url: HttpUrl = Field(description="The URL of the MCP server")
+    url: HttpUrl | None = Field(default=None, description="The URL of the MCP server (for SSE mode)")
     mcp_tool_name: str = Field(description="The name of the tool served by the MCP Server that you want to use")
+    client_type: str = Field(default="sse", description="The type of client to use ('sse' or 'stdio')")
+    command: str | None = Field(default=None, description="The command to run for stdio mode (e.g. 'mcp-server')")
+    args: list[str] | None = Field(default=None, description="Additional arguments for the stdio command")
+    env: dict[str, str] | None = Field(default=None, description="Environment variables to set for the stdio process")
```
mdemoret-nv: Should have validation that the stdio properties are exclusively set from the SSE properties.

mdemoret-nv (changes requested):
Could improve design a bit now that we have two client types by removing `MCPBuilder`

src/aiq/tool/mcp/mcp_client.py
AnuradhaKaruppiah: Agree, good suggestion

src/aiq/tool/mcp/mcp_client.py
AnuradhaKaruppiah: Ack

src/aiq/tool/mcp/mcp_tool.py
AnuradhaKaruppiah: Ack

src/aiq/tool/mcp/mcp_client.py
AnuradhaKaruppiah: Ack

PR #347, opened 2025-06-06
fix(phoenix): set project name when using phoenix telemetry exporter (#337)

### LINKED ISSUE #337: [BUG]: Phoenix telemetry exporter ignores project value specified in configuration
### Version

1.1.0

Source

### Describe the bug.

When configuring a Phoenix telemetry exporter that includes the project parameter in a workflow YAML file, the project value is ignored. Traces are always written to the Phoenix server's default project.

```shell
This issue can be reproduced by following the simple calculator tracing example: https://github.com/NVIDIA/AIQToolkit/blob/develop/examples/simple_calculator/src/aiq_simple_calculator/configs/config-tracing.yml

When running the example, the Phoenix server will only receive traces into the default project.
```

### Relevant log output

</pre></details>

### Other/Misc.

_No response_

### Code of Conduct

- [x] I have searched the [open bugs](https://github.com/NVIDIA/AIQToolkit/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

briancaffey:
This adds a fix for the Phoenix telemetry exporter. Currently it does not use the project name in the configuration. For Phoenix this needs to be set in the TraceProvider like so: 

```
provider = TracerProvider(resource=Resource(attributes={ResourceAttributes.PROJECT_NAME: trace_exporter_config.project}))
```

where `ResourceAttributes.PROJECT_NAME` is `openinference.project.name`.

This allows for traces to be sent to projects other than the default project: 

phoenix_projects

This also fixes a typo in the Langsmith telemetry exporter that allows for traces to go to the correct project. (header changed from`LANGSMITH_PROJECT` to `Langsmith-Project`)

Langsmith and Langfuse set their `project` differently: Langsmith uses a header value (`Langsmith-Project` and Langfuse uses per-project keys.

This could be problematic if other telemetry exporters also relied on `attributes` in the `TracerProvider` to set the project value. It would also not work if there were multiple `telemetry.tracing` configurations using phoenix defined in a config file.  So this might not be an ideal solution, but it should work for simple configurations where you want to use different projects for different configuration files with phoenix.

## Description

Closes #337 

### DIFF
--- a/src/aiq/builder/workflow_builder.py
+++ b/src/aiq/builder/workflow_builder.py
@@ -167,11 +167,24 @@ class WorkflowBuilder(Builder, AbstractAsyncContextManager):
 
             # pylint: disable=unused-variable,redefined-outer-name
             opentelemetry = try_import_opentelemetry()  # noqa: F841
+            from openinference.semconv.resource import ResourceAttributes
             from opentelemetry import trace
+            from opentelemetry.sdk.resources import Resource
             from opentelemetry.sdk.trace import TracerProvider
             from opentelemetry.sdk.trace.export import BatchSpanProcessor
 
+            from aiq.observability.register import PhoenixTelemetryExporter
+
+            # Create a default provider first
             provider = TracerProvider()
+
+            # Check if we have a phoenix telemetry exporter and use its project name
+            for key, trace_exporter_config in telemetry_config.tracing.items():
+                if isinstance(trace_exporter_config, PhoenixTelemetryExporter):
+                    provider = TracerProvider(resource=Resource(
+                        attributes={ResourceAttributes.PROJECT_NAME: trace_exporter_config.project}))
+                    break
+
             trace.set_tracer_provider(provider)
 
             for key, trace_exporter_config in telemetry_config.tracing.items():
--- a/src/aiq/observability/register.py
+++ b/src/aiq/observability/register.py
@@ -103,7 +103,7 @@ async def langsmith_telemetry_exporter(config: LangsmithTelemetryExporter, build
     if not api_key:
         raise ValueError("API key is required for langsmith")
 
-    headers = {"x-api-key": api_key, "LANGSMITH_PROJECT": config.project}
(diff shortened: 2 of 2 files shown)


src/aiq/builder/workflow_builder.py
```diff
@@ -167,11 +167,24 @@ async def __aenter__(self):
...
 
+            from aiq.observability.register import PhoenixTelemetryExporter
+
+            # Create a default provider first
             provider = TracerProvider()
+
+            # Check if we have a phoenix telemetry exporter and use its project name
+            for key, trace_exporter_config in telemetry_config.tracing.items():
+                if isinstance(trace_exporter_config, PhoenixTelemetryExporter):
+                    provider = TracerProvider(resource=Resource(
```
mdemoret-nv, line 184: Is it possible to alter the attributes after the provider has been created?

mdemoret-nv (approved):
Approving this even though it breaks the plugin encapsulation. We have plans to re-architect the tracing system to be less dependent on OTel in the future which will clean this up. Merging this will solve the problem in the short term

src/aiq/builder/workflow_builder.py
briancaffey: @mdemoret-nv I can try that. I think the `otelcollector` also needs this fix for the project name. I'll check to see if I can change the attribute for each tracing config (project name)

briancaffey: I don't think this is possible. From `opentelemetry-python` docs: 

**opentelemetry.trace.set_tracer_provider(tracer_provider)[[source]](https://opentelemetry-python.readthedocs.io/en/latest/_modules/opentelemetry/trace.html#set_tracer_provider)[](https://opentelemetry-python.readthedocs.io/en/latest/api/trace.html#opentelemetry.trace.set_tracer_provider)**
> Sets the current global [TracerProvider](https://opentelemetry-python.readthedocs.io/en/latest/api/trace.html#opentelemetry.trace.TracerProvider) object.

> This can only be done once, a warning will be logged if any further attempt is made.

PR #355, opened 2025-06-10
Provide a config to pass the complete dataset entry as an EvalInputItem field to evaluators

AnuradhaKaruppiah:
## Description

Closes

### DIFF
--- a/src/aiq/eval/dataset_handler/dataset_handler.py
+++ b/src/aiq/eval/dataset_handler/dataset_handler.py
@@ -81,6 +81,7 @@ class DatasetHandler:
                 output_obj=row.get(self.generated_answer_key, "") if structured else "",
                 trajectory=row.get(self.trajectory_key, []) if structured else [],
                 expected_trajectory=row.get(self.expected_trajectory_key, []) if structured else [],
+                full_dataset_entry=row.to_dict() if self.dataset_config.pass_full_entry else None,
             )
 
         # if input dataframe is empty return an empty list
--- a/docs/source/reference/evaluate.md
+++ b/docs/source/reference/evaluate.md
@@ -87,6 +87,20 @@ eval:
         disable: true
 ```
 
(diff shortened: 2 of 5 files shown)


src/aiq/eval/dataset_handler/dataset_handler.py
```diff
@@ -81,6 +81,7 @@ def create_eval_item(row: pd.Series, structured: bool) -> EvalInputItem:
                 output_obj=row.get(self.generated_answer_key, "") if structured else "",
                 trajectory=row.get(self.trajectory_key, []) if structured else [],
                 expected_trajectory=row.get(self.expected_trajectory_key, []) if structured else [],
+                full_dataset_entry=row.to_dict() if self.dataset_config.pass_full_entry else None,
```
mdemoret-nv: Whats the downside to always including it? Memory? Minimizing output size?

mdemoret-nv (changes requested):
There is a disconnect here between where this setting is specified and where it will be used. The setting is on the dataset specification but it would only be used by evaluators. So its possible to set `pass_full_entry=False` and then use a metric which requires the full entry.

Instead, why dont we just always include the full entry and then add options to filter it from the saved output? I dont see any reason why we shouldnt include the full entry by default to the evaluators.

src/aiq/eval/dataset_handler/dataset_handler.py
AnuradhaKaruppiah: 1. Yes on the memory increase, I am not entirely sure what else is maintained in a (non-ragas) dataset. But as I am also trying to allow dataset transformation in a different PR that may become less of a problem.
2.  I still need to maintain the other fields (input, expected_output) etc. in the EvalInputItem to avoid breaking backwards compatibility.

I can enable `pass_full_entry` and re-add something similar if including the full entry becomes a problem.

PR #399, opened 2025-06-27
WIP: Sizing calculator

AnuradhaKaruppiah:
## Description

Closes


docs/source/workflows/sizing-calc.md
```diff
@@ -0,0 +1,330 @@
...
+        "question": "What is the product of 3 and 7, and is it greater than the current hour?",
+    },
+    {
+        "id": 2,
+        "question": "What is the product of 4 and 5, and is it greater than the current hour?",
+    }
+]
+```
+
+### Specifying the Concurrency Range
```
slopp, line 86: Should we explain `num_passes`

docs/source/workflows/sizing-calc.md
```diff
@@ -0,0 +1,330 @@
...
+- `Estimated GPU count (LLM Latency)`: Estimated GPU count based on the target LLM latency.
+
+You can use a maximum of the two estimates as the final GPU count to accommodate the target users.
+
+**Sample output:**
+```
+Targets: LLM Latency ≤ 0.0s, Workflow Runtime ≤ 10.0s, Users = 100
+Test parameters: GPUs = 8
+Per concurrency results:
+|   Concurrency |   p95 LLM Latency |   p95 WF Runtime |   Total Runtime |   Runtime OOR |   GPUs (WF Runtime, Rough) |
```
slopp, line 279: What is Runtime OOR ?

docs/source/workflows/sizing-calc.md
```diff
@@ -0,0 +1,330 @@
...
+**Target Parameters**
+
+To estimate the GPU cluster size, you need to specify the target number of users and the target workflow runtime i.e. the maximum acceptable response time for the workflow.
+
+Optionally you can specify the target p95 LLM latency if the LLM latency is a defining factor for the workflow and if it is possible to measure the maximum acceptable LLM latency.
+- `target_users`: Target number of users to support.
+- `target_workflow_runtime`: Target p95 workflow runtime (seconds). Can be set to 0 to ignore.
+- `target_llm_latency`: Target p95 LLM latency (seconds). Can be set to 0 to ignore.
+
+**Test Parameters**
```
slopp, line 210: I wonder if `--test_gpu_count` should be an input to the online mode that we store for users, rather than an input to the offline calculator?

docs/source/workflows/sizing-calc.md
```diff
@@ -0,0 +1,330 @@
...
+
+In addition to the concurrency range you can specify the number of passes made with each concurrency with the `--num_passes` command line parameter. By default the number of passes is one or a multiple of the concurrency if the dataset is larger than the concurrency value.
+
+If the size of the dataset is smaller than the concurrency range specified, the dataset is repeated to match the concurrency range.
+
+### Sample Output
+The per-concurrency metrics are stored in the `output_dir` specified in the command line. It is recommended to use a separate output directory for the calculator than the one used for the evaluation (specified via `eval.general.output_dir` in the workflow configuration file). This is to avoid accidental deletion of the calculator metrics when the evaluation jobs are cleaned up.
```
slopp: The re-use of `output_dir` might be confusing. I wonder if we should pick a different name to better distinguish the calculator run outputs from the per-concurrency eval outputs?

Similarly, we may want to re-name the append flag. 

Perhaps `calc_output_dir` and `append_calc_outputs` ?

AnuradhaKaruppiah:
Thanks for the review @slopp and @dnandakumar-nv. I will be renaming the tool to `aiq sizing calc` to disambiguate from the other profiling methods and address the rest of the recommendations.

docs/source/workflows/sizing-calc.md
AnuradhaKaruppiah: Agree, will fixup

docs/source/workflows/sizing-calc.md
AnuradhaKaruppiah: I am a bit torn about this. "test_gpu_count" is not needed for the "gathering metrics" part. 

At the same time it is directly associated with the online and only used as an artifact for the offline computations.

Based on that I will add it as an optional param to the `online` and store it in the results while allowing `offline` to override it.

docs/source/workflows/sizing-calc.md
AnuradhaKaruppiah: Addressed.

docs/source/workflows/sizing-calc.md
AnuradhaKaruppiah: I have piggybacked `num_passes`  on the same section

docs/source/workflows/sizing-calc.md
AnuradhaKaruppiah: These are dataset items with out-of-range values compared to the Target. 
Example: 
1. 5xitems with runtimes - 10, 9, 12, 7, 8
2. If the target p95 runtime is set to 10, this column will show 1 in the Runtime OOR column. I can add a legend to explain the OOR or pull it from the summary.

src/aiq/profiler/calc/calculations.py
```diff
@@ -0,0 +1,296 @@
...
+        Tuple of (cleaned_x, cleaned_y, list_of_removed_concurrencies)
+    """
+    # if the number of concurrency points is less removing outliers can be challenging
+    # as exteme outliers can skew the results.
+    # We use a threshold of 10 points to check for extreme outliers in raw y-values first.
+    small_concurrency_range_threshold = 8
+    # Extreme outlier threshold is 2.0 times the IQR, extreme outliers are removed.
+    extreme_outlier_threshold = 2.0
+    # Conservative outlier threshold is 1.5 times the IQR, conservative outliers are removed
+    conservative_outlier_threshold = 1.5
```
mpenn: Should we consider making these values configurable with sensible defaults?

src/aiq/profiler/calc/calculations.py
```diff
@@ -0,0 +1,296 @@
...
+
+    # Standard residual-based outlier detection
+    # Calculate residuals from a simple linear fit
+    sum_x = x.sum()
+    sum_y = y.sum()
+    sum_xy = (x * y).sum()
+    sum_x2 = (x**2).sum()
+
+    slope = (n * sum_xy - sum_x * sum_y) / (n * sum_x2 - sum_x**2)
+    intercept = (sum_y - slope * sum_x) / n
```
mpenn, line 149: Should we ensure n is >0 so we do not run into divide by zero edge cases?

src/aiq/cli/commands/sizing/calc.py
```diff
@@ -0,0 +1,312 @@
...
+logger = logging.getLogger(__name__)
+
+
+@click.command("calc", help="Estimate GPU count and plot metrics for a workflow")
+@click.option(
+    "--config_file",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, path_type=Path),
+    required=False,
+    default=None,
+    help="A YAML config file for the workflow and evaluation. This is not needed in offline mode.",
```
mdemoret-nv, line 36: What is offline mode?

mdemoret-nv: Do we need this after development?

src/aiq/cli/commands/sizing/calc.py
```diff
@@ -0,0 +1,312 @@
...
+    required=False,
+    default=False,
+    help="Run in offline mode. This is used to estimate the GPU count for a workflow without running the workflow. ")
+@click.option(
+    "--target_llm_latency",
+    type=float,
+    required=False,
+    default=0,
+    help="Target p95 LLM latency (seconds). Can be set to 0 to ignore.",
+)
```
mdemoret-nv, line 50: Should we remove the target llm latency per our discussion?

src/aiq/cli/commands/sizing/calc.py
```diff
@@ -0,0 +1,312 @@
...
+@click.option(
+    "--target_workflow_runtime",
+    type=float,
+    required=False,
+    default=0,
+    help="Target p95 workflow runtime (seconds). Can be set to 0 to ignore.",
+)
```
mdemoret-nv, line 57: Can we call this latency too?

src/aiq/eval/config.py
```diff
@@ -17,20 +17,33 @@
...
     reps: int = 1
     override: tuple[tuple[str, str], ...] = ()
+    # If false, the output will not be written to the output directory. This is
+    # useful when running evaluation via another tool.
+    write_output: bool = True
+    # if true, the dataset is adjusted to a multiple of the concurrency
+    adjust_dataset_size: bool = False
+    # number of passes at each concurrency, if 0 the dataset is adjusted to a multiple of the
+    # concurrency. The is only used if adjust_dataset_size is true
+    num_passes: int = 0
```
mdemoret-nv, line 46: This setting can be used when profiling, normal evaluation, and with the sizing calc. Does it make sense in all contexts?

docs/source/workflows/sizing-calc.md
AnuradhaKaruppiah: I have removed these from the summary table. It is available in the json output (with more description in the pydantic model).

docs/source/workflows/sizing-calc.md
AnuradhaKaruppiah: I am going to defer this change to a subsequent iteration.

src/aiq/cli/commands/sizing/calc.py
AnuradhaKaruppiah: It would be consistent but could also be confusing in comparison with the profiler metrics. 
Latency has been used for single LLM calls and runtime for workflow stats in the profiler. Changing this keyword to `target_workflow_latency` may create the confusion that it only includes the LLM call times.

src/aiq/eval/config.py
AnuradhaKaruppiah: It does, and single evaluation runs do have that as a part of the config for regular eval/profiling.

In the case of calc it means a bit more because of the need to first adjust the dataset to be a mulltiple of the concurrency .

src/aiq/cli/commands/sizing/calc.py
AnuradhaKaruppiah: There have been some differing opinions on what metric is useful. I would like to leave it there for now. I can phase it out based on perf experiments.

src/aiq/cli/commands/sizing/calc.py
AnuradhaKaruppiah: It can be useful post dev. Gathering metrics can be very time consuming. If someone wanted to just change the targets while experimenting with their cluster having to run everything again becomes a pain point.

src/aiq/profiler/calc/calculations.py
AnuradhaKaruppiah: Good catch. I added a check/ValueError exception for n=0

src/aiq/profiler/calc/calculations.py
AnuradhaKaruppiah: Agree, I have add a FitConfig that gives programmatic users fine-gained control of the parameters used for the linear fit

mpenn (approved):

PR #414, opened 2025-07-07
Responses API support

dnandakumar-nv:
## Description
Adds support for OpenAI Responses API and an agent that uses it


examples/advanced/alert_triage_agent/tests/test_alert_triage_agent_workflow.py
```diff
@@ -37,19 +37,24 @@ async def test_full_workflow():
...
+    repo_root = Path(__file__).resolve().parents[4]  # adjust if directory depth changes
+    input_filepath_abs = (repo_root / input_filepath).resolve()
 
     # Load input data
-    with open(input_filepath_abs, 'r') as f:
+    with open(input_filepath_abs, 'r', encoding='utf-8') as f:
         input_data = json.load(f)
 
-    # Run the workflow
+    # # Run the workflow
```
mpenn: Might have an extra `#` here

examples/advanced/alert_triage_agent/tests/test_alert_triage_agent_workflow.py
```diff
@@ -37,19 +37,24 @@ async def test_full_workflow():
...
     results = []
     async with load_workflow(config_file) as workflow:
+        pass
```
mpenn: Any reason for this `pass`?

src/aiq/profiler/callbacks/langchain_callback_handler.py
```diff
@@ -48,7 +49,10 @@ def _extract_tools_schema(invocation_params: dict) -> list:
     tools_schema = []
     if invocation_params is not None:
         for tool in invocation_params.get("tools", []):
-            tools_schema.append(ToolSchema(**tool))
+            try:
+                tools_schema.append(ToolSchema(**tool))
+            except Exception:
+                pass
```
mpenn: Do we have a good understanding of what might cause this to fail? Does the `ToolSchema` data model need to be revisited? Also wondering if we should log a debug or warning message here?

src/aiq/profiler/callbacks/langchain_callback_handler.py
```diff
@@ -33,6 +33,7 @@
 from aiq.builder.framework_enum import LLMFrameworkEnum
 from aiq.data_models.intermediate_step import IntermediateStepPayload
 from aiq.data_models.intermediate_step import IntermediateStepType
+from aiq.data_models.intermediate_step import ServerToolUseSchema
 from aiq.data_models.intermediate_step import StreamEventData
 from aiq.data_models.intermediate_step import ToolSchema
 from aiq.data_models.intermediate_step import TraceMetadata
```
mpenn: Is the `ServerToolUseSchema` going to be a consistent `TraceMetadata.tool_outputs` data model that can be reused for each LLM framework? If so, we should consider revisiting the `TraceMetadata` data model.

src/aiq/profiler/callbacks/llama_index_callback_handler.py
```diff
@@ -176,7 +188,8 @@ def on_event_end(
                         name=model_name,
                         UUID=event_id,
                         data=StreamEventData(input=self._run_id_to_llm_input.get(event_id), output=llm_text_output),
-                        metadata=TraceMetadata(chat_responses=response.message if response.message else None),
+                        metadata=TraceMetadata(chat_responses=response.message if response.message else None,
```
mpenn, line 212: In our callback handlers, default values are sometimes empty lists and sometimes `None`. We should unify this to help to simplify use of our `IntermediateStep` object.

packages/aiqtoolkit_agno/src/aiq/plugins/agno/llm.py
```diff
@@ -18,18 +18,29 @@
...
 from aiq.llm.nim_llm import NIMModelConfig
 from aiq.llm.openai_llm import OpenAIModelConfig
 
 
+def _validate_no_responses_api(llm_config):
+    """Validate that the LLM config does not use the Responses API."""
+
+    if llm_config.api_type == APITypeEnum.RESPONSES:
+        raise ValueError("Responses API is not supported with Agno. Please use a different API type.")
+
```
mpenn: A very similar function is used in the crew LLM client file. We do a similar check in other LLM clients. Would it make sense to move this to a utility function?

examples/advanced/alert_triage_agent/tests/test_alert_triage_agent_workflow.py
dnandakumar-nv: Accidental. Removing

src/aiq/profiler/callbacks/langchain_callback_handler.py
dnandakumar-nv: No, its nothing to do with the data model. This is an artifact of the responses API, where OpenAI has tools on server side whose schemas we don't have access to. 

So there's no schema to parse. I can log a one time warning, but I want to avoid logging a warning every time because that can get quite verbose

src/aiq/profiler/callbacks/langchain_callback_handler.py
dnandakumar-nv: It is only consistent for LLMs that support the responses API and is not universally applicable. I've annotated the `TraceMetadata.tool_outputs` model with the `ServerToolUseSchema` type hint. But I don't think we should limit the `tool_outputs` to just that schema as other APIs may have a different behavior. Thoughts?

src/aiq/profiler/callbacks/llama_index_callback_handler.py
dnandakumar-nv: Good catch. Made consistent by assigning empty list like we do with other fields

src/aiq/profiler/callbacks/langchain_callback_handler.py
dnandakumar-nv: Increased standardization of tool inputs and outputs to be addressed in a separate PR

src/nat/data_models/llm.py
```diff
@@ -14,14 +14,29 @@
...
     """Base configuration for LLM providers."""
-    pass
+
+    api_type: APITypeEnum = Field(default=APITypeEnum.CHAT_COMPLETION,
+                                  description="The type of API to use for the LLM provider.",
+                                  json_schema_extra={
+                                      "enum": [e.value for e in APITypeEnum],
+                                      "examples": [e.value for e in APITypeEnum],
+                                  },
+                                  exclude=True)
```
willkill07: I know you want `exclude=True` to work, but it doesn't.

Excluding means that the deep copying we do with configurations will strip this away.

src/nat/utils/responses_api.py
```diff
@@ -0,0 +1,25 @@
...
+
+from nat.data_models.llm import APITypeEnum
+
+
+def validate_no_responses_api(llm_config):
+    """Validate that the LLM config does not use the Responses API."""
+
+    if llm_config.api_type == APITypeEnum.RESPONSES:
+        raise ValueError("Responses API is not supported for config %s. Please use a different API type.",
+                         str(llm_config))
```
willkill07: This will dump out the entire configuration. Instead, maybe just print out the type?

```suggestion
        raise ValueError(f"Responses API is not supported for config {str(type(llm_config))}. Please use a different API type.")
```

src/nat/profiler/callbacks/llama_index_callback_handler.py
```diff
@@ -168,6 +189,18 @@ def on_event_end(
...
+                if response and response.additional_kwargs and "built_in_tool_calls" in response.additional_kwargs:
+                    tools_outputs = response.additional_kwargs["built_in_tool_calls"]
+                    if isinstance(tools_outputs, list):
+                        for tool in tools_outputs:
+                            try:
+                                tool_outputs_list.append(ServerToolUseSchema(**tool.model_dump()))
+                            except Exception:
+                                pass
+
+                # Append usage data to AIQ Toolkit usage stats
```
willkill07:
```suggestion
                # Append usage data to NAT usage stats
```

src/nat/profiler/callbacks/langchain_callback_handler.py
```diff
@@ -92,12 +100,22 @@ def _extract_token_base_model(self, usage_metadata: dict[str, Any]) -> TokenUsag
...
+            cache_tokens = 0
+            reasoning_tokens = 0
+
+            if "input_token_details" in usage_metadata:
+                if "cache_read" in usage_metadata["input_token_details"]:
+                    cache_tokens = usage_metadata["input_token_details"]["cache_read"]
+
+            if "output_token_details" in usage_metadata:
+                if "reasoning" in usage_metadata["output_token_details"]:
+                    reasoning_tokens = usage_metadata["output_token_details"]["reasoning"]
```
willkill07: Style nit
```suggestion
            cache_tokens = usage_metadata.get("input_token_details", {}).get("cache_read", 0)

            reasoning_tokens = usage_metadata.get("output_token_details", {}).get("reasoning", 0)
```

packages/nvidia_nat_llama_index/src/nat/plugins/llama_index/llm.py
```diff
@@ -82,6 +83,10 @@ async def aws_bedrock_llama_index(llm_config: AWSBedrockModelConfig, _builder: B
 
     from llama_index.llms.bedrock import Bedrock
 
+    if llm_config.api_type != APITypeEnum.CHAT_COMPLETION:
+        raise ValueError("AWS Bedrock LLM only supports chat completion API type. "
+                         f"Received: {llm_config.api_type}")
```
willkill07: Use/prefer `validate_no_responses_api` ?

packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
```diff
@@ -134,6 +147,10 @@ async def nim_langchain(llm_config: NIMModelConfig, _builder: Builder):
 
     from langchain_nvidia_ai_endpoints import ChatNVIDIA
 
+    if llm_config.api_type != APITypeEnum.CHAT_COMPLETION:
+        raise ValueError("NVIDIA AI Endpoints only supports chat completion API type. "
+                         f"Received: {llm_config.api_type}")
```
willkill07: What about `openai/gpt-oss-120b` ?
https://build.nvidia.com/openai/gpt-oss-120b/modelcard

or is this because langchain-nvidia doesn't?

packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
```diff
@@ -110,6 +115,10 @@ async def aws_bedrock_langchain(llm_config: AWSBedrockModelConfig, _builder: Bui
 
     from langchain_aws import ChatBedrockConverse
 
+    if llm_config.api_type != APITypeEnum.CHAT_COMPLETION:
+        raise ValueError("AWS Bedrock only supports chat completion API type. "
+                         f"Received: {llm_config.api_type}")
+
```
willkill07: Prefer `validate_no_responses_api` ?

src/nat/utils/responses_api.py
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# pylint: disable=raising-format-tuple
+
+from nat.data_models.llm import APITypeEnum
+
+
+def validate_no_responses_api(llm_config):
```
willkill07: I think this should also accept the framework enum for a clearer message.

packages/nvidia_nat_llama_index/pyproject.toml
```diff
@@ -23,17 +23,17 @@ dependencies = [
...
   "llama-index-embeddings-azure-openai~=0.3.9",
   "llama-index-embeddings-nvidia~=0.3.1",
   "llama-index-embeddings-openai~=0.3.1",
   "llama-index-llms-azure-openai~=0.3.2",
   "llama-index-llms-bedrock~=0.3.8",
   "llama-index-llms-litellm~=0.5.1",
-  "llama-index-llms-nvidia~=0.3.1",
-  "llama-index-llms-openai~=0.3.42",
+  "llama-index-llms-nvidia~=0.3.4",
+  "llama-index-llms-openai>=0.4.2",
```
willkill07: To maintain `~=` like semantics
```suggestion
  "llama-index-llms-openai>=0.4.2,<1.0.0",
```

packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
dnandakumar-nv: Yes, the `ChatNVIDIA` client doesn't support it. So developers will need to use `_type: openai` with NIMs

examples/agents/tool_calling/README.md
```diff
@@ -177,3 +178,105 @@ curl --request POST \
...
+- Connect to remote tools exposed over the Model Context Protocol (MCP).
+
+For current capabilities and model support, see OpenAI's documentation for the Responses API.
+
+#### Prerequisites
+- Set your OpenAI API key in the environment:
+  ```bash
+  export OPENAI_API_KEY=<YOUR_OPENAI_API_KEY>
+  ```
+- Use an OpenAI model that supports the Responses API and the tools you enable (for example, `o3`, `o4-mini`, or a compatible `gpt-*` model that advertises Responses support).
```
willkill07: This should be moved up to the Installation and Setup Section (see other examples).

Additionally, it would be nice to show a config using an NVIDIA NIM (like gpt-oss-120b) with this responses API too. I think it could just be an additional configuration file.

Issue #161, opened 2025-04-29
[FEA]: `ReAct` agent should suppress the warning message for `JSONDecodeError`


yczhang-nv:
### Is this a new feature, an improvement, or a change to existing functionality?

Improvement

### How would you describe the priority of this feature request

Low (would be nice)

### Please provide a clear description of problem this feature solves

Currently `ReAct` agent generates the error messages below when the `tool_input` can not be parsed as valid JSON. It is misleading to users since the agent will take raw string as input in that case and still generate the correct answer. Should suppress that error message and move it into `debug` level.

```bash
------------------------------
2025-04-29 18:13:41,059 - aiq.agent.react_agent.agent - WARNING - [AGENT] Unable to parse structured tool input from Action Input. Using Action Input as is.
Parsing error: Expecting value: line 1 column 1 (char 0)
Traceback (most recent call last):
  File "/mnt/schilton/agentiq/src/aiq/agent/react_agent/agent.py", line 256, in tool_node
    tool_input_dict = json.loads(tool_input_str) if tool_input_str != 'None' else tool_input_str
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
2025-04-29 18:13:45,537 - aiq.agent.react_agent.agent - INFO - 
------------------------------
```

### Describe your ideal solution

Should suppress that error message and move it into `debug` level.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct
- [x] I have searched the [open feature requests](https://github.com/NVIDIA/AgentIQ/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22%2Cimprovement%2Cenhancement) and have found no duplicates for this feature request

Tintin-Wu:
[...]

PR #418, opened 2025-07-08
Improve error handling and recovery mechanisms in agents

yczhang-nv:
## Description

This PR has been updated after [PR-462](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/462) is merged to utilize the retry logic patch feature.

This PR enhances the robustness of the agents by implementing comprehensive error handling and recovery mechanisms across the codebase.

- Enhanced Agent Error Handling:
  - Added retry logic with exponential backoff for tool execution
  - Implemented graceful error handling that returns error messages instead of raising exceptions
  - Improved the configuration options of `ReAct` agent according to https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues/161#issuecomment-2935966614, let users to have better control over the retry logics. 
    - This modified the option names. To avoid breaking changes, the options are backward compatible to the old names, by using `Pydantic` alias.
  - Modified ReAct, ReWOO, and Tool Calling agents to use new error handling patterns
  - Added consistent logging and response formatting for better debugging

- Type Conversion Safety:
  - Introduced `try_convert()` method in TypeConverter that handles conversion failures gracefully
  - Added warning logging when type conversions fail, allowing processing to continue with original data

Closes AIQ-938
Closes https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues/161


src/aiq/agent/react_agent/register.py
```diff
@@ -42,11 +42,15 @@ class ReActAgentWorkflowConfig(FunctionBaseConfig, name="react_agent"):
...
     verbose: bool = Field(default=False, description="Set the verbosity of the react agent's logging.")
-    retry_parsing_errors: bool = Field(default=True, description="Specify retrying when encountering parsing errors.")
-    max_retries: int = Field(default=1, description="Sent the number of retries before raising a parsing error.")
+    parsing_agent_response_max_retries: int = Field(
+        default=1, description="The number of retries of parsing the agent's response before raising a parsing error.")
+    tool_call_max_retries: int = Field(default=1, description="The number of retries before raising a tool call error.")
+    max_tool_calls: int = Field(default=15, description="Maximum number of tool calls before stopping the agent.")
+    pass_tool_call_errors_to_agent: bool = Field(
+        default=True,
+        description="Whether to pass tool call errors to agent. If False, failed tool calls will raise an exception.")
```
mdemoret-nv: This will be a breaking change. Is there a way that we can reuse the old options and only add new ones?

src/aiq/agent/react_agent/register.py
```diff
@@ -101,15 +107,15 @@ async def _response_fn(input_message: AIQChatRequest) -> AIQChatResponse:
...
+            state = await graph.ainvoke(state, config={'recursion_limit': (config.max_tool_calls + 1) * 2})
             # setting recursion_limit: 4 allows 1 tool call
             #   - allows the ReAct Agent to perform 1 cycle / call 1 single tool,
             #   - but stops the agent when it tries to call a tool a second time
 
             # get and return the output from the state
             state = ReActGraphState(**state)
             output_message = state.messages[-1]  # pylint: disable=E1136
-            return AIQChatResponse.from_string(output_message.content)
+            return AIQChatResponse.from_string(str(output_message.content))
```
mdemoret-nv, line 129: I'm worried about unintended consequences of this. What is the content of `output_message.content`?

src/aiq/utils/type_converter.py
```diff
@@ -109,6 +109,18 @@ def convert(self, data, to_type: type[_T]) -> _T:
             return result
         raise ValueError(f"Cannot convert type {type(data)} to {to_type}. No match found.")
 
+    def convert_safe(self, data, to_type: type[_T]) -> _T:
```
mdemoret-nv: Can we rename this to `try_convert`?

mdemoret-nv (changes requested):
Can we try and refactor this to minimize breaking changes? This is our most used agent so backwards compatibility is a must

src/aiq/agent/react_agent/register.py
yczhang-nv: In the current implementation of `ReAct` agent, `output_message.content` only contains strings, so the code change here does not make any difference. Do you suggest to revert this change?

src/aiq/agent/react_agent/register.py
yczhang-nv: Provided backward compatibility by using `Pydantic` alias, for example
```python
    retry_agent_response_parsing_errors: bool = Field(
        default=True,
        validation_alias=AliasChoices("retry_agent_response_parsing_errors", "retry_parsing_errors"),
        description="Whether to retry when encountering parsing errors in the agent's response.")
```
Ensured this works by adding additional unit tests. 

As a result, config files that use the old option names will still work as expected, and the new users will understand the options better by looking into the new option names.

I think re-naming is necessary as this PR added more options to `ReAct` agent, it makes the original naming even more confusing, as mentioned [here](https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues/161#issuecomment-2935966614). Please let me know if this looks good to you.

src/aiq/agent/react_agent/register.py
```diff
@@ -42,11 +43,23 @@ class ReActAgentWorkflowConfig(FunctionBaseConfig, name="react_agent"):
                                           description="The list of tools to provide to the react agent.")
     llm_name: LLMRef = Field(description="The LLM model to use with the react agent.")
     verbose: bool = Field(default=False, description="Set the verbosity of the react agent's logging.")
-    retry_parsing_errors: bool = Field(default=True, description="Specify retrying when encountering parsing errors.")
-    max_retries: int = Field(default=1, description="Sent the number of retries before raising a parsing error.")
+    retry_agent_response_parsing_errors: bool = Field(
+        default=True,
+        validation_alias=AliasChoices("retry_agent_response_parsing_errors", "retry_parsing_errors"),
+        description="Whether to retry when encountering parsing errors in the agent's response.")
+    parsing_agent_response_max_retries: int = Field(
```
mdemoret-nv:
```suggestion
    parse_agent_response_max_retries: int = Field(
```
This sounds better

PR #462, opened 2025-07-21
Added automatic retry logic to clients and memory methods with exponential backoff

dnandakumar-nv:
## Description
Introduced a generic retry mechanism to handle transient errors in LLM, embedder, and memory clients. Configurable retry options such as retry count, status codes, and error messages were integrated, alongside expanded tests and implementation for sync, async, and generator methods. 

Retries are automatically added with sensible defaults. Defaults can be modified in a config as follows:

```yaml
llms:
  nim_llm:
    _type: nim
    model_name: meta/llama-3.1-70b-instruct
    temperature: 0.0
    max_tokens: 1024
    
    # Retry configurations
    do_auto_retry: True
    num_retries: 5
    # Filters of HTTP Status Codes are in exc
    retry_on_status_codes: [429,500,504]
    # Filters of error messages contain the following strings
    retry_on_errors: ["Too Many Requests"]
```

Any other class in the toolkit or developers' source code can also be automatically patched with retries as follows: 

```python
from aiq.utils.exception_handlers.automatic_retries import patch_with_retry

class APIError(Exception):
    def __init__(self, status: int, msg: str):
        self.status = status
        super().__init__(msg)

class Client:
    def fetch(self):
        print("fetching …")
        # 503 with a typical server message
        raise APIError(503, "Service temporarily unavailable")

svc = Client()

svc = patch_with_retry(
    svc,
    retries=4,
    retry_codes=["5xx", 429],          # any 5xx or 429
    retry_on_messages=["temporarily unavailable",
                       "rate limit"]
)

svc.fetch()     # will retry (matches both 5xx + message filter)

```


src/aiq/utils/exception_handlers/automatic_retries.py
```diff
@@ -0,0 +1,268 @@
...
+
+        original = attr.__func__ if isinstance(attr, types.MethodType) else attr
+        wrapped = deco(original)
+
+        # Try an instance‑level monkey‑patch first (if we have an instance)
+        if not inspect.isclass(obj):
+            try:
+                object.__setattr__(obj, name, types.MethodType(wrapped, obj))
+                continue  # succeed, go to next attribute
+            except Exception as exc:  # noqa: BLE001
```
mpenn, line 268: The class level patch will be part of the class definition and effect all instances. I wonder if some additional context in the log message will allow users to know the implications of the class patch fallback mechanism.

mpenn (commented):
Overall, this looks good, just one minor comment re: the class level patching fallback.

src/aiq/utils/exception_handlers/automatic_retries.py
dnandakumar-nv: Good call! I'll update it to be a more detailed error log

PR #487, opened 2025-07-25
Pin `datasets` version in toplevel `pyproject.toml`

willkill07:
## Description

`uv`'s solver incorrectly chooses wildly different `datasets` versions based on platform. On linux, 4.0.0 is selected while on non-linux systems an older version (2.14.X) was chosen.

With the recent release of pyarrow 21, they removed deprecated code that the older version of datasets was dependent upon.

This pins datasets accordingly so the solver doesn't yield package solves that vary so much.

Closes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,8 +21,7 @@ dependencies = [
   "aioboto3>=11.0.0",
   "click~=8.1",
   "colorama~=0.4.6",
-  # workaround for uv's solver choosing different versions of datasets based on sys_platform
-  "datasets~=4.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -21,6 +21,8 @@ dependencies = [
   "aioboto3>=11.0.0",
   "click~=8.1",
   "colorama~=0.4.6",
+  # workaround for uv's solver choosing different versions of datasets based on sys_platform
+  "datasets~=4.0",
```
mdemoret-nv: Put this on the same line so sorting still works

mdemoret-nv (approved):

PR #489, opened 2025-07-26
Fix shared field reference bug in `TypedBaseModel` inheritance

mpenn:
## Description

All classes inheriting from `TypedBaseModel` shared the same Field object 
for the type field, causing the last-loaded class to overwrite the type 
names of all previous classes. This affected deep 
inheritance chains and mixin inheritance patterns.

- Store type name as class attribute instead of modifying shared field
- Implement model_post_init to set correct type after instance creation  
- Set type field default to "unknown" for validation compatibility

This fixes component discovery (e.g. `aiq info components -t tracing`) across all `TypedBaseModel` subclasses and 
enables proper inheritance patterns throughout the codebase.

### DIFF
--- a/tests/aiq/data_models/test_common.py
+++ b/tests/aiq/data_models/test_common.py
@@ -85,3 +85,185 @@ def test_subclass_depth():
                          ids=["dict-with-_type", "dict-with-type", "dict with both", "no_type", "object"])
 def test_type_discriminator(v: typing.Any, expected_value: str | None):
     assert common.TypedBaseModel.discriminator(v) == expected_value
+
+
+class TestTypedBaseModelInheritance:
+    """Test suite for TypedBaseModel inheritance and type handling."""
+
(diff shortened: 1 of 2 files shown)


tests/aiq/data_models/test_common.py
```diff
@@ -85,3 +85,185 @@ class GrandChild(Child):
...
+
+    def test_class_attribute_storage(self):
+        """Test that type names are stored as class attributes."""
+
+        class AttributeTestComponent(common.TypedBaseModel, name="attribute_test"):
+            pass
+
+        # Should have the class attribute
+        assert hasattr(AttributeTestComponent, '_typed_model_name')
+        assert AttributeTestComponent._typed_model_name == "attribute_test"
```
mdemoret-nv: This is testing the implementation and should be avoided

mdemoret-nv (approved):

PR #494, opened 2025-07-28
Extend authentication providers with HTTP request capabilities

ericevans-nv:
## Description

This PR introduces a mixin component that extends all authentication providers with a unified HTTP client interface for both authenticated and unauthenticated requests.

### Key Improvements

- **Backwards Compatibility**: Maintains full compatibility with existing authentication workflows while adding comprehensive HTTP request capabilities
- **Unified HTTP Interface**: All authentication providers now inherit consistent HTTP methods (GET, POST, PUT, DELETE, PATCH, HEAD, OPTIONS) through mixin architecture
- **Multiple Inheritance Architecture**: Clean integration of `httpx.AsyncClient` capabilities without breaking existing `AuthProviderBase` abstractions  
- Authenticated Request Tool: Added authenticated_request function to make both authenticated and unauthenticated HTTP requests.
- JIRA OAuth Integration Example: Complete example demonstrating OAuth 2.0 Authorization Code Flow with Atlassian JIRA.
- **Comprehensive Testing**: Full test coverage ensuring all HTTP methods work across all authentication providers


src/aiq/authentication/oauth2/oauth2_auth_code_flow_provider_config.py
```diff
@@ -27,6 +39,8 @@ class OAuth2AuthCodeFlowProviderConfig(AuthProviderBaseConfig, name="oauth2_auth
     token_endpoint_auth_method: str | None = Field(description="The authentication method for the token endpoint.",
                                                    default=None)
     scopes: list[str] = Field(description="The scopes for OAuth 2.0 authentication.", default_factory=list)
+    response_type: str | None = Field(default="code", description="The response type for OAuth 2.0 authentication.")
+    access_type: str | None = Field(default="offline", description="The access type for OAuth 2.0 authentication.")
```
mdemoret-nv: What is `access_type` used for? I dont see any references to it.

src/aiq/authentication/oauth2/oauth2_auth_code_flow_provider_config.py
```diff
@@ -13,10 +13,22 @@
...
+
 from pydantic import Field
+from pydantic import ValidationInfo
+from pydantic import field_validator
 
+from aiq.authentication.exceptions.auth_code_grant_exceptions import AuthCodeGrantConfigAuthorizationUrlFieldError
+from aiq.authentication.exceptions.auth_code_grant_exceptions import AuthCodeGrantConfigClientIDFieldError
+from aiq.authentication.exceptions.auth_code_grant_exceptions import AuthCodeGrantConfigClientSecretFieldError
+from aiq.authentication.exceptions.auth_code_grant_exceptions import AuthCodeGrantConfigClientServerUrlFieldError
+from aiq.authentication.exceptions.auth_code_grant_exceptions import AuthCodeGrantConfigScopeFieldError
```
mdemoret-nv: Lets re-use exceptions. These could all be handed with standard exceptions. Creating a new exception for each field makes it hard to catch these specific exceptions and is a lot of extra work.

src/aiq/authentication/oauth2/oauth2_auth_code_flow_provider_config.py
```diff
@@ -47,3 +61,134 @@ class OAuth2AuthCodeFlowProviderConfig(AuthProviderBaseConfig, name="oauth2_auth
...
+            raise AuthCodeGrantConfigAuthorizationUrlFieldError('invalid_url',
+                                                                '{field_name} must have a valid hostname',
+                                                                {'field_name': info.field_name})
+
+        # Ensure the URL includes a specific path and is not just the domain root
+        if not parsed.path or parsed.path == '/':
+            raise AuthCodeGrantConfigAuthorizationUrlFieldError(
+                'path_missing',
+                '{field_name} should include a valid endpoint path (e.g., /oauth/authorize)',
+                {'field_name': info.field_name})
```
mdemoret-nv: Is there a reason were adding all of this logic? Seems very restrictive.

src/aiq/authentication/oauth2/oauth2_auth_code_flow_provider_config.py
```diff
@@ -47,3 +61,134 @@ class OAuth2AuthCodeFlowProviderConfig(AuthProviderBaseConfig, name="oauth2_auth
...
+        if not parsed.netloc:
+            raise AuthCodeGrantConfigClientServerUrlFieldError('invalid_url', 'client_url must have a valid hostname')
+
+        if parsed.scheme == 'http' and not parsed.netloc.startswith(('localhost', '127.0.0.1')):
+            logger.warning(
+                'HTTP is not recommended for production environment. '
+                'Use HTTPS instead for production environment. Value: %s://',
+                parsed.scheme)
+
+        return value
```
mdemoret-nv: `client_url` will be removed.

src/aiq/authentication/oauth2/oauth2_auth_code_flow_provider_config.py
```diff
@@ -47,3 +61,134 @@ class OAuth2AuthCodeFlowProviderConfig(AuthProviderBaseConfig, name="oauth2_auth
...
+                                                     'Scopes cannot contain only whitespace',
+                                                     {'whitespace_scopes': whitespace_scopes})
+
+        # Check for duplicate scopes
+        if len(value) != len(set(value)):
+            duplicates = [scope for scope in set(value) if value.count(scope) > 1]
+            raise AuthCodeGrantConfigScopeFieldError('duplicate_found',
+                                                     'Duplicate scope found: {duplicates}', {'duplicates': duplicates})
+
+        return value
```
mdemoret-nv: These validators for `scope` are good.

src/aiq/tool/authenticated_request.py
```diff
@@ -0,0 +1,188 @@
...
+from aiq.authentication.interfaces import AuthProviderBaseConfig
+from aiq.data_models.authentication import AuthResult
+from aiq.data_models.authentication import HTTPMethod
+from aiq.data_models.authentication import HTTPResponse
+from aiq.utils.request_utils import HTTPRequestValidator
+
+logger = logging.getLogger(__name__)
+
+
+async def make_authenticated_request(url: str,
```
mdemoret-nv: Were setting ourselves up to need coverage for each and every possible combination of settings that can be made in a request.

What if we took a different approach like what authlib did? Where we derive from `httpx.AsyncClient` and add in our auth pieces? https://github.com/authlib/authlib/blob/main/authlib/integrations/httpx_client/oauth2_client.py#L63

src/aiq/tool/authenticated_request.py
```diff
@@ -0,0 +1,188 @@
...
+                                     auth_client: AuthProviderBaseConfig,
```
mdemoret-nv: This should be a full instance of `AuthProviderBase`, not the config.

src/aiq/tool/authentication_tools.py
```diff
@@ -56,3 +60,73 @@ async def _arun(user_id: str) -> AuthResult:
             return f"HTTP Basic authentication for '{user_id}' failed: {str(e)}"
 
     yield FunctionInfo.from_fn(_arun, description="Perform authentication with a given user ID.")
+
```
mdemoret-nv: I think we have a misunderstanding of how tools work. This is not how tools work. An agent should never need to use a tool like this.

src/aiq/utils/request_utils.py
```diff
@@ -0,0 +1,246 @@
...
+    """
+    HTTP request validation utility class.
+
+    Provides comprehensive validation for HTTP request components including URLs, methods,
+    headers, query parameters, and body data. Supports flexible input parsing for both
+    JSON strings and dictionaries.
+    """
+
+    @staticmethod
+    def parse_flexible_input(input_data: str | dict | None, input_name: str) -> Optional[Dict[str, Any]]:
```
mdemoret-nv: This is a lot of validation were adding that would already belong in `httpx`, right?

examples/front_ends/simple_auth/src/simple_auth/configs/config.yml
```diff
@@ -40,6 +46,34 @@ llms:
...
+      - write:jira-work
+      - manage:jira-webhook
+      - manage:jira-data-provider
+      - read:servicedesk-request
+      - manage:servicedesk-customer
+      - write:servicedesk-request
+      - read:servicemanagement-insight-objects
+      - "offline_access"
+    use_pkce: false
+
```
mdemoret-nv: This will prevent the example from running on its own without a Jira instance the user can test. I would prefer a second config or including it in a different example.

mdemoret-nv (changes requested):
Needs some work before we can merge

src/aiq/authentication/oauth2/oauth2_auth_code_flow_provider_config.py
ericevans-nv: That was necessary when I was writing the OAuth clients from scratch, but since we're using Authlib now, it's no longer needed.

src/aiq/tool/authentication_tools.py
ericevans-nv: I believe this may have been a test tool used during the early stages of the refactor to validate the initial iterations of the interface improvements.

src/nat/authentication/mixins/auth_provider_mixin.py
dnandakumar-nv, line 1: Can we rename `AuthProviderMixin` to something more descriptive of the role of the class? Perhaps `AsyncHTTPClientMixin`? I feel `AuthProviderMixin` is a little too broad and does not provide info into what the mixin does.

src/nat/authentication/mixins/auth_provider_mixin.py
```diff
@@ -0,0 +1,225 @@
...
+        All HTTP configuration (timeouts, SSL, proxies, etc.) is handled
+        at the request level
+        """
+        super().__init__()
+
+    async def request(self,
+                      method: str,
+                      url: str,
+                      user_id: str | None = None,
+                      apply_auth: bool = True,
```
dnandakumar-nv, line 50: Open to feedback on this, but, IMO if performing authentication is optional, this class shouldn't live inside the `authenticaton` directory. Perhaps it can live in utils as a standalone `AsyncHTTPClientMixin` and then the auth providers can inherit from it?

src/nat/authentication/mixins/auth_provider_mixin.py
ericevans-nv: I am in favor of this suggestion.

src/nat/authentication/mixins/auth_provider_mixin.py
ericevans-nv: I support this approach, as it cleanly separates core authentication logic from its extended capabilities.

examples/front_ends/authenticated_requests/src/authenticated_requests/api_requests.py
```diff
@@ -0,0 +1,109 @@
...
+@register_function(config_type=AuthenticatedRequestConfig)
+async def authenticated_request_function(config: AuthenticatedRequestConfig, builder: Builder):
+
+    auth_provider = await builder.get_auth_provider(config.auth_provider)
+
+    async def _inner(url: str,
+                     method: Literal["GET", "POST", "PUT", "DELETE", "PATCH"] = "GET",
+                     user_id: str | None = None,
+                     authenticated: bool = True,
+                     json_data: dict | None = None) -> str:
```
mdemoret-nv, line 49: This tool would never exist. You wouldnt want to give an agent this much freedom in making a request. It would very rarely succeed unless you gave it explicit instructions (which takes much of the benefit out of the tool).

Instead, this should call a predefined endpoint. Like what about fetching the available list of projects from JIRA? Getting a list of assigned issues for the currently logged in user? etc.

src/nat/authentication/oauth2/oauth2_auth_code_flow_provider.py
```diff
@@ -20,14 +20,15 @@
...
+from nat.authentication.mixins import AuthProviderMixin
 from nat.authentication.oauth2.oauth2_auth_code_flow_provider_config import OAuth2AuthCodeFlowProviderConfig
 from nat.builder.context import Context
 from nat.data_models.authentication import AuthFlowType
 from nat.data_models.authentication import AuthResult
 from nat.data_models.authentication import BearerTokenCred
 
 
-class OAuth2AuthCodeFlowProvider(AuthProviderBase[OAuth2AuthCodeFlowProviderConfig]):
+class OAuth2AuthCodeFlowProvider(AuthProviderBase[OAuth2AuthCodeFlowProviderConfig], AuthProviderMixin):
```
mdemoret-nv, line 31: This doesnt need to be a mixin. We should just use the `AuthProviderMixin` as a utility class (needs a better name, too). 

```python
provider = builder. get_auth_provider("jira")

async with AuthProviderMixin(provider) as client:
    response = await client.post(....)
```

examples/front_ends/authenticated_requests/src/authenticated_requests/api_requests.py
ericevans-nv: I’m in favor of this suggestion, as it helps scope the tool for greater success.

PR #510, opened 2025-07-29
Update prerequisite and system prompt of `redis` memory example

yczhang-nv:
## Description

For the `redis` example, updated the `uv sync` command to install necessary dependencies in the prerequisite section. Also updated the system prompt of the agent to ensure more robust behavior.

### DIFF
--- a/examples/memory/redis/README.md
+++ b/examples/memory/redis/README.md
@@ -41,7 +41,7 @@ Ensure that Docker is installed and the Docker service is running before proceed
 
 - Install Docker: Follow the official installation guide for your platform: [Docker Installation Guide](https://docs.docker.com/engine/install/)
 - Start Docker Service:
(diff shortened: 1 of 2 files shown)


examples/memory/redis/README.md
```diff
@@ -52,6 +52,11 @@ docker info
 
 If you have not already done so, follow the instructions in the [Install Guide](../../../docs/source/quick-start/installing.md#install-from-source) to create the development environment and install NeMo Agent toolkit.
 
+To run this example, install the required dependencies by running the following command:
```
willkill07, line 55: This works if in a development environment but not if they installed wheels.

I'm not sure what the correct base assumption here should be...

willkill07 (approved):
Approving with one comment/note

examples/memory/redis/README.md
mdemoret-nv: We assume for all examples that the user has cloned the repo.

PR #606, opened 2025-08-08
Remove `AIQ` prefix from class and function names

dagardner-nv:
## Description
* Strip the `AIQ` prefix from class names such that `AIQContext` => `Context`, the idea here is that we don't need the name of the library in the class names as they are already in the package/namespace of the library.
* Compatibility aliases were left in place to ease transition


packages/nvidia_nat_redis/src/aiq/plugins/redis/redis_editor.py
```diff
@@ -32,7 +32,7 @@
 
 class RedisEditor(MemoryEditor):
     """
-    Wrapper class that implements AIQ Toolkit Interfaces for Redis memory storage.
+    Wrapper class that implements NAT tnterfaces for Redis memory storage.
```
yczhang-nv:
```suggestion
    Wrapper class that implements NAT interfaces for Redis memory storage.
```

yczhang-nv (approved):
Almost all the changes are simply re-naming and LGTM. There are some "AIQ" left in `tool_converter.py` that needs to be updated.

PR #660, opened 2025-08-16
Timezone Support for `datetime` Tool and Normalize Response Time Handling

### LINKED ISSUE #445: [BUG]: Timezone Support to Client Requests and Normalize Response Time Handling
### Version

1.2.0

Source

### Describe the bug.

The current implementation does not account for user timezone information when making API requests or processing time-based responses. This leads to discrepancies between the time the user expects (in their local timezone) and the data returned from the server (which uses UTC).

```shell

```

### Relevant log output

User query: How many incidents occurred between 7:00 and 8:00 on 10-07? --- 7:00 GMT+7 and 8:00 GMT+7
My API tool agent: The API tool agent converts the user query into the following parameters: start_time=7:00, end_time=8:00. Convert gmt_to_utc: params {start_time=0:00, end_time=1:00}
My Web API: Return data in UTC time --- 0:00 UTC and 1:00 UTC.
LLM: Convert utc_to_gmt "There were 3 incidents that occurred between 00:00 and 01:00" -> "There were 3 incidents that occurred between 07:00 and 08:00"

</pre></details>

### Other/Misc.

_No response_

### Code of Conduct

zhongxuanwang-nv:
## Description

- Added timezone support for the `datetime_tools`, so now it will automatically extract the `x-timezone` from the headers from context, which will be used to format the current time. 
    - If the `x-timezone`  is not present in the headers, default UTC time is used. 
    - The documentations should also be clear now.

- Fixed a small issue in `NeMo-Agent-Toolkit/src/nat/builder/user_interaction_manager.py`. Before the timestamp is in UTC (as seen by the `Z` suffix in the timestamp), but the timestamp is in the local timezone, which is most likely not in UTC timezone.

- Made a tiny change for the `NeMo-Agent-Toolkit-UI` repository. So now the front-end also stores the timezone in IANA format in its header to be sent to the backend.
    - The change: [https://github.com/NVIDIA/NeMo-Agent-Toolkit-UI/pull/38](https://github.com/NVIDIA/NeMo-Agent-Toolkit-UI/pull/38)
### Q: why did you use UTC time by default?
- A: UTC is industry standard's default timezone.

### Q: why did you use IANA format for timezone?
- A: IANA time zone (like "America/New_York" or "Etc/UTC") provides a standardized identifier that encodes daylight-saving rules and ensures consistent behavior across platforms and regions, and also widely accepted in industry.

## Demo
Screenshot 2025-08-16 at 1 46 57 PM

Closes #445

### DIFF
--- a/src/nat/settings/global_settings.py
+++ b/src/nat/settings/global_settings.py
@@ -47,6 +47,12 @@ class Settings(HashableBaseModel):
     # Registry Handeler Configuration
     channels: dict[str, RegistryHandlerBaseConfig] = {}
 
+    # Timezone fallback behavior
+    # Options:
+    #  - "utc": default to UTC
+    #  - "system": use the system's local timezone
+    fallback_timezone: typing.Literal["system", "utc"] = "utc"
+
     _configuration_directory: typing.ClassVar[str]
     _settings_changed_hooks: typing.ClassVar[list[Callable[[], None]]] = []
     _settings_changed_hooks_active: bool = True
@@ -165,7 +171,11 @@ class Settings(HashableBaseModel):
             loaded_config = {}
         else:
             with open(file_path, mode="r", encoding="utf-8") as f:
-                loaded_config = json.load(f)
+                try:
+                    loaded_config = json.load(f)
+                except Exception as e:
+                    logger.exception("Error loading configuration file %s: %s", file_path, e)
+                    loaded_config = {}
 
         settings = Settings(**loaded_config)
         settings.set_configuration_directory(configuration_directory)
@@ -214,6 +224,8 @@ class Settings(HashableBaseModel):
                 match field:
                     case "channels":
                         self.channels = validated_data.channels
+                    case "fallback_timezone":
+                        self.fallback_timezone = validated_data.fallback_timezone
                     case _:
                         raise ValueError(f"Encountered invalid model field: {field}")
 
--- a/src/nat/tool/datetime_tools.py
+++ b/src/nat/tool/datetime_tools.py
@@ -36,7 +36,7 @@ async def current_datetime(_config: CurrentTimeToolConfig, _builder: Builder):
 
     from starlette.datastructures import Headers
 
-    async def _get_current_time(_unused: str) -> str:
+    async def _get_current_time(unused: str) -> str:
 
         from nat.builder.context import Context
         nat_context = Context.get()
--- a/src/nat/tool/datetime_tools.py
+++ b/src/nat/tool/datetime_tools.py
@@ -36,6 +36,8 @@ async def current_datetime(_config: CurrentTimeToolConfig, _builder: Builder):
 
     from starlette.datastructures import Headers
 
+    from nat.settings.global_settings import GlobalSettings
+
     async def _get_current_time(unused: str) -> str:
 
         del unused  # Unused parameter to avoid linting error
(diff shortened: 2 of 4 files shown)


src/nat/tool/datetime_tools.py
```diff
@@ -21,22 +21,49 @@
...
+    Simple tool which returns the current date and time in human readable format with timezone information. By default,
+    the timezone is in Etc/UTC. If the user provides a timezone in the header, we will use it. Timezone will be
+    provided in IANA zone name format. For example, "America/New_York" or "Etc/UTC".
     """
     pass
 
 
 @register_function(config_type=CurrentTimeToolConfig)
 async def current_datetime(config: CurrentTimeToolConfig, builder: Builder):
+    del config, builder  # Unused parameters to avoid linting error
```
willkill07: You can prefix the parameters with _ to satisfy pylint rather than invoke del

src/nat/tool/datetime_tools.py
```diff
@@ -21,22 +21,49 @@
 
 class CurrentTimeToolConfig(FunctionBaseConfig, name="current_datetime"):
     """
-    Simple tool which returns the current date and time in human readable format.
+    Simple tool which returns the current date and time in human readable format with timezone information. By default,
+    the timezone is in Etc/UTC. If the user provides a timezone in the header, we will use it. Timezone will be
+    provided in IANA zone name format. For example, "America/New_York" or "Etc/UTC".
```
willkill07, line 32: The only problem with the IANA zone format is that I suspect LLMs are more likely to get operations wrong based on time comparisons with IANA vs the UTC offset indicators (+5, -7) present with the `%z` format string.

I don't have a particular strong opinion on this, but log files and error messages more often include the time zone offset rather than the IANA name. I think for an agent tool, we should try to use whatever will most likely be encountered for date time comparisons.

src/nat/tool/datetime_tools.py
```diff
@@ -21,22 +21,49 @@
...
-        now = datetime.datetime.now()  # Get current time
+        del unused  # Unused parameter to avoid linting error
+
+        from nat.builder.context import Context
+        nat_context = Context.get()
+
+        headers: Headers | None = nat_context.metadata.headers
+
+        # Default timezone is UTC
+        timezone_obj = zoneinfo.ZoneInfo("Etc/UTC")
```
willkill07: If people are running NAT locally, they may prefer the timezone to be the same as the system. I wonder if exposing an optional configuration value (`fallback_timezone`) to one of `system` | `utc` would provide greater flexibility.

zhongxuanwang-nv: That's right, because the current approach only considers the invocation from front-end case. I do agree such a configuration value will offer greater flexibility; do you think I should also try adding this parameter as an optimal parameter to the configuration here?

willkill07: I think it would be welcome for this PR! Though I am unsure what the default should be.

src/nat/tool/datetime_tools.py
```diff
@@ -21,22 +21,57 @@
...
+
+    from starlette.datastructures import Headers
+
+    from nat.settings.global_settings import GlobalSettings
 
     async def _get_current_time(unused: str) -> str:
 
-        now = datetime.datetime.now()  # Get current time
-        now_human_readable = now.strftime(("%Y-%m-%d %H:%M:%S"))
+        del unused  # Unused parameter to avoid linting error
```
willkill07, line 65:
```suggestion
async def _get_current_time(_unused: str) -> str:
```

src/nat/tool/datetime_tools.py
```diff
@@ -21,22 +21,57 @@
...
+
+        from nat.builder.context import Context
+        nat_context = Context.get()
+
+        # Default to UTC
+        timezone_obj = zoneinfo.ZoneInfo("Etc/UTC")
```
willkill07: This seems a bit like a misnomer to have the default be UTC when the fallback is definable.

src/nat/tool/datetime_tools.py
```diff
@@ -21,22 +21,57 @@
...
+                    timezone_obj = zoneinfo.ZoneInfo(timezone_header)
+                except Exception:
+                    pass
+        else:
+            # Only if a timezone is not in the header, we will determine default timezone based on global settings
+            fallback_tz = GlobalSettings.get().fallback_timezone
+
+            if fallback_tz == "system":
+                # Use the system's local timezone. Avoid requiring external deps.
+                timezone_obj = datetime.datetime.now().astimezone().tzinfo or zoneinfo.ZoneInfo("Etc/UTC")
```
willkill07: I'd hoist the timezone logic out to its own function for clarity

src/nat/tool/datetime_tools.py
```diff
@@ -21,22 +21,57 @@
 
 class CurrentTimeToolConfig(FunctionBaseConfig, name="current_datetime"):
     """
-    Simple tool which returns the current date and time in human readable format.
+    Simple tool which returns the current date and time in human readable format with timezone information. By default,
+    the timezone is in Etc/UTC. If the user provides a timezone in the header, we will use it. Timezone will be
+    provided in IANA zone name format. For example, "America/New_York" or "Etc/UTC".
```
willkill07, line 32: It is still IANA?

src/nat/settings/global_settings.py
```diff
@@ -47,6 +47,12 @@ class Settings(HashableBaseModel):
     # Registry Handeler Configuration
     channels: dict[str, RegistryHandlerBaseConfig] = {}
 
+    # Timezone fallback behavior
+    # Options:
+    #  - "utc": default to UTC
+    #  - "system": use the system's local timezone
+    fallback_timezone: typing.Literal["system", "utc"] = "utc"
```
willkill07, line 54: 🎉

PR #663, opened 2025-08-18
Weave: Group workflow traces under the parent evaluation call

AnuradhaKaruppiah:
## Description

When an Evaluation run is started the eval_call id is pushed to the call stack so subsequent traces can be grouped underneath it. This allows the user to debug the predictions, scores and traces for each run easily.

- This PR is a workaround as the eval (score) exporting will be migrated to the observability exporter_manager in the future
- This workaround only works for local eval. Remote eval is yet to be solved.
- This PR also adds a change to wait on the trace export to finish before finishing the evaluation run

image

### DIFF
--- a/packages/nvidia_nat_opentelemetry/tests/observability/test_otel_span_adapter_exporter.py
+++ b/packages/nvidia_nat_opentelemetry/tests/observability/test_otel_span_adapter_exporter.py
@@ -241,7 +241,7 @@ class TestOTLPSpanAdapterExporter:
                 exporter.export(sample_end_event)
 
                 # Wait for async processing
-                await exporter._wait_for_tasks()
+                await exporter.wait_for_tasks()
 
             # Verify that export was called (span was processed and exported)
             mock_otlp_exporter.export.assert_called()
@@ -295,7 +295,7 @@ class TestOTLPSpanAdapterExporter:
                 exporter.export(end_event)
 
             # Wait for batch processing
-            await exporter._wait_for_tasks()
+            await exporter.wait_for_tasks()
 
         # Verify that export was called (batching should trigger export)
         mock_otlp_exporter.export.assert_called()
--- a/packages/nvidia_nat_opentelemetry/tests/observability/test_otel_span_adapter_integration.py
+++ b/packages/nvidia_nat_opentelemetry/tests/observability/test_otel_span_adapter_integration.py
@@ -124,7 +124,7 @@ class TestOTLPSpanAdapterExporterIntegration:
             exporter.export(end_event)
 
             # Wait for async export to complete
-            await exporter._wait_for_tasks()
+            await exporter.wait_for_tasks()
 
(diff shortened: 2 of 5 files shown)


src/nat/eval/utils/eval_trace_ctx.py
```diff
@@ -0,0 +1,86 @@
...
+
+    This class provides a framework-agnostic way to:
+    1. Track evaluation calls/contexts
+    2. Ensure proper parent-child relationships in traces
+    """
+
+    def __init__(self):
+        self.eval_call = None  # Store the evaluation call/context for propagation
+
+    def set_eval_call(self, eval_call: Any) -> None:
```
mpenn: Is this truly a `typing.Any`? or is there a more specific type hint we can use here?

src/nat/eval/utils/eval_trace_ctx.py
```diff
@@ -0,0 +1,86 @@
...
+        """Set the evaluation call/context for propagation to traces."""
+        self.eval_call = eval_call
+        if eval_call:
+            logger.debug("Set evaluation call context: %s", getattr(eval_call, 'id', str(eval_call)))
+
+    def get_eval_call(self) -> Any:
```
mpenn: Is this truly a `typing.Any`? or is there a more specific type hint we can use here?

src/nat/eval/utils/eval_trace_ctx.py
```diff
@@ -0,0 +1,86 @@
...
+
+class WeaveEvalTraceContext(EvalTraceContext):
+    """
+    Weave-specific implementation of evaluation trace context.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.available = False
+        self.set_call_stack: Callable[[Any], Any] | None = None
```
mpenn: Is this truly a `typing.Any`? or is there a more specific type hint we can use here?

src/nat/eval/utils/weave_eval.py
```diff
@@ -32,11 +32,12 @@ class WeaveEvaluationIntegration:  # pylint: disable=too-many-public-methods
     Class to handle all Weave integration functionality.
     """
 
-    def __init__(self):
+    def __init__(self, eval_trace_context):
```
mpenn: Suggest adding a type hint here.

src/nat/eval/utils/eval_trace_ctx.py
AnuradhaKaruppiah: Updated

src/nat/eval/utils/weave_eval.py
AnuradhaKaruppiah: Done

src/nat/eval/utils/eval_trace_ctx.py
AnuradhaKaruppiah: Updated

PR #669, opened 2025-08-19
Fix: Remove `pickle` from MySQL-based Object Store

willkill07:
## Description

Using pickle for SerDe is not secure, so switch to JSON + Base64 encode/decode for Object Store backing in MySQL.

This PR also updates necessary files under the `examples/object_store/user_report` example. The leading `/` was removed as it never should have been added in the first place.

Closes AIQ-1724

### DIFF
--- a/examples/object_store/user_report/serialize_file.py
+++ b/examples/object_store/user_report/serialize_file.py
@@ -13,10 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import pickle
+import base64
+import json
+import mimetypes
 import sys
-
-from nat.object_store.models import ObjectStoreItem
 
 # Usage: python serialize_file.py <file_path>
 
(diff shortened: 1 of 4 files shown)


examples/object_store/user_report/serialize_file.py
```diff
@@ -29,7 +29,11 @@
 with open(file_path, "rb") as f:
     data = f.read()
 
-item = ObjectStoreItem(data=data, )
+payload = {
```
mdemoret-nv: Why did we remove this? It should be an `ObjectStoreItem`

examples/object_store/user_report/src/nat_user_report/user_report_tools.py
```diff
@@ -37,7 +37,7 @@ async def get_user_report(config: GetUserReportConfig, builder: Builder):
 
     async def _inner(user_id: str, date: str | None = None) -> str:
         date = date or "latest"
-        key = f"/reports/{user_id}/{date}.json"
+        key = f"reports/{user_id}/{date}.json"
```
mdemoret-nv, line 40: Why do I feel like we have made this change before?

examples/object_store/user_report/serialize_file.py
mdemoret-nv: Can we use an object instead of a naked dictionary?

examples/object_store/user_report/src/nat_user_report/user_report_tools.py
willkill07: On `develop` PR we have :)

PR #694, opened 2025-08-21
Adopt ruff in CI

dagardner-nv:
* Replace isort, flake8 and pylint with `ruff check`
* Retain yapf for formatting. The reason is that `ruff format` is explicitly focused on being a drop-in replacement for Black, offering very few configuration options ref: https://docs.astral.sh/ruff/formatter/#configuration
* Update code to pass `ruff check`. Note previously not all of our code was passing with Pylint, so retaining Pylint would have likely required many of the same changes.


ci/scripts/path_checks.py
```diff
@@ -123,18 +123,14 @@
     "user/assistant",
     "validate/sanitize",
     "Workflows/tools",
-    "Yes/No",  #
-    # numbers
-    r"\d+/\d+(/\d+)*",  #
-    # LLM model names
+    "Yes/No",  # numbers
```
willkill07: All of the comments need shifted down one line for proper attribution :(

pyproject.toml
```diff
@@ -257,693 +254,46 @@ asyncio_mode = "auto"
...
+  "PLR0911", # too-many-return-statements
+  "PLR0912", # too-many-branches
+  "PLR0913", # too-many-arguments
+  "PLR0914", # too-many-locals
+  "PLR0915", # too-many-statements
+  "PLR0917", # too-many-positional-arguments
+  "PLR1702", # too-many-nested-blocks
+  "PLR2004", # Magic value used in comparison this happens a lot within unittests
+  "PLW0108", # unnecessary-lambda
+  "PLW2901", # allow re-assigning a loop variable
```
willkill07, line 288: I think this one is okay temporarily but we should probably (eventually) update code so we can remove this exception

dagardner-nv: Depends.

We have a lot of code that looks like:
```python
with open(file_name) as fh:
    for line in fh:
        line = line.strip() # Re-assigned loop variable
```

I think usage like that is OK. I think the warning might be valuable in a situation where
```python
for (i, var) in enumerate(somelist):
    # do something weird with i
```

but I'm not sure how often something like that would come up.

willkill07 (approved):

dagardner-nv:
@coderabbitai review

PR #716, opened 2025-08-25
Add aiqtoolkit-data-flywheel subpackage with NeMo Data Flywheel integrations

mpenn:
## Description

Closes


packages/nvidia_nat_data_flywheel/src/nat/plugins/data_flywheel/observability/processor/trace_conversion/span_to_dfw_record.py
```diff
@@ -0,0 +1,118 @@
...
+    """Extract string value from enum or literal type safely.
+
+    Args:
+        value (Any): Could be an Enum, string, or other type
+
+    Returns:
+        str: String representation of the value
+    """
+    if isinstance(value, Enum):
+        return value.value
```
willkill07: If an enum is non-str based, then you need to wrap
```suggestion
        return str(value.value)
```

packages/nvidia_nat_data_flywheel/src/nat/plugins/data_flywheel/observability/schema/sink/elasticsearch/dfw_es_record.py
```diff
@@ -0,0 +1,200 @@
...
+from pydantic import Field
+from pydantic import field_validator
+from pydantic import model_validator
+
+from nat.plugins.data_flywheel.observability.schema.schema_registry import register_schema
+
+logger = logging.getLogger(__name__)
+
+
+class ContractVersion(str, Enum):
```
willkill07: Why not use `ElasticsearchContractVersion` ?

Alternatively, why do we need both of these definitions?

willkill07 (commented):
Completed a first-pass

packages/nvidia_nat_data_flywheel/src/nat/plugins/data_flywheel/observability/register.py
```diff
@@ -0,0 +1,60 @@
...
+                                        BatchConfigMixin,
+                                        name="data_flywheel_elasticsearch"):
+    """A telemetry exporter to transmit traces to NVIDIA Data Flywheel via Elasticsearch."""
+
+    client_id: str = Field(description="The data flywheel client ID.")
+    index: str = Field(description="The elasticsearch index name.")
+    endpoint: str = Field(description="The elasticsearch endpoint.")
+    contract_version: ContractVersion = Field(default=ContractVersion.V1_1,
+                                              description="The DFW Elasticsearch record schema version to use.")
+    elasticsearch_auth: tuple[str, str] = Field(
```
willkill07: We should be more pragmatic here for authentication and explicitly use `username` and `password`, alternatively considering the exposition of environment variables like `NAT_ELASTICSEARCH_USERNAME` and `NAT_ELASTICSEARCH_PASSWORD` so users don't need to hard-code credentials.

Internally, it's fine to use the tuple, but for public exposition we should be clear

packages/nvidia_nat_data_flywheel/src/nat/meta/pypi.md
```diff
@@ -0,0 +1,23 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+![NVIDIA NeMo Agent Toolkit](https://media.githubusercontent.com/media/NVIDIA/NeMo-Agent-Toolkit/refs/heads/main/docs/source/_static/banner.png "NeMo Agent toolkit banner image")
+
+# NVIDIA Agent Toolkit Subpackage
```
willkill07, line 20:
```suggestion
# NVIDIA NeMo Agent Toolkit Subpackage
```

src/nat/profiler/decorators/function_tracking.py
```diff
@@ -252,3 +252,108 @@ def sync_wrapper(*args, **kwargs):
...
+    # --- Now detect the function type and wrap accordingly ---
+    if inspect.isasyncgenfunction(func):
+        # ---------------------
+        # ASYNC GENERATOR
+        # ---------------------
+
+        @functools.wraps(func)
+        async def async_gen_wrapper(*args, **kwargs):
+            context = Context.get()
+            input_data = args[0] if args else kwargs
```
willkill07: Is there any particular reason why we only grab the first parameter (or all kwargs)? Wouldn't all parameters be useful?

If it's not too much information to include all arguments, I propose:
```python
# tuple of expanded args with dict[str, Any] as last entry
input_data = (*args, kwargs, )
```
This would always provide the full context (all positional and keyword parameters).

PR #717, opened 2025-08-25
feat: Support for Nemotron thinking configuration

willkill07:
## Description

* Introduce a `ThinkingMixin` which defaults to `None` -- currently is only supported for NVIDIA Nemotron-like models.
* Add support to all providers to automatically insert an additional "thinking" system prompt if requested
* Add tests to validate expected behavior

Closes AIQ-1609

## Summary by CodeRabbit

* **New Features**
  * Optional "Thinking" setting for supported Nemotron models (defaults to None); when enabled, provides a model-specific system prompt and is applied across providers and SDK wrappers.

* **Improvements**
  * Clarified gated-field semantics and "by a model" applicability for mixins (e.g., temperature, top_p).
  * Centralized thinking injection and retry application across integrations for consistent behavior.
  * Simplified gated-field typing by removing generic parameterization.

* **Documentation**
  * Renamed gating docs to “Gated Field Mixins”, added Thinking mixin guidance; AWS Bedrock docs updated (top_p, context_size, max_retries).

* **Tests**
  * Added comprehensive tests for Thinking behavior, prompt injection, and gated-field cases.

* **Chores**
  * Added “Nemotron” to accepted vocabulary.

### DIFF
--- a/src/nat/data_models/thinking_mixin.py
+++ b/src/nat/data_models/thinking_mixin.py
@@ -20,6 +20,8 @@ from pydantic import Field
 
 from nat.data_models.gated_field_mixin import GatedFieldMixin
 
+# The system prompt format for thinking is different for these, so we need to distinguish them here with two separate
+# regex patterns
 _NVIDIA_NEMOTRON_REGEX = re.compile(r"^nvidia/nvidia.*nemotron", re.IGNORECASE)
 _LLAMA_NEMOTRON_REGEX = re.compile(r"^nvidia/llama.*nemotron", re.IGNORECASE)
 _MODEL_KEYS = ("model_name", "model", "azure_deployment")


packages/nvidia_nat_llama_index/src/nat/plugins/llama_index/llm.py
```diff
@@ -13,90 +13,97 @@
...
+@register_llm_client(config_type=AWSBedrockModelConfig, wrapper_type=LLMFrameworkEnum.LLAMA_INDEX)
+async def aws_bedrock_llama_index(llm_config: AWSBedrockModelConfig, _builder: Builder):
+
+    from llama_index.llms.bedrock import Bedrock
 
-    yield llm
+    # LlamaIndex uses context_size instead of max_tokens
+    llm = Bedrock(**llm_config.model_dump(exclude={"type", "top_p"}, by_alias=True), )
+
+    yield _patch_llm_based_on_config(llm, llm_config)
```
dagardner-nv, line 79: Would it be better to apply the `if isinstance(llm_config, ThinkingMixin)` check here and potentially avoid the call to `_patch_llm_based_on_config`?

src/nat/data_models/thinking_mixin.py
```diff
@@ -0,0 +1,66 @@
...
+
+import re
+
+from pydantic import BaseModel
+from pydantic import Field
+
+from nat.data_models.gated_field_mixin import GatedFieldMixin
+
+_NVIDIA_NEMOTRON_REGEX = re.compile(r"^nvidia/nvidia.*nemotron", re.IGNORECASE)
+_LLAMA_NEMOTRON_REGEX = re.compile(r"^nvidia/llama.*nemotron", re.IGNORECASE)
```
dagardner-nv, line 26: Just adding a comment

```suggestion
# The system prompt format for thinking is different for these, so we need to distinguish them here with two separate
# regex patterns
_NVIDIA_NEMOTRON_REGEX = re.compile(r"^nvidia/nvidia.*nemotron", re.IGNORECASE)
_LLAMA_NEMOTRON_REGEX = re.compile(r"^nvidia/llama.*nemotron", re.IGNORECASE)
```

dagardner-nv (approved):

packages/nvidia_nat_llama_index/src/nat/plugins/llama_index/llm.py
willkill07: I have hoisted out all patches (including the retry). The code was being duplicated across all functions, so I figured hoisting was the best approach.

PR #746, opened 2025-09-02
Update version of numpy to be more recent

zhongxuanwang-nv:
## Description

Closes [AIQ-1843](https://jirasw.nvidia.com/browse/AIQ-1843)

## UPDATE: 
- We can't update `numpy` to any newer versions. The log is below.
- As brought by @willkill07, since it's not recommended to use different numpy versions based on different python versions, and since we can't update langchain-aws because for python version <3.12, langchain-aws can only accept numpy<=1.26.*, so I will just close this PR and keep the existing version constraint, until langchain-aws might have an update on the version :)

# PREVIOUS:

Our repo does have many direct usage of `numpy`. This PR ~~will try to update the numpy to its latest version safely~~.

## Why I made the change
- `nvidia-nat-langchain` will pull in `langchain-aws`, in which if the python version is `<3.12`, the package requires `NumPy < 2.x`
- Besides that special case, I found we can just use the latest version of `numpy` (`>=2.3`, kept the version as latest as possible and kept the version as broad as possible) if python version is `>= 3.12`

## Tests conducted
- `uv lock` command is able to successfullly execute without any unresolved dependency conflicts.

The following targets were tested manually with the new numpy version
- All tests in `tests` folder
- All usage of numpy were isolated and double-checked with manually running the code snippet and checking for documentations noting potential change.
- `.../NeMo-Agent-Toolkit/src/nat/tool/code_execution/test_code_execution_sandbox.py` by spinning up a docker container first
    -  except for one test `TestCodeExecutionSandbox.test_plotly_import` is failing due to `ValueError: current limit exceeds maximum limit`  because I'm using MacOS. This test is not related to `numpy` version change

## Notes about `uv`
I'm using the latest version of `uv` per the "installing.md" guide that says `latest version [of uv] is recommended`, and the `uv.lock` file is generated using the command `uv lock`. However, I observed that in different versions of `uv`, the generated `uv.lock` file differ. For example, version `0.6.15` and above introduces the `upload-time` field, and running on different systems may introduce slight difference, so that's why in this PR there are so many lines of code change for `uv.lock` file

## Summary by CodeRabbit

- Documentation
  - Updated profiler workflow example to remove an unnecessary NumPy import, reflecting pandas-only usage. No behavioral changes.

- Chores
  - Adjusted NumPy dependency to be Python-version specific for improved compatibility:
    - Installs NumPy 1.26.x on Python < 3.12
    - Installs NumPy 2.3.x on Python ≥ 3.12
  - No changes to public APIs or functionality.


pyproject.toml
```diff
@@ -30,7 +30,8 @@ dependencies = [
   "jsonpath-ng~=1.7",
   "mcp~=1.13",
   "networkx~=3.4",
-  "numpy~=1.26",
+  "numpy~=1.26; python_version < '3.12'",
+  "numpy~=2.3; python_version >= '3.12'",
```
willkill07, line 35: We should not be listing two different major versions based on the python version in use. This could result in someone developing for 3.12 using new incompatible APIs and then being surprised that it cannot work in 3.11 with the older version of numpy.

Either:
1. upgrade unilaterally to `numpy~=2.3` (this will require a sane bump for `langchain-aws`)
2. keep the existing version constraint

zhongxuanwang-nv: Thank you for this insight!! Since we can't update `langchain-aws` because for python version `<3.12`, `langchain-aws` can only accept `numpy<=1.26.*`, so I will just close this PR and keep the existing version constraint.

pyproject.toml
```diff
@@ -30,7 +30,8 @@ dependencies = [
   "jsonpath-ng~=1.7",
   "mcp~=1.13",
   "networkx~=3.4",
-  "numpy~=1.26",
+  "numpy~=1.26; python_version < '3.12'",
```
willkill07, line 34: Let's add a comment above this line:
```
# langchain-aws pins numpy to 1.* for Python <3.12 -- remove once package is updated
```

PR #749, opened 2025-09-03
Support additional provider parameters in LLM and Embedder config

### LINKED ISSUE #671: OpenAIEmbedderModelConfig: Inability to pass tiktoken_enabled and other parameters

Improvement

Low (would be nice)

When creating an `OpenAIEmbeddings` client, we cannot natively pass parameters like `tiktoken_enabled`, `show_progress_bar`, or a custom `http_client` to handle specific needs. This is preventing us from integrating third-party models that do not use the tiktoken tokenizer, leading to conflicts.

The relevant class is located at: `aiq.embedder.openai_embedder`.

We would expect the `OpenAIEmbedderModelConfig` to support the passing of a wider range of parameters, or to have a dedicated mechanism for passing custom parameters directly to the underlying `OpenAIEmbeddings` client. This would allow for greater flexibility and broader compatibility.

### Additional context

As a workaround, we have extended the `OpenAIEmbedderModelConfig` class to include the necessary parameters and then passed them to the underlying client. This approach solves the immediate problem but requires a custom implementation that may become a maintenance issue.

1. Are there plans to modify the `OpenAIEmbedderModelConfig` to natively support additional parameters, which would make this workaround unnecessary?
2. Is our current method of extending the class considered the recommended best practice for this type of issue, or is there a better way to approach it?

### Code of Conduct

YosiElias:
## Description

Closes #671 

  ### Summary

  Adds dynamic validation support for additional OpenAI embedding parameters in `OpenAIEmbedderModelConfig`, enabling users to configure 20+ OpenAI-specific parameters through YAML while maintaining type safety and validation.

  ### Changes

  - **Enhanced Configuration**: Added `extra="allow"` and `@model_validator` to dynamically accept OpenAI parameters
  - **Dynamic Validation**: Uses `TypeAdapter` for full type validation and coercion against live OpenAI model fields
  - **20+ New Parameters**: Supports configurable parameters like `tiktoken_enabled`, `show_progress_bar`, `max_retries`, `chunk_size`, etc.
  - **Comprehensive Tests**: Added 15 test cases covering validation, type coercion, and error handling

  ### Benefits

  - ✅ **Always Current**: Automatically discovers parameters from OpenAI's implementation
  - ✅ **Type Safe**: Full Pydantic validation with automatic type coercion
  - ✅ **Clear Errors**: Descriptive validation messages for invalid parameters
  - ✅ **No Maintenance**: No static parameter lists to maintain

  ### Example

  ```yaml
  embedders:
    openai_embedder:
      _type: openai
      model_name: text-embedding-ada-002
      tiktoken_enabled: false        # Now supported
      show_progress_bar: true        # Now supported 
      max_retries: 5                # Now supported 
   ```

  This enhancement enables third-party model integration by supporting models that don't use tiktoken tokenizers through the `tiktoken_enabled` parameter, while allowing comprehensive configuration of all OpenAI parameters with dynamic validation.
  This is a general solution that can be applied to other LLM and embedding configuration
  classes throughout the codebase for consistent parameter handling, and I can implement it for other classes if needed.

## Summary by CodeRabbit

* New Features
  * Configurations for embedders and LLM providers now accept additional provider-specific options without validation errors.

* Improvements
  * Relaxed config validation across supported embedders and LLMs to allow extra keys, simplifying integration with advanced or evolving provider capabilities while preserving existing behavior for defined options.

### DIFF
--- a/src/nat/embedder/openai_embedder.py
+++ b/src/nat/embedder/openai_embedder.py
@@ -27,7 +27,7 @@ from nat.data_models.retry_mixin import RetryMixin
 class OpenAIEmbedderModelConfig(EmbedderBaseConfig, RetryMixin, name="openai"):
     """An OpenAI LLM provider to be used with an LLM client."""
 
-    model_config = ConfigDict(protected_namespaces=())
+    model_config = ConfigDict(protected_namespaces=(), extra="allow")
 
     api_key: str | None = Field(default=None, description="OpenAI API key to interact with hosted model.")
     base_url: str | None = Field(default=None, description="Base url to the hosted model.")
--- a/src/nat/embedder/nim_embedder.py
+++ b/src/nat/embedder/nim_embedder.py
@@ -50,7 +50,7 @@ class NIMEmbedderModelConfig(EmbedderBaseConfig, RetryMixin, name="nim"):
                                        description=("The truncation strategy if the input on the "
                                                     "server side if it's too large."))
 
-    model_config = ConfigDict(protected_namespaces=())
+    model_config = ConfigDict(protected_namespaces=(), extra="allow")
 
 
 @register_embedder_provider(config_type=NIMEmbedderModelConfig)
--- a/src/nat/llm/aws_bedrock_llm.py
+++ b/src/nat/llm/aws_bedrock_llm.py
@@ -30,7 +30,7 @@ from nat.data_models.top_p_mixin import TopPMixin
 class AWSBedrockModelConfig(LLMBaseConfig, RetryMixin, TemperatureMixin, TopPMixin, ThinkingMixin, name="aws_bedrock"):
     """An AWS Bedrock llm provider to be used with an LLM client."""
 
-    model_config = ConfigDict(protected_namespaces=())
+    model_config = ConfigDict(protected_namespaces=(), extra="allow")
 
     # Completion parameters
     model_name: str = Field(validation_alias=AliasChoices("model_name", "model"),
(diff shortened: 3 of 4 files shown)


src/nat/embedder/openai_embedder.py
```diff
@@ -27,13 +29,43 @@
...
+    @model_validator(mode='before')
+    def validate_extra_params(cls, values: dict[str, Any]) -> dict[str, Any]:
+        """Dynamically validate extra parameters with full field validation"""
+        defined_fields = set(cls.model_fields.keys())
+        extra_params = set(values.keys()) - defined_fields
+        
+        if not extra_params:
+            return values  # No extra params to validate
+        
+        from langchain_openai.embeddings.base import OpenAIEmbeddings
```
willkill07: You are assuming `langchain` when this class must also be compatible for use with other frameworks. Specifically, both LangChain and LlamaIndex have support for `OpenAIEmbedderModelConfig`. Whatever solution you have must be framework agnostic

YosiElias: You're totally right, thanks for the comment @willkill07 ! 
Given that we want to be compatible with a few frameworks, we can move the responsibility of validation to the original framework by setting `model_config.extra` to `allow`. 

Currently, it's set (by default) to `ignore`, which causes a problematic case where a user defines, for example, `tiktoken_enabled`, and this field isn't used. In some cases, this can be hard to debug and find what field was ignored. This approach keeps the overridden fields as they are and just allows more fields to be set, which will eventually be validated against the Pydantic class of the relevant framework.

LMK what you think and I'll update the PR.

willkill07: Let's definitely reduce the scope to just
```python
model_config = ConfigDict(protected_namespaces=(), extra="allow")
```
and remove this validation.

willkill07:
/ok to test [a03cc17](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/749/commits/a03cc179c897712a677f6145b097d173879ab3a2)

PR #764, opened 2025-09-05
Add test LLM provider to support testing

zhongxuanwang-nv:
## Description

Closes AIQ-439

* Added a deterministic testing LLM (nat_test_llm) in the nvidia_nat_test plugin to validate workflows without real API calls.  Supports a cycling `response_seq` and artificial latency via `delay_ms`.
* Made the testing LLM adaptable to all five core frameworks (`LangChain`, `LlamaIndex`, `CrewAI`, `SemanticKernel`, `Agno`)
* Added related tutorial and documents.
* Added the test LLM to the package loader (packages/nvidia_nat_test/src/nat/test/register.py).
* Added unit tests for `nat_test_llm` behavior (cycling, latency, and stream yielding a single chunk).
* Fixed a deprecation warning for `model_fields` and a small typo.

## Summary by CodeRabbit

- **New Features**
  - Added a deterministic test LLM provider for development/CI that cycles predefined responses with optional per-call delay and integrates with multiple LLM frameworks.

- **Documentation**
  - New tutorial and provider docs with install, YAML examples, CLI and programmatic usage, and links to LLM configuration guidance.

- **Tests**
  - Comprehensive cross-framework tests validating response cycling, delays, special characters, independent configs, and persistence.

- **Refactor**
  - Improved dependency-graph field handling.

- **Style**
  - Minor documentation typo fix.


zhongxuanwang-nv:
@coderabbitai review

docs/source/tutorials/testing-with-nat-test-llm.md
willkill07, line 1: It would also be really nice to have a cursor rule on writing tests which need to mock LLM behavior to use this

docs/source/workflows/llms/index.md
```diff
@@ -124,6 +124,33 @@ The Azure OpenAI LLM provider is defined by the {py:class}`~nat.llm.azure_openai
 * `seed` - The seed to use for the model
 * `max_retries` - The maximum number of retries for the request
 
+## Testing Provider
```
willkill07, line 131: Your additions need to go beneath the :::{note} about `temperature`

docs/source/workflows/llms/index.md
```diff
@@ -124,6 +124,33 @@ The Azure OpenAI LLM provider is defined by the {py:class}`~nat.llm.azure_openai
...
+workflow:
+  _type: chat_completion
+  llm_name: main
+  system_prompt: "Say only the answer."
+```
+
+* Learn how to add your own LLM provider: [Adding an LLM Provider](../../extend/adding-an-llm-provider.md)
+<!-- vale off -->
+* See a short tutorial using YAML and `nat_test_llm`: [Testing with nat_test_llm](../../tutorials/testing-with-nat-test-llm.md)
+<!-- vale on -->
```
willkill07, line 156: I'm curious what was flagged by vale here? If you think that it's a word that is valid, you can add it to the `accept.txt` file.

packages/nvidia_nat_test/src/nat/test/llm.py
```diff
@@ -0,0 +1,207 @@
...
+@register_llm_provider(config_type=TestLLMConfig)
+async def test_llm_provider(config: TestLLMConfig, builder: Builder):
+    del builder  # suppress linting error
+    yield LLMProviderInfo(config=config, description="Test LLM provider")
+
+
+@register_llm_client(config_type=TestLLMConfig, wrapper_type=LLMFrameworkEnum.LANGCHAIN)
+async def test_llm_langchain(config: TestLLMConfig, builder: Builder):
+    """LLM client for LangChain"""
+    from langchain_core.messages import AIMessage
```
willkill07: Given that we depend on a `langchain_core` import, `nvidia-nat-test` needs `nvidia-nat[langchain]` as a dependency

packages/nvidia_nat_test/src/nat/test/llm.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+@register_llm_provider(config_type=TestLLMConfig)
+async def test_llm_provider(config: TestLLMConfig, builder: Builder):
+    del builder  # suppress linting error
+    yield LLMProviderInfo(config=config, description="Test LLM provider")
+
+
+@register_llm_client(config_type=TestLLMConfig, wrapper_type=LLMFrameworkEnum.LANGCHAIN)
+async def test_llm_langchain(config: TestLLMConfig, builder: Builder):
```
willkill07, line 76: Always prefer prefixing with _ (e.g. `_unused` to `del unused`)
```suggestion
async def test_llm_langchain(config: TestLLMConfig, _builder: Builder):
```

packages/nvidia_nat_test/src/nat/test/llm.py
```diff
@@ -0,0 +1,207 @@
...
+            return chooser.next_response()
+
+    yield CrewAITestLLM()
+
+
+@register_llm_client(config_type=TestLLMConfig, wrapper_type=LLMFrameworkEnum.SEMANTIC_KERNEL)
+async def test_llm_semantic_kernel(config: TestLLMConfig, builder: Builder):
+    """LLM client for SemanticKernel"""
+    from semantic_kernel.contents.chat_message_content import ChatMessageContent
+    from semantic_kernel.contents.utils.author_role import AuthorRole
```
willkill07: Given that we depend on a `semantic_kernel` imports, `nvidia-nat-test` needs `nvidia-nat[semantic-kernel]` as a dependency

packages/nvidia_nat_test/src/nat/test/llm.py
```diff
@@ -0,0 +1,207 @@
...
+    class LIChatMessage:
+
+        def __init__(self, content: str) -> None:
+            self.content = content
+
+    class LIChatResponse:
+
+        def __init__(self, text: str) -> None:
+            self.message = LIChatMessage(text)
+            self.text = text
```
willkill07: It is valid to mock a response like this without using the "proper" Llama Index types?

packages/nvidia_nat_test/tests/test_test_llm.py
```diff
@@ -0,0 +1,389 @@
...
+    delay_ms: 0
+workflow:
+  _type: chat_completion
+  llm_name: main
+  system_prompt: "irrelevant"
+"""
+    file_a = tmp_path / "chat_completion_a.yml"
+    file_b = tmp_path / "chat_completion_b.yml"
+    file_a.write_text(yaml_a)
+    file_b.write_text(yaml_b)
```
willkill07, line 183: nit cleanup the files after -- should delete

packages/nvidia_nat_test/tests/test_test_llm.py
```diff
@@ -0,0 +1,389 @@
...
+
+    async with WorkflowBuilder() as builder:
+        cfg = TestLLMConfig(response_seq=list(seq), delay_ms=0)
+        await builder.add_llm("main", cfg)
+        client = await builder.get_llm("main", wrapper_type=wrapper)
+
+        outs: list[str] = []
+
+        if wrapper == LLMFrameworkEnum.LANGCHAIN.value:
+            from langchain_core.messages import AIMessage
```
willkill07: Shouldn't this also be an import or skip?

packages/nvidia_nat_test/tests/test_test_llm.py
willkill07, line 1: You are unit testing the test_llm really well here.

My only concern is how it interacts with the various frameworks (e.g. langchain, llama-index, semantic-kernel) for proper "integration" testing. Otherwise we may run into an issue where we attempt to use these e2e without the need for an API key.

Even if it were extremely primitive by including a known system prompt with known answers and having it validated for each framework for a basic e2e agent.

However, if you think this is out of scope for this PR and is intended for a follow-up PR that's also fine with me!

willkill07 (approved):
This is pretty cool. Just some minor feedback but otherwise LGTM.

docs/source/workflows/llms/index.md
zhongxuanwang-nv: It was for `nat_test_llm`! I took a look at `accept.txt` and I think since it appears that the entries there are primarily single-words (no words that have a underscore or dash, etc) so I guess I would keep the `vale` flag for now?

packages/nvidia_nat_test/src/nat/test/llm.py
zhongxuanwang-nv: Got it!! the only reason I didn't use prefix was because I feared the register would use `builder` as a keyword argument, so like `builder=xxx...`, but I guess since I already disabled linting for this file, I would just remove the prefix!

willkill07: We do pass positionally so the name of the formal parameter doesn't actually matter.

docs/source/tutorials/testing-with-nat-test-llm.md
zhongxuanwang-nv: Thank you Will! I thought about it and looked at all the existing cursor rules, I think since it's my first time writing a cursor rule and I'm debating on where should I put the new rules, I guess it's probably safer for me to not include cursor rules for now and do it in another PR?

packages/nvidia_nat_test/src/nat/test/llm.py
zhongxuanwang-nv: I think a great thing for `nvidia-nat-test` is they do depend on "`langchain-community~=0.3"`, which has `langchain-core` has a dependency (**[source](https://github.com/langchain-ai/langchain-community/blob/main/libs/community/pyproject.toml)**)! So I think we can safely not include nvidia-nat[langchain]?

packages/nvidia_nat_test/src/nat/test/llm.py
zhongxuanwang-nv: ~~I looked through the [LlamaIndex repo code related to this](https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/base/llms/types.py#L479) and confirmed this; I've also made sure to include a comment in the source code in case someone in the future comes across this suspicious code!~~

Actually, I think it's not super safe to do it here (they might change sth in the future). You are totally right in questioning this and I really appreciate it!! I will import the proper Llama Index types instead!

packages/nvidia_nat_test/src/nat/test/llm.py
zhongxuanwang-nv: Thanks Will for pointing this out! I've now included it in try-catch clause, and normally people that don't have the mentioned `wrapper_type` framework installed shouldn't access that particular part of the code.

packages/nvidia_nat_test/tests/test_test_llm.py
zhongxuanwang-nv: Thank you Will for pointing this out!! At this moment I've been going through deep into these documentations and manually check one by one, but I think there are definitely ways to design unit tests that can achieve the same thing. I would do so in another PR if possible!

PR #771, opened 2025-09-08
fix: include `thinking` in model_dump for `serve` config serialization

willkill07:
## Description
Originally, `thinking` (part of `ThinkingMixin`) had its field exposition set to False. This would result in any model_dump always omitting thinking. This was done originally because it made no sense to expose this.

However, it turns out that `nat serve` serializes the current configuration before passing it to a separate worker. To this end, we must expose `thinking` to ensure no loss of configuration data.

To accomodate this, model dumping for LLM instantiation must now exclude `thinking` explicitly. The logic has been unified across most LLM providers and frameworks to reduce the dancing between config objects created via `model_dump()` and the original LLM configs.

Closes

### DIFF
--- a/packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
+++ b/packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
@@ -117,7 +117,7 @@ async def azure_openai_langchain(llm_config: AzureOpenAIModelConfig, _builder: B
 
     from langchain_openai import AzureChatOpenAI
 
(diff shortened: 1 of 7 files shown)


packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
zhongxuanwang-nv, line 1: A small nit — but I think might be great to also exclude `thinking` for the `aws_bedrock_langchain` client?

PR #788, opened 2025-09-14
Fix NAT FastAPI front end with MCP Host using Stdio-MCP server fails to initialize

zhongxuanwang-nv:
## Description

Closes [AIQ-1853](https://jirasw.nvidia.com/browse/AIQ-1853)

In the past, when you do
```
nat serve --config_file examples/MCP/simple_calculator_mcp/configs/config-mcp-date-stdio.yml
```
You will get something like
```
    watcher = events.get_child_watcher()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/devcontainers/.local/share/uv/python/cpython-3.12.8-linux-x86_64-gnu/lib/python3.12/asyncio/events.py", line 828, in get_child_watcher
    return get_event_loop_policy().get_child_watcher()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/devcontainers/.local/share/uv/python/cpython-3.12.8-linux-x86_64-gnu/lib/python3.12/asyncio/events.py", line 645, in get_child_watcher
    raise NotImplementedError  
```
This is because when you start server with `uvicorn.run("nat.front_ends.fastapi.main:get_app",...`, `Uvicorn` by default uses `loop="auto"`, which prefers `uvloop` if it’s installed. 

However, on macOS with Python 3.12, uvloop’s event loop policy doesn’t provide a child watcher, so calls that spawn subprocesses (used by MCP stdio via asyncio) call `asyncio.events.get_child_watcher()` and raise NotImplementedError.

Therefore, we can fix it by setting loop="asyncio", which forces Uvicorn to use the standard asyncio event loop (which includes child-watcher support) when the system is MacOS. That enables asyncio.create_subprocess_exec to work, so the stdio MCP client can launch its subprocess cleanly.

## Follow up, why uvloop’s policy doesn’t provide a child watcher?
`uvloop` is a high‑performance event loop for Python’s asyncio, implemented on top of `libuv` (the C library used by Node.js). It’s a drop‑in replacement for the default asyncio loop and typically yields faster networking and scheduling.

`uvloop`’s loop policy design relies on libuv’s own process handling rather than Python’s legacy child‑watcher abstraction.

When code (e.g., anyio/asyncio paths) calls asyncio.get_event_loop_policy().get_child_watcher() under uvloop, the policy raises NotImplementedError because that API isn’t supported for MacOS with Python 3.12 and above.

### DIFF
--- a/src/nat/front_ends/fastapi/fastapi_front_end_plugin.py
+++ b/src/nat/front_ends/fastapi/fastapi_front_end_plugin.py
@@ -16,6 +16,7 @@
 import asyncio
 import logging
 import os
+import sys
 import tempfile
 import typing
 
@@ -153,13 +154,25 @@ class FastApiFrontEndPlugin(DaskClientMixin, FrontEndBase[FastApiFrontEndConfig]
 
                 reload_excludes = ["./.*"]
 
+                # By default, Uvicorn uses "auto" event loop policy, which prefers `uvloop` if installed. However,
+                # uvloop’s event loop policy for MacOS with Python 3.12 doesn’t provide a child watcher (which is needed
+                # for MCP server), so setting loop="asyncio" forces Uvicorn to use the standard event loop, which
+                # includes child-watcher support.
+                if sys.platform == "darwin":
+                    # For MacOS
+                    event_loop_policy = "asyncio"
+                else:
+                    # For non-MacOS platforms
+                    event_loop_policy = "auto"
+
                 uvicorn.run("nat.front_ends.fastapi.main:get_app",
                             host=self.front_end_config.host,
                             port=self.front_end_config.port,
                             workers=self.front_end_config.workers,
                             reload=self.front_end_config.reload,
                             factory=True,
-                            reload_excludes=reload_excludes)
+                            reload_excludes=reload_excludes,
+                            loop=event_loop_policy)
 
             else:
                 app = get_app()


src/nat/front_ends/fastapi/fastapi_front_end_plugin.py
```diff
@@ -153,13 +154,25 @@ async def run(self):
 
                 reload_excludes = ["./.*"]
 
+                # By default, Uvicorn uses "auto" event loop policy, which prefers `uvloop` if installed. However,
+                # uvloop’s event loop policy for macOS doesn’t provide a child watcher (which is needed for MCP server),
+                # so setting loop="asyncio" forces Uvicorn to use the standard event loop, which includes child-watcher
+                # support.
+                if sys.platform == "darwin":
```
AnuradhaKaruppiah: Nice find! 
This fix is specific to macos but we also need this to work on straight ubuntu/linux (ideally it should work the same way for all platforms).

Question: 
Is setting event_loop_policy to asyncio same as disabling "general.use_uvloop"  in the config file? If yes, we can document that as a limitation of stdio.

zhongxuanwang-nv: Thanks Anuradha!! Actually I apologize for not fixing it for `ubuntu/linux`, because from the description of the bug from `uvloop` I only saw complains about macOS, but I just tested on an ubuntu machine and it's not working there too :/ I'll fix it for ubuntu/linux too!

I scanned the codebase carefully manually and with AI, and I found there's actually neither internal nor external reference of the `use_uvloop`flag in the config file, so it's actually just a "dummy" flag for now. 

Therefore, I'm wondering should we consider removing it and smartly determine the `use_uvloop` based on the system, or should we add additional clarifications in the documentation and lets the users choose?

zhongxuanwang-nv: I'm nearly certainly sure that `uvloop` works with Windows, but I'll test it tonight after I go home where I can test it on the Windows system too!

AnuradhaKaruppiah: No worries! You have root cause so iterating from there should be easy.

I didn't know use_uvloop was dummy flag. Can you please confirm that with @dagardner-nv ?

zhongxuanwang-nv: I think David did agree that there's no direct reference of the `uv_loop`, I guess at the moment (as David suggested), I'll make another PR to deprecate the field, and also remove the reference of the `uv_loop` field in the config files, etc.

I'll merge the code for now (after adding `linux` too), because I'm almost certain `Windows` is supported too (will run a test later tonight)! Because from my quick research, it does turn out that `uvloop` is not usable with our python version and system version, even though it seems to give us a performance boost in some situations :/

Issue #798, opened 2025-09-16
`chat_completion` function returning `AIMessage` instead of `str`


zhongxuanwang-nv:
### Version

1.3.0

### Which installation method(s) does this occur on?

Source

### Describe the bug.

This happens even though `nat.llm/nim` is registered with `LangChain` framework.

### Minimum reproducible example

```shell
Put this to config.yaml

llms:
  main:
    _type: nat.llm/nim
    model_name: meta/llama-3.1-70b-instruct
    temperature: 0.0
workflow:
  _type: chat_completion
  llm_name: main
  system_prompt: "hi"


And run `nat run --config_file config.yaml --input "hey"`, you will get the log below:
```

### Relevant log output

<details><summary>Click here to see error details</summary><pre>

```
2025-09-16 10:59:26,525 - nat.builder.function - ERROR - Error with ainvoke in function with input: hey. Error: Cannot convert type <class 'langchain_core.messages.ai.AIMessage'> to <class 'str'>. No match found.
2025-09-16 10:59:26,530 - nat.runtime.runner - ERROR - Error running workflow: Cannot convert type <class 'langchain_core.messages.ai.AIMessage'> to <class 'str'>. No match found.
2025-09-16 10:59:26,533 - nat.cli.commands.start - ERROR - Failed to initialize workflow
Error: Cannot convert type <class 'langchain_core.messages.ai.AIMessage'> to <class 'str'>. No match found.
```
</pre></details>

### Other/Misc.

I'll work on it right now

### Code of Conduct

- [x] I agree to follow the NeMo Agent toolkit Code of Conduct
- [x] I have searched the [open bugs](https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

PR #802, opened 2025-09-16
Fixes `chat_completion` returning wrong type and substitute `.content` with `.text()`

zhongxuanwang-nv:
## Description

Closes [#798](https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues/798) 

- Fixed an issue that `chat_completion` will return the wrong type when using most LLMs with `LangChain` framework.
- Substituted `.content` with `.text()` in places where the object type is **obviously** either [AIMessage](https://python.langchain.com/api_reference/core/messages/langchain_core.messages.ai.AIMessage.html) or [BaseMessage](https://python.langchain.com/api_reference/core/messages/langchain_core.messages.base.BaseMessage.html#langchain_core.messages.base.BaseMessage.text), for more robust operation on getting the string.
  - To define "**obviously**", I made sure the LLM Framework Wrapper is `LangChain`, and only after that, I made sure I only change in places where `.content` was used

### DIFF
--- a/docs/source/extend/adding-an-llm-provider.md
+++ b/docs/source/extend/adding-an-llm-provider.md
@@ -211,9 +211,9 @@ async def test_nim_langchain_agent():
 
         response = await agent.ainvoke({"input": "What is 1+2?"})
         assert isinstance(response, AIMessage)
-        assert response.text() is not None
-        assert isinstance(response.text(), str)
-        assert "3" in response.text().lower()
+        assert response.content is not None
+        assert isinstance(response.content, str)
+        assert "3" in response.content.lower()
 ```
 
 Note: Since this test requires an API key, it's marked with `@pytest.mark.integration` to exclude it from CI runs. However, these tests are necessary for maintaining and verifying the functionality of LLM providers and their client integrations.
--- a/docs/source/extend/adding-an-llm-provider.md
+++ b/docs/source/extend/adding-an-llm-provider.md
@@ -211,9 +211,9 @@ async def test_nim_langchain_agent():
 
         response = await agent.ainvoke({"input": "What is 1+2?"})
         assert isinstance(response, AIMessage)
-        assert response.content is not None
-        assert isinstance(response.content, str)
-        assert "3" in response.content.lower()
+        assert response.text() is not None
+        assert isinstance(response.text(), str)
+        assert "3" in response.text().lower()
 ```
 
 Note: Since this test requires an API key, it's marked with `@pytest.mark.integration` to exclude it from CI runs. However, these tests are necessary for maintaining and verifying the functionality of LLM providers and their client integrations.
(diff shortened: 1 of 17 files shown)


docs/source/extend/adding-an-llm-provider.md
```diff
@@ -211,9 +211,9 @@ async def test_nim_langchain_agent():
 
         response = await agent.ainvoke({"input": "What is 1+2?"})
         assert isinstance(response, AIMessage)
-        assert response.content is not None
-        assert isinstance(response.content, str)
-        assert "3" in response.content.lower()
+        assert response.text() is not None
+        assert isinstance(response.text(), str)
+        assert "3" in response.text().lower()
```
willkill07: I think these were originally fine -- `text()` cannot return `None`

examples/advanced_agents/alert_triage_agent/src/nat_alert_triage_agent/utils.py
```diff
@@ -69,7 +69,7 @@ async def llm_ainvoke(config, builder, user_prompt, system_prompt=None):
         prompt = ChatPromptTemplate([MessagesPlaceholder("msgs")])
     chain = prompt | llm
     result = await chain.ainvoke({"msgs": [HumanMessage(content=user_prompt)]})
-    return result.content
+    return result.text()
```
willkill07, line 72: Is this function guaranteed to always return a string? `content` may be better here

willkill07 (commented):
I think overall we shouldn't greedily replace `content` with `text()` but I do think there are specific situations where it's okay to do so

examples/advanced_agents/alert_triage_agent/src/nat_alert_triage_agent/utils.py
zhongxuanwang-nv: Thank you Will! I think upon browsing all the usage of this function (from `src/nat_alert_triage_agent/telemetry_metrics_host_performance_check_tool.py` `telemetry_metrics_host_heartbeat_check_tool.py`, `network_connectivity_check_tool`, `monitoring_process_check_tool.py`, `host_performance_check_tool.py` and `hardware_check_tool.py`)

They all have a function signature that prescribes the output is `str` and only `str` (_except for `_host_performance_check_tool.py` that only says so in the function docstring but not in the signature, but I believe it was just a slight oversight before_) and they all just directly return the output of calling that  `llm_ainvoke` function, so I think it should be safe to change it like that! 

But I do agree this may cause confusion, so I'll make it explicit in the return type of this function!

docs/source/extend/adding-an-llm-provider.md
zhongxuanwang-nv: That's true! Thanks Will! I'll actually revert these changes, and will make sure I avoid testing things that will look obviously true (like this case)!

PR #803, opened 2025-09-16
Prevent retry storms in nested method calls

dnandakumar-nv:
Added `instance_context_aware` flag to the retry decorator to avoid retry storms caused by nested method calls. This ensures retries are not triggered recursively in the same instance's execution context. Additional tests were implemented to verify functionality across edge cases, including nested, async, deep nesting, and multi-instance scenarios.

## Description
Added `instance_context_aware` flag to the retry decorator to avoid retry storms caused by nested method calls. This ensures retries are not triggered recursively in the same instance's execution context. Additional tests were implemented to verify functionality across edge cases, including nested, async, deep nesting, and multi-instance scenarios.

### Summary
- Fixes exponential “retry storms” when a public method (with retries) calls other public methods (also wrapped with retries) on the same instance.
- Ensures, per instance, only the outermost call performs retries; nested calls execute once per outer attempt.
- Adds comprehensive tests for sync, async, deep nesting, multi-instance isolation, and exception propagation.

### Problem
- `patch_with_retry` wrapped every public method with exponential backoff.
- When method A calls method B (both wrapped), failures caused compounded retries (e.g., 3×3 = 9 calls), creating a retry storm and unexpected behavior.

### Approach
- Introduce an instance-level retry context to gate nested retries:
  - Add `instance_context_aware` to `_retry_decorator`.
  - When a wrapped method starts retries, set a `_in_retry_context` flag on the instance.
  - Nested calls on the same instance detect the flag and skip their own retry loops, letting the outermost call control retries.
  - Flag is set/unset with try/finally to avoid leaks; if setting the flag fails (immutable/slots), we safely no-op and continue without context tracking.
- Enable this behavior from `patch_with_retry` by default.

### Implementation
- File: `src/nat/utils/exception_handlers/automatic_retries.py`
  - `_retry_decorator(...)`:
    - New param: `instance_context_aware: bool = False`.
    - New helpers: `_check_retry_context`, `_set_retry_context`.
    - Apply context gating to sync, async, and (a)generator wrappers.
  - `patch_with_retry(...)`:
    - Pass `instance_context_aware=True`.
- Tests: `tests/nat/utils/test_retry_wrapper.py`
  - Added/updated:
    - `test_nested_retry_prevention`
    - `test_inner_method_can_still_retry_when_called_directly`
    - `test_deep_nesting`
    - `test_async_nested_retry_prevention`
    - `test_multiple_instances_dont_interfere`
    - `test_exception_propagation_in_nested_calls`
    - `test_retry_storm_prevention_with_all_methods_failing`

### Behavior Change
- Previously: nested public calls each executed their own retry loops (e.g., O(retries^depth)).
- Now: only the outermost call retries; nested calls run once per outer attempt (O(retries × depth)). Direct calls to methods still retry as configured.

### Backward Compatibility
- Public API unchanged.
- Behavior change only affects nested calls within the same instance; direct method calls retain original retry behavior.
- If any user relied on nested retries compounding, they’ll now see fewer total attempts (intended fix).

### Performance/Robustness
- Eliminates exponential blow-ups in requests and sleep delays.
- Reduces load, latency, and noise from nested backoff.

### Edge Cases
- If `_in_retry_context` cannot be set on an object (e.g., restricted attributes), we gracefully skip context tracking for that instance; behavior reverts to previous semantics for that object.

### Testing
- Covers sync, async, generator pathways.
- Verifies:
  - Nested calls don’t compound retries.
  - Direct calls still retry as configured.
  - Deep nesting behaves linearly with the outer retry count.
  - Multiple instances are isolated.
  - Exceptions propagate as expected.

### Files Changed
- `src/nat/utils/exception_handlers/automatic_retries.py`
- `tests/nat/utils/test_retry_wrapper.py`

### Checklist
- [x] Prevent retry storms via instance context
- [x] Sync/async/(a)gen parity
- [x] Unit tests added and adjusted
- [x] No public API changes
- [x] Comments and naming are clear

### Risk
- Low. Targeted change with comprehensive tests; default behavior remains intuitive and safer under nesting.

## Summary by CodeRabbit

- Improvements
  - Instance-aware retry handling to prevent redundant retries in nested calls (sync and async), improving reliability, performance, and reducing duplicate requests and rate-limit pressure.

- Bug Fixes
  - Eliminated retry storms when a retried method invokes other retried methods on the same object; outer retries proceed while inner calls execute once as intended.

- Tests
  - Comprehensive tests covering nested, deep, async, multi-instance, and failure-propagation scenarios to ensure correct behavior and stability.

### DIFF
--- a/src/nat/utils/exception_handlers/automatic_retries.py
+++ b/src/nat/utils/exception_handlers/automatic_retries.py
@@ -142,30 +142,42 @@ def _retry_decorator(
         use_deepcopy = deepcopy
         use_context_aware = instance_context_aware
 
-        def _check_retry_context(args) -> bool:
-            """Check if we're already in a retry context for this instance."""
-            if not use_context_aware or not args:
-                return False
-            obj = args[0]
-            return getattr(obj, '_in_retry_context', False)
+        class _RetryContext:
+            """Context manager for instance-level retry gating."""
 
-        def _set_retry_context(args, value: bool) -> None:
-            """Set the retry context flag on the instance."""
-            if use_context_aware and args:
-                obj = args[0]
+            __slots__ = ("_obj", "_enabled", "_active")
+
+            def __init__(self, args: tuple[Any, ...]):
+                self._obj = args[0] if (use_context_aware and args) else None
+                self._enabled = bool(self._obj)
+                self._active = False
+
+            def __enter__(self):
+                if not self._enabled:
+                    return False
                 try:
-                    object.__setattr__(obj, '_in_retry_context', value)
+                    # If already in retry context, signal caller to skip retries
+                    if getattr(self._obj, "_in_retry_context", False):
+                        return True
+                    object.__setattr__(self._obj, "_in_retry_context", True)
+                    self._active = True
+                    return False
                 except Exception:
-                    # If we can't set the attribute, just continue without context tracking
-                    pass
+                    # If we cannot set the attribute, behave as if context isn't enabled
+                    self._enabled = False
+                    return False
+
+            def __exit__(self, _exc_type, _exc, _tb):
+                if self._enabled and self._active:
+                    try:
+                        object.__setattr__(self._obj, "_in_retry_context", False)
+                    except Exception:
+                        pass
 
         async def _call_with_retry_async(*args, **kw) -> T:
-            # Skip retries if already in retry context
-            if _check_retry_context(args):
-                return await fn(*args, **kw)
-
-            _set_retry_context(args, True)
-            try:
+            with _RetryContext(args) as already_in_context:
+                if already_in_context:
+                    return await fn(*args, **kw)
                 delay = base_delay
                 for attempt in range(retries):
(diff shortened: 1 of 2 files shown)


src/nat/utils/exception_handlers/automatic_retries.py
```diff
@@ -130,69 +131,127 @@ def _retry_decorator(
...
+                    try:
+                        return await fn(*call_args, **call_kwargs)
+                    except retry_on as exc:
+                        if (not _want_retry(exc, code_patterns=retry_codes, msg_substrings=retry_on_messages)
+                                or attempt == retries - 1):
+                            raise
+                        await asyncio.sleep(delay)
+                        delay *= backoff
+            finally:
+                _set_retry_context(args, False)
```
willkill07: I know that is going to be annoying, but whenever I see a pattern like this, I think we should be using some sort of context manager so we don't need to wrap with `try...finally`. Instead it would be:

```python
with RetryContext(args):
    delay = base_delay
    for attempt in range(retries):
        ...
```

PR #817, opened 2025-09-18
Move router agent to `control_flow` category

yczhang-nv:
## Description

Move router agent to `control_flow` category to make it clearer that it is a control flow component, along with the newly added sequential executor. This PR added the missing documentation for sequential executor as well.

## Summary by CodeRabbit

* **Documentation**
  * Added a "Sequential Executor" workflow page with features, configuration, YAML examples, usage notes, and installation guidance.
  * Clarified Router Agent docs and removed the custom plugin mention.
  * Updated Workflow Types index links and relocated Router Agent example paths in guides.

* **Refactor**
  * Reorganized Router Agent components under the Control Flow area to align docs, examples, and registrations.


docs/source/workflows/about/sequential-executor.md
```diff
@@ -0,0 +1,156 @@
...
+  raise_type_incompatibility: false
+```
+
+### Configurable Options
+
+* `tool_list`: **Required**. A list of functions to execute sequentially. Each function's output becomes the input for the next function in the chain.
+
+* `raise_type_incompatibility`: Defaults to `False`. Whether to raise an exception if the type compatibility check fails. The type compatibility check runs before executing the tool list, based on the type annotations of the functions. When set to `True`, any incompatibility immediately raises an exception. When set to `False`, incompatibilities generate warning messages and the sequential executor continues execution. Set this to `False` when functions in the tool list include custom type converters, as the type compatibility check may fail even though the sequential executor can still execute the tool list.
+
+* `tool_execution_config`: Optional configuration for each tool in the sequential execution tool list. Keys should match the tool names from the `tool_list`.
```
willkill07: We should be more prescriptive with a mandate :)
```suggestion
* `tool_execution_config`: Optional configuration for each tool in the sequential execution tool list. Keys must match the tool names from the `tool_list`.
```

PR #834, opened 2025-09-22
feat!: Make function group filter functions async

willkill07:
### Breaking Changes
* All `get_*` functions of FunctionGroup are now `async`
* `Builder.get_tools` is now `async`
* `Builder.build` is now `async`

This PR ensures that all callbacks that enable filtering on function group functions should be marked async. Unfortunately, this results in `async` needing to propagate all the way up to `Builder.build()` (resulting in a breaking change).

This is worthwhile due to the asynchronous nature of filtering functions (e.g. Authentication, Dynamic Checks via endpoint queries).

Closes


zhongxuanwang-nv (approved):
Thanks Will! I think this looks great!

PR #840, opened 2025-09-23
Improve robustness of MCP client remote tool calling 

yczhang-nv:
## Description

Adds configurable tool-call timeouts and backoff-based reconnect logic to MCP clients, propagates these settings through transport constructors and client config, enforces parent linkage for tool clients, updates lifecycle and error-handling flows, and adds tests covering reconnect/backoff, concurrency, and timeout propagation.

Mainly focused on fixing the following scenarios, the MCP client should not hang but return comprehensive information
- MCP server is not available (MCP client should try to re-connect before returning error information)
- MCP server relaunched but missing some tools previously registered with the MCP client, and MCP client is calling that missing tool
- MCP server takes too long to execute a tool

Closes AIQ-1942

## Summary by CodeRabbit

- New Features
  - Added configurable automatic reconnect (enable, max attempts, initial/max backoff) and per-call tool timeouts; settings propagate across transports and to tool clients.

- Refactor
  - Improved lifecycle and connection orchestration with single-retry wrapper, reconnection backoff, session state handling, and clearer error formatting for tool calls.

- Tests
  - Added extensive tests for reconnect scenarios, backoff timing, concurrency safety, timeout propagation, and lifecycle/state transitions.


yczhang-nv:
@coderabbitai full review

packages/nvidia_nat_mcp/src/nat/plugins/mcp/client_base.py
```diff
@@ -415,22 +536,24 @@ async def acall(self, tool_args: dict) -> str:
...
+                    # Log non-text content for now
+                    logger.warning("Got not-text output from %s of type %s", self.name, type(res))
+            result_str = "\n".join(output)
+
+            if result.isError:
+                raise RuntimeError(result_str)
+
+        except MCPError as e:
+            format_mcp_error(e, include_traceback=False)
+            result_str = "Tool call failed: %s" % e.original_exception
```
zhongxuanwang-nv: Thanks Yuchen!! A quick question: on line 552, seems that if it fails it would raise a `RuntimeError` which would break the code, but line 555 seems to hope to capture the error instead of breaking the code, so I'm wondering should line 553 be `raise MCPError(result_str)` instead?

packages/nvidia_nat_mcp/src/nat/plugins/mcp/client_base.py
```diff
@@ -181,30 +201,88 @@ def server_name(self):
...
+        """
+        Attempt to reconnect by tearing down and re-establishing the session.
+        """
+        async with self._reconnect_lock:
+            backoff = self._reconnect_initial_backoff
+            attempt = 0
+            last_error: Exception | None = None
+
+            while attempt < self._reconnect_max_attempts:
+                attempt += 1
```
willkill07, line 222: Nit: prefer a range loop
```suggestion
            last_error: Exception | None = None

            for attempt in range(1, self._reconnect_max_attempts + 1):
```

packages/nvidia_nat_mcp/src/nat/plugins/mcp/client_base.py
```diff
@@ -371,15 +487,20 @@ class MCPToolClient:
 
     def __init__(self,
                  session: ClientSession,
+                 parent_client: "MCPBaseClient",
                  tool_name: str,
                  tool_description: str | None,
                  tool_input_schema: dict | None = None,
-                 parent_client: "MCPBaseClient | None" = None):
+                 tool_call_timeout: timedelta = timedelta(seconds=5)):
```
willkill07, line 497: If you provide a default in the pydantic config, perhaps the default isn't necessary here?

AnuradhaKaruppiah:
Looks great to me! Nicely done @yczhang-nv

packages/nvidia_nat_mcp/src/nat/plugins/mcp/client_base.py
yczhang-nv: Good catch Daniel!

packages/nvidia_nat_mcp/src/nat/plugins/mcp/client_base.py
yczhang-nv: Correct, it is not necessary. I think it is not a bad idea to ensure there's always a default value here, since the Python MCP SDK defaults to wait forever if it is not provided with a timeout value, and it caused one of the issue that the PR wants to solve.

PR #853, opened 2025-09-26
Add `mcp/client/tool/list` endpoint

yczhang-nv:
## Description

Closes AIQ-1943

## Summary by CodeRabbit

- **New Features**
  - Added an MCP client tool-list HTTP endpoint showing per-group tool availability, session health, and workflow alignment.
  - CLI now awaits asynchronously produced function mappings when listing functions.

- **Bug Fixes**
  - Added timeout handling when fetching MCP tools to avoid hangs and surface timeouts.
  - Skips invalid/missing tool functions and normalizes input schemas to prevent wiring failures.

- **Documentation**
  - Added docs for the new HTTP tool-list endpoint with usage and sample output.

- **Tests**
  - End-to-end tests for healthy/unhealthy MCP tool-list scenarios; updated exception expectations.


packages/nvidia_nat_mcp/src/nat/plugins/mcp/client_impl.py
```diff
@@ -179,6 +179,13 @@ async def mcp_client_function_group(config: MCPClientConfig, _builder: Builder):
     group = FunctionGroup(config=config)
 
     async with client:
+        # Expose the live MCP client on the function group instance so other components (e.g., HTTP endpoints)
+        # can reuse the already-established session instead of creating a new client per request.
+        # These attributes are intentionally internal/private to avoid API surface commitments.
+        setattr(group, "_mcp_client", client)
+        setattr(group, "_mcp_client_server_name", client.server_name)
+        setattr(group, "_mcp_client_transport", client.transport)
+
```
willkill07, line 231: I love that we are using Function Groups :)

I do have an idea though... what if we had an `MCPFunctionGroup` whose base class was `FunctionGroup`? That way all extra members would exist safely with proper types/safety.

Seeing `getattr` and `setattr` in multiple source files seems like a code smell to me.

src/nat/front_ends/fastapi/fastapi_front_end_plugin_worker.py
```diff
@@ -1088,6 +1089,168 @@ async def redirect_uri(request: Request):
...
+
+                            # Get workflow function group configuration (configured client-side tools)
+                            configured_short_names: set[str] = set()
+                            configured_full_to_fn: dict[str, Any] = {}
+                            try:
+                                accessible_functions = await group_instance.get_accessible_functions()
+                                configured_full_to_fn = accessible_functions
+                                configured_short_names = {
+                                    name.split('.', 1)[1] if '.' in name else name
+                                    for name in accessible_functions.keys()
```
willkill07: Function groups always return qualified names e.g. `group_name.function_name` so the `if '.' in name else name` is unnecessary

src/nat/front_ends/fastapi/fastapi_front_end_plugin_worker.py
```diff
@@ -1088,6 +1089,168 @@ async def redirect_uri(request: Request):
...
+                                session_healthy = True
+                            except Exception as e:
+                                logger.warning(f"Failed to connect to MCP server {client.server_name}: {e}")
+                                session_healthy = False
+
+                            # Get workflow function group configuration (configured client-side tools)
+                            configured_short_names: set[str] = set()
+                            configured_full_to_fn: dict[str, Any] = {}
+                            try:
+                                accessible_functions = await group_instance.get_accessible_functions()
```
willkill07: I'm not sure of the right approach here, but I was envisioning for mcp function groups, they would all have a default filter function which tests the health :)

If that's the case, this would result in infinite recursion. Maybe we want to explicitly pass `filter_fn=lambda x:x` to avoid this scenario?

src/nat/front_ends/fastapi/fastapi_front_end_plugin_worker.py
```diff
@@ -1088,6 +1089,168 @@ async def redirect_uri(request: Request):
...
+            """
+            mcp_clients_info = []
+
+            try:
+                # Get all function groups from the builder
+                function_groups = builder._function_groups
+
+                # Find MCP client function groups
+                for group_name, configured_group in function_groups.items():
+                    if configured_group.config.type == "mcp_client":
```
willkill07: Nit: to reduce extra indentation, having an early-continue path helps:
```python
if configured_group.config.type != "mcp_client":
    continue
```

src/nat/front_ends/fastapi/fastapi_front_end_plugin_worker.py
```diff
@@ -1088,6 +1089,168 @@ async def redirect_uri(request: Request):
...
+                                logger.warning(f"Failed to get accessible functions for group {group_name}: {e}")
+
+                            # Build alias->original mapping from overrides
+                            alias_to_original: dict[str, str] = {}
+                            try:
+                                overrides = getattr(config, "tool_overrides", None) or {}
+                                for orig_name, override in overrides.items():
+                                    alias = getattr(override, "alias", None)
+                                    if alias:
+                                        alias_to_original[alias] = orig_name
```
willkill07: Defaulting to a dict means that overrides is no longer type-safe.
```suggestion
                                if config.tool_overrides is not None:
                                    for orig_name, override in overrides.items():
                                        if override.alias is not None:
                                            alias_to_original[override.alias] = orig_name
```

src/nat/front_ends/fastapi/fastapi_front_end_plugin_worker.py
```diff
@@ -1088,6 +1089,168 @@ async def redirect_uri(request: Request):
...
+                            try:
+                                server_tools = await client.get_tools()
+                                session_healthy = True
+                            except Exception as e:
+                                logger.warning(f"Failed to connect to MCP server {client.server_name}: {e}")
+                                session_healthy = False
+
+                            # Get workflow function group configuration (configured client-side tools)
+                            configured_short_names: set[str] = set()
+                            configured_full_to_fn: dict[str, Any] = {}
```
willkill07:
```suggestion
                            configured_full_to_fn: dict[str, Function] = {}
```

src/nat/front_ends/fastapi/fastapi_front_end_plugin_worker.py
```diff
@@ -1088,6 +1089,168 @@ async def redirect_uri(request: Request):
...
+                                available = session_healthy and (orig_name in server_tools)
+                                if available:
+                                    available_count += 1
+
+                                # Prefer the workflow function description (includes overrides)
+                                full_name = f"{group_name}.{fn_short}"
+                                wf_fn = configured_full_to_fn.get(full_name)
+                                description = getattr(
+                                    wf_fn, "description",
+                                    None) or (server_tools[orig_name].description if available else "")
```
willkill07: wf_fn is definitely a `Function` type and will always have a description member.

What I don't understand is the logic chain here...

1. description is always going to be the description of the function within the function group if it exists.
2. If it doesn't exist then default to the description within server_tools ONLY if it's available
3. else it will be an empty string

I would think the logic would be:
```python
(server_tools[orig_name].description if available else None) or wf_fn.description or ""
```
Since we'd want to prefer the server_tools description over the function's description, and only fall back to empty string if still `None`

src/nat/front_ends/fastapi/fastapi_front_end_plugin_worker.py
```diff
@@ -1088,6 +1089,168 @@ async def redirect_uri(request: Request):
...
+                                    alias = getattr(override, "alias", None)
+                                    if alias:
+                                        alias_to_original[alias] = orig_name
+                            except Exception:
+                                pass
+
+                            # Create tool info list (always return configured tools; mark availability)
+                            tools_info: list[dict[str, Any]] = []
+                            available_count = 0
+                            for fn_short in sorted(configured_short_names):
```
willkill07: Is there a reason we want this to be sorted here? could we not sort the list afterwards? I ask because if the `configured_short_names` are always in the same order as `configured_full_to_fn`, then you could just `zip` them here so you don't have to do the name reconstruction or lookup

src/nat/front_ends/fastapi/fastapi_front_end_plugin_worker.py
yczhang-nv: Thanks for providing that insight : )

willkill07 (approved):

PR #862, opened 2025-09-27
Implement OAuth2 security test coverage

ericevans-nv:
## Description

This PR adds comprehensive security testing for OAuth2 authorization code flows and enhances token validation coverage. The implementation includes parameterized tests for redirect URI validation, authorization code security (one-time use, expiration, client binding), and state parameter handling to prevent CSRF attacks. Additionally, it extends bearer token validation with expiration testing for both JWT and opaque tokens, and implements authentication log suppression to prevent sensitive credential leakage. The test suite covers critical OAuth2 security requirements including PKCE validation, error response handling, and proper authentication flow component integration across both console and WebSocket handlers.

### DIFF
--- a/src/nat/front_ends/console/authentication_flow_handler.py
+++ b/src/nat/front_ends/console/authentication_flow_handler.py
@@ -14,14 +14,17 @@
 # limitations under the License.
 
 import asyncio
+import logging
 import secrets
 import webbrowser
 from dataclasses import dataclass
 from dataclasses import field
 
 import click
+import httpx
 import pkce
 from authlib.integrations.httpx_client import AsyncOAuth2Client
+from authlib.common.errors import AuthlibBaseError as OAuthError
 from fastapi import FastAPI
 from fastapi import Request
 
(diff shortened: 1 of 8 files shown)


src/nat/front_ends/console/authentication_flow_handler.py
```diff
@@ -150,13 +190,16 @@ async def _handle_oauth2_auth_code_flow(self, cfg: OAuth2AuthCodeFlowProviderCon
             self._active_flows += 1
 
         click.echo("Your browser has been opened for authentication.")
-        webbrowser.open(auth_url)
+        try:
+            webbrowser.open(auth_url)
```
willkill07, line 194: This is me asking for understanding:

In the general case here, does `auth_url` get printed to the end-user so they have a chance to copy-paste somewhere else in case `webbrowser.open()` fails?

If not, I feel like waiting for the redirect would be impossible and we should fast-fail

ericevans-nv: No, unfortunately the URL contains sensitive credentials, so we cannot log it.

PR #863, opened 2025-09-28
Add hierarchical IDs for consistent telemetry and reporting

zhongxuanwang-nv:
## Description

Closes 1897

**Added**:
- `ContextState` fields: `workflow_run_id`, `workflow_trace_id` (`src/nat/builder/context.py`)
- Non-zero ID generators & validators for OpenTelemetry-sized IDs (`src/nat/data_models/span.py`)
- Workflow start/end event emission with metadata: `workflow_run_id`, `workflow_trace_id`, `conversation_id` (`src/nat/runtime/runner.py`)
- Span attributes for correlation: `.workflow.run_id`, `.workflow.trace_id`, `.conversation.id` (`src/nat/observability/exporter/span_exporter.py`)
- Tests: `tests/nat/opentelemetry/test_otel_span_ids.py`, `tests/nat/runtime/test_runner_trace_ids.py`, `tests/nat/runtime/test_session_traceparent.py`, updates to `tests/nat/runtime/test_runner.py`

**Changed**:
- Ensure non-zero 128-bit trace IDs and 64-bit span IDs in OTEL plugin (`packages/nvidia_nat_opentelemetry/src/nat/plugins/opentelemetry/otel_span.py`)
- Runner establishes workflow-scoped run/trace IDs and resets them on exit (`src/nat/runtime/runner.py`)
- Exporter prefers parent/context trace ID; falls back to workflow trace ID and sets correlated attributes (`src/nat/observability/exporter/span_exporter.py`)
- `SpanContext` defaults & field validators enforce size/range and non-zero invariants (`src/nat/data_models/span.py`)
- Parse/propagate incoming W3C `traceparent` and workflow headers to session/runner (`src/nat/runtime/session.py`, `tests/nat/runtime/test_session_traceparent.py`)

**Removed**:
- None

**Why it matters**:
- Consistent cross-span correlation and easier log/trace joins across agents and sub-spans
- Fewer broken traces: eliminates zero/invalid IDs, aligns with OpenTelemetry sizing
- Improves debugging and fleet-wide reporting via explicit `.workflow.*` and `.conversation.id` attributes
- Safer header handling for external callers; preserves upstream traces

**Breaking changes / Migrations**:
- None

**Docs**:
- None

## Summary by CodeRabbit

* **New Features**
  * Workflow-level run and trace IDs are created, stored in context, propagated, and emitted with WORKFLOW_START/WORKFLOW_END events.
  * Incoming HTTP tracing headers (traceparent, workflow-trace-id, workflow-run-id) are parsed into context.

* **Improvements**
  * Span metadata now includes workflow IDs and richer attributes (event type, function info, timestamps, conversation/run IDs).
  * ID generation and validation tightened to ensure non-zero, correctly sized trace and span IDs; span context ID generation approach updated.

* **Tests**
  * Added tests validating span ID formats, runner trace/run ID behavior, and header parsing.


zhongxuanwang-nv:
@coderabbitai  review

zhongxuanwang-nv:
/ok to review [667efdb](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/863/commits/667efdb4d0c1219bba406b1beb54d07d73bed3c9)

packages/nvidia_nat_opentelemetry/src/nat/plugins/opentelemetry/otel_span.py
```diff
@@ -86,8 +86,15 @@ def __init__(
...
-            span_id = uuid.uuid4().int & ((1 << 64) - 1)
+            # Generate non-zero IDs per OTel spec
+            while True:
+                trace_id = uuid.uuid4().int & ((1 << 128) - 1)
+                if trace_id != 0:
+                    break
+            while True:
+                span_id = uuid.uuid4().int & ((1 << 64) - 1)
+                if span_id != 0:
+                    break
```
willkill07: UUID v4 have the following properties:
- always 128-bit
- always have some bits set to indicate the UUID version
- the upper 64-bits are guaranteed to be non-zero (since one nibble always has a value of `4`)
- the lower 64-bits are guaranteed to be non-zero (since three bits are used to indicate a UUIDv4 variant)


```suggestion
            # Generate non-zero IDs per OTel spec
            trace_id = uuid.uuid4().int
            span_id = uuid.uuid4().int >> 64
```

src/nat/data_models/span.py
```diff
@@ -128,10 +128,57 @@ class SpanStatus(BaseModel):
     message: str | None = Field(default=None, description="The status message of the span.")
 
 
+def _generate_nonzero_trace_id() -> int:
+    """Generate a non-zero 128-bit trace ID."""
+    while True:
+        value = uuid.uuid4().int & ((1 << 128) - 1)
+        if value != 0:
+            return value
```
willkill07:
```suggestion
    return uuid.uuid4().int
```

src/nat/data_models/span.py
```diff
@@ -128,10 +128,57 @@ class SpanStatus(BaseModel):
...
+
+
+def _generate_nonzero_span_id() -> int:
+    """Generate a non-zero 64-bit span ID."""
+    while True:
+        value = uuid.uuid4().int & ((1 << 64) - 1)
+        if value != 0:
+            return value
```
willkill07:
```suggestion
    return uuid.uuid4().int >> 64
```

src/nat/data_models/span.py
```diff
@@ -128,10 +128,57 @@ class SpanStatus(BaseModel):
...
+    @field_validator("trace_id", mode="before")
+    @classmethod
+    def _validate_trace_id(cls, v: int | str | None) -> int:
+        """Ensure trace_id is within 128-bit range and non-zero; regenerate if invalid."""
+        if v is None:
+            return _generate_nonzero_trace_id()
+        try:
+            value = int(v)
+        except Exception:
+            return _generate_nonzero_trace_id()
```
willkill07: This seems dangerous:
1. What if the `str` has the form of a UUID? e.g. `8D8AC610-566D-4EF0-9C22-186B2A5ED793` (or lower-case / without dashes)
2. Why fall back to generating a non-zero trace ID in case of failure. It should most definitely fail / raise an error

src/nat/data_models/span.py
```diff
@@ -128,10 +128,57 @@ class SpanStatus(BaseModel):
...
+        value = value & ((1 << 128) - 1)
+        if value == 0:
+            return _generate_nonzero_trace_id()
+        return value
```
willkill07:
```suggestion
        if isinstance(v, str):
            v = uuid.UUID(v).int
        if isinstance(v, type(None)):
            v = _generate_nonzero_trace_id()
        if v == 0:
            raise Exception("A null UUID is invalid")
        return v
```

src/nat/data_models/span.py
```diff
@@ -128,10 +128,57 @@ class SpanStatus(BaseModel):
...
+
+    @field_validator("span_id", mode="before")
+    @classmethod
+    def _validate_span_id(cls, v: int | str | None) -> int:
+        """Ensure span_id is within 64-bit range and non-zero; regenerate if invalid."""
```
willkill07: Similar logic to above can be done here. Though this has me question why span_ids are only 64-bit and not 128-bit

src/nat/runtime/runner.py
```diff
@@ -130,17 +135,65 @@ async def result(self, to_type: type | None = None):
...
+            workflow_run_id = existing_run_id if existing_run_id else str(uuid.uuid4())
+
+            if existing_trace_id:
+                workflow_trace_id = existing_trace_id
+            else:
+                # Ensure non-zero 128-bit workflow trace id
+                while True:
+                    workflow_trace_id = uuid.uuid4().int & ((1 << 128) - 1)
+                    if workflow_trace_id != 0:
+                        break
```
willkill07:
```suggestion
            workflow_trace_id = existing_trace_id or uuid.uuid4().int
```

src/nat/runtime/runner.py
```diff
@@ -155,25 +208,77 @@ async def result(self, to_type: type | None = None):
...
             self._state = RunnerState.RUNNING
 
             if (not self._entry_fn.has_streaming_output):
                 raise ValueError("Workflow does not support streaming output")
 
+            # Establish workflow run and trace identifiers
+            existing_run_id = self._context_state.workflow_run_id.get()
+            existing_trace_id = self._context_state.workflow_trace_id.get()
+
+            workflow_run_id = existing_run_id if existing_run_id else str(uuid.uuid4())
```
willkill07:
```suggestion
            workflow_run_id = existing_run_id or str(uuid.uuid4())
```

src/nat/runtime/runner.py
```diff
@@ -155,25 +208,77 @@ async def result(self, to_type: type | None = None):
...
+
+            if existing_trace_id:
+                workflow_trace_id = existing_trace_id
+            else:
+                # Ensure non-zero 128-bit workflow trace id
+                while True:
+                    workflow_trace_id = uuid.uuid4().int & ((1 << 128) - 1)
+                    if workflow_trace_id != 0:
+                        break
```
willkill07:
```suggestion
            workflow_trace_id = existing_trace_id or uuid.uuid4().int
```

willkill07 (changes requested):
I know too much about UUIDs :)
This is nice, but we can clean it up a lot!

src/nat/data_models/span.py
zhongxuanwang-nv: Thank you Will for helping me to simplify a lot of these code!! I think it's a convention that `span_id` in these tracing systems (OTel, etc) are 64 bits, and `trace_id` is 128 bits; and this exists primarily because a `trace_id` may have thoughts of `span_id`, and to access a `span_id` you should at least use a `trace_id`, so I think for the sake of saving storage, `span_id` is shorter, and also the fact that since `trace_id` has pretty much guaranteed non-collision, `span_id` would not be super long.

Some sources: https://www.w3.org/TR/trace-context/?utm_source=chatgpt.com#parent-id

src/nat/data_models/span.py
zhongxuanwang-nv: That is true!! Here I accidentally forgot to convert it to the 128-bit int form.. So it would most of the time fail and regenerate a new one :/ Thank you so much Will for catching this!!

zhongxuanwang-nv:
/ok to test [3a2186b](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/863/commits/3a2186b27a2967c71deab807394fe721dcdebb55)

src/nat/data_models/span.py
```diff
@@ -128,10 +128,45 @@ class SpanStatus(BaseModel):
...
+        if v == 0:
+            raise Exception("A null UUID is invalid")
+        return v
+
+    @field_validator("span_id", mode="before")
+    @classmethod
+    def _validate_span_id(cls, v: int | str | None) -> int:
+        """Regenerate if span_id is None; raise an exception if span_id is invalid;"""
+        if isinstance(v, str):
+            v = uuid.UUID(v).int
```
willkill07: Since span is 64-bit, we cannot rely on it being a valid UUID :(

Doing `int(v)` is enough here IMO

src/nat/data_models/span.py
```diff
@@ -128,10 +128,45 @@ class SpanStatus(BaseModel):
...
+        if isinstance(v, type(None)):
+            v = _generate_nonzero_span_id()
+        if v == 0:
+            raise Exception("A null UUID is invalid")
+        return v
```
willkill07, line 172: Also add a check `if (v >> 64)` (e.g. beyond 64-bit)

src/nat/runtime/session.py
```diff
@@ -161,6 +161,31 @@ def set_metadata_from_http_request(self, request: Request) -> None:
...
 
+        # W3C Trace Context header: traceparent: 00-<trace-id>-<span-id>-<flags>
+        traceparent = request.headers.get("traceparent")
+        if traceparent:
+            try:
+                parts = traceparent.split("-")
+                if len(parts) >= 4:
+                    trace_id_hex = parts[1]
+                    if len(trace_id_hex) == 32:
+                        trace_id_int = int(trace_id_hex, 16)
```
willkill07: you could just attempt to parse `uuid.UUID(trace_id_hex).int` :)

src/nat/runtime/session.py
```diff
@@ -161,6 +161,31 @@ def set_metadata_from_http_request(self, request: Request) -> None:
...
+                        self._context_state.workflow_trace_id.set(trace_id_int)
+            except Exception:
+                pass
+
+        if not self._context_state.workflow_trace_id.get():
+            workflow_trace_id = request.headers.get("workflow-trace-id")
+            if workflow_trace_id:
+                try:
+                    self._context_state.workflow_trace_id.set(int(workflow_trace_id, 16))
```
willkill07: Same here... rely on the fact that it should be a UUID

tests/nat/runtime/test_session_traceparent.py
```diff
@@ -0,0 +1,64 @@
...
+class _DummyWorkflow:
+    config = None
+
+
+@pytest.mark.parametrize("header_case", [
+    "traceparent",
+    "workflow-trace-id",
+])
+@pytest.mark.asyncio
+async def test_session_sets_trace_id_from_headers(header_case: str):
```
willkill07: Maybe also test error cases?

src/nat/data_models/span.py
zhongxuanwang-nv: That's true!! `uuid.UUID(sth)` will yield a 128-bit value... Thank you so much Will for catching this!!

src/nat/data_models/span.py
```diff
@@ -128,10 +128,48 @@ class SpanStatus(BaseModel):
...
+
+    @field_validator("trace_id", mode="before")
+    @classmethod
+    def _validate_trace_id(cls, v: int | str | None) -> int:
+        """Regenerate if trace_id is None; raise an exception if trace_id is invalid;"""
+        if isinstance(v, str):
+            v = uuid.UUID(v).int
+        if isinstance(v, type(None)):
+            v = _generate_nonzero_trace_id()
+        if v == 0:
```
willkill07:
```suggestion
        if v <= 0 or v >> 128:
```

src/nat/data_models/span.py
```diff
@@ -128,10 +128,48 @@ class SpanStatus(BaseModel):
...
+    def _validate_span_id(cls, v: int | str | None) -> int:
+        """Regenerate if span_id is None; raise an exception if span_id is invalid;"""
+        if isinstance(v, str):
+            try:
+                v = int(v, base=0)
+            except ValueError:
+                raise Exception("span_id is unable to be parsed")
+        if isinstance(v, type(None)):
+            v = _generate_nonzero_span_id()
+        if (v == 0) or (v >> 64) or (v < 0):
```
willkill07:
```suggestion
        if v <= 0 or v >> 64:
```

tests/nat/runtime/test_session_traceparent.py
```diff
@@ -0,0 +1,159 @@
...
+    ([(b"workflow-trace-id", ("f" * 33).encode())], None),
+]
+
+
+@pytest.mark.parametrize(
+    "headers,expected_trace_id",
+    TRACE_ID_CASES,
+)
+@pytest.mark.asyncio
+async def test_session_trace_id_from_headers_parametrized(headers: list[tuple[bytes, bytes]],
```
willkill07:
```suggestion
async def test_session_trace_id_from_headers_parameterized(headers: list[tuple[bytes, bytes]],
```

tests/nat/runtime/test_session_traceparent.py
```diff
@@ -0,0 +1,159 @@
...
+    ),
+]
+
+
+@pytest.mark.parametrize(
+    "headers,expected_conv,expected_msg,expected_run",
+    METADATA_CASES,
+)
+@pytest.mark.asyncio
+async def test_session_metadata_headers_parametrized(headers: list[tuple[bytes, bytes]],
```
willkill07:
```suggestion
async def test_session_metadata_headers_parameterized(headers: list[tuple[bytes, bytes]],
```

src/nat/data_models/span.py
```diff
@@ -128,10 +128,48 @@ class SpanStatus(BaseModel):
...
+            raise Exception("span_id is invalid")
+        return v
```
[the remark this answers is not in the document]
willkill07: > base=0 allows prefix interpretation: int(v, base=0) accepts 0x, 0o, 0b prefixes—is this intentional for span IDs?

This is a good review comment. I'm unsure of your expected behavior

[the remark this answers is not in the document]
zhongxuanwang-nv: > > base=0 allows prefix interpretation: int(v, base=0) accepts 0x, 0o, 0b prefixes—is this intentional for span IDs?
> 
> This is a good review comment. I'm unsure of your expected behavior

@willkill07 Thank you Will!! I was thinking putting `base=0` will ask python to infer the base from the string, but indeed, the string must've started with `0x`, `0b`, etc, in which this case doesn't guarantee.. And this indeed can become extra..

PR #899, opened 2025-10-02
chore: additional workflow template cleanup

willkill07:
## Description
Cleans up the default configuration + workflow template a little more

## Summary by CodeRabbit

- New Features
  - None
- Refactor
  - Simplified generated workflow configuration by removing redundant logging and frontend blocks and a deprecated retry option.
  - Aligned workflow registration naming to a consistent convention.
- Documentation
  - Added comprehensive docstrings to the workflow and helper to improve clarity and editor hints.
- Chores
  - Cleaned up default config output to reduce noise and simplify generated templates.

### DIFF
--- a/src/nat/cli/commands/workflow/templates/config.yml.j2
+++ b/src/nat/cli/commands/workflow/templates/config.yml.j2
@@ -1,15 +1,3 @@
-general:
-  logging:
-    console:
-      _type: console
-      level: WARN
-
-  front_end:
-    _type: fastapi
-
-  front_end:
-    _type: console
-
 functions:
   current_datetime:
     _type: current_datetime
@@ -27,4 +15,3 @@ workflow:
   _type: react_agent
   llm_name: nim_llm
   tool_names: [current_datetime, {{python_safe_workflow_name}}]
-  parse_agent_response_max_retries: 3
--- a/src/nat/cli/commands/workflow/templates/register.py.j2
+++ b/src/nat/cli/commands/workflow/templates/register.py.j2
@@ -1,4 +1,4 @@
 # flake8: noqa
 
 # Import the generated workflow function to trigger registration
-from {{package_name}} import {{ python_safe_workflow_name }}_function
+from .{{package_name}} import {{ python_safe_workflow_name }}_function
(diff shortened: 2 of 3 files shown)


ericevans-nv (approved):

src/nat/cli/commands/workflow/templates/register.py.j2
```diff
@@ -1,4 +1,4 @@
 # flake8: noqa
 
 # Import the generated workflow function to trigger registration
-from {{package_name}} import {{ python_safe_workflow_name }}
+from {{package_name}} import {{ python_safe_workflow_name }}_function
```
zhongxuanwang-nv: Thanks Will! I think for this place what's after import should be the workflow python file instead, and since the file is renamed and no longer have the _function suffix, I guess we should remove this suffix here too? I think here it can run because it would try to import the {{ python_safe_workflow_name }}_function from the workflow python file, but we actually intend to import the entire file, because there might be more than one registered functions.

```suggestion
from {{package_name}} import {{ python_safe_workflow_name }}
```

src/nat/cli/commands/workflow/templates/config.yml.j2
```diff
@@ -1,15 +1,3 @@
-general:
```
zhongxuanwang-nv, line 1: I wasn't actually sure why these were there :/ but I do agree these look kinda verbose!

src/nat/cli/commands/workflow/templates/config.yml.j2
```diff
@@ -27,4 +15,3 @@ workflow:
   _type: react_agent
   llm_name: nim_llm
   tool_names: [current_datetime, {{python_safe_workflow_name}}]
-  parse_agent_response_max_retries: 3
```
zhongxuanwang-nv, line 30: Hey Will, I added this because when I experimented with the agent workflow I created, I found out in some cases (when the instruction is not so clear, so for example `nat run --config_file ... --input "hey"`), it would take some extra turns in order to use `echo` function correctly, so I had this `parse_agent_response_max_retries: 3` to let it retry, and also because I think letting users know we have this might be good? But I do agree it's also reasonable to take it out because this might make the yaml file too verbose!

willkill07: The default value is one which means it fails sooner. With an appropriate input though, this is unnecessary.

src/nat/cli/commands/workflow/templates/register.py.j2
willkill07: already fixed :)

PR #1074, opened 2025-10-22
fix: pin langchain to prevent upgrade

willkill07:
## Description

langchain 1.0.0 has officially been released.

To prevent pip from pulling this down, we should proactively pin to the last release before 1.0

Closes

## Summary by CodeRabbit

* **Chores**
  * Added a new dependency (langchain) to the project dependencies to keep libraries up to date and ensure compatibility with related components.
  * Updated dependency list to reflect the change and maintain ecosystem alignment.

### DIFF
--- a/packages/nvidia_nat_langchain/pyproject.toml
+++ b/packages/nvidia_nat_langchain/pyproject.toml
@@ -21,7 +21,7 @@ dependencies = [
   # version when adding a new package. If unsure, default to using `~=` instead of `==`. Does not apply to nvidia-nat packages.
   # Keep sorted!!!
   "nvidia-nat~=1.3",
-  "langchain~=0.3.27",
+  "langchain~=0.3.27", # ensure a pin to latest version <1.0.0 (note: langchain packages update separately)
   "langchain-aws~=0.2.31",
   "langchain-core~=0.3.75",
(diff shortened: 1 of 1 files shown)


packages/nvidia_nat_langchain/pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
   # version when adding a new package. If unsure, default to using `~=` instead of `==`. Does not apply to nvidia-nat packages.
   # Keep sorted!!!
   "nvidia-nat~=1.3",
+  "langchain~=0.3.27",
```
mdemoret-nv: Why are we specifying different versions of `lanchain` and `langchain-core`?

willkill07: Because they are not in sync.

For instance, right now the latest version of each on PyPI is:
- `langchain-core==1.0.0`
- `langchain==1.0.2`

I chose the version based on what already existed in the `uv.lock` file

mdemoret-nv: Got it. Can you add a comment illustrating that?

willkill07: sure!

PR #1076, opened 2025-10-22
feat: nat optimizer support for Optuna GridSearch 

bbednarski9:
## Description
Currently `nat optimizer` only supports sampler=None for Optuna, which defaults the search spaces for each OptimizableField parameter to use either bayesian search (search dim=1) or genetic algo. (search space dim > 1). See the [create_study](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.create_study.html) docs. There are many cases where a deterministic search space is beneficial, such as evaluating multiple models, where we want unique combinations in the search space rather than an imputed search space.

It's a simple feature improvement to support grid search. Notebook 6 will be backwards compatible to this new feature in the develop branch.
Closes

## Summary by CodeRabbit

* **New Features**
  * Added numeric.sampler option (None, bayesian, grid) and exhaustive grid-search support; search spaces can be expanded into explicit grid values for optimization.

* **Documentation**
  * Updated optimizer reference with sampler choices, grid-search requirements (values or range+step), prompt exclusions, and defaults.

* **Tests**
  * Added extensive tests for grid discretization, step validation, types, prompt exclusions, and error cases.

* **Bug Fixes**
  * Improved validation and clearer error messages for invalid grid configurations.

### DIFF
--- a/docs/source/reference/optimizer.md
+++ b/docs/source/reference/optimizer.md
@@ -113,7 +113,7 @@ This is the main configuration object for the optimizer.
 -   `eval_metrics: dict[str, OptimizerMetric] | None`: A dictionary of evaluation metrics to optimize. The keys are custom names for the metrics, and the values are `OptimizerMetric` objects.
 -   `numeric.enabled: bool`: Enable numeric optimization (Optuna). Defaults to `true`.
 -   `numeric.n_trials: int`: Number of numeric trials. Defaults to `20`.
--   `numeric.sampler: str | None`: Sampling strategy for numeric optimization. `None` and `bayesian` use Optuna default: TPE for single-objective, NSGA-II for multi-objective. `grid` performs exhaustive grid search over parameter combinations. For grid search, optimizable parameters must either specify explicit `values` or provide `low`, `high`, and `step` to create the range. Defaults to `None`.
+-   `numeric.sampler: SamplerType | None`: Sampling strategy for numeric optimization. Valid values: `"bayesian"`, `"grid"`, or `None`. `None` and `"bayesian"` use Optuna default (TPE for single-objective, NSGA-II for multi-objective). `"grid"` performs exhaustive grid search over parameter combinations. For grid search, optimizable parameters must either specify explicit `values` or provide `low`, `high`, and `step` to create the range. Defaults to `None`.
 -   `prompt.enabled: bool`: Enable GA-based prompt optimization. Defaults to `false`.
 -   `prompt.ga_population_size: int`: Population size for GA prompt optimization. Larger populations increase diversity but cost more per generation. Defaults to `10`.
 -   `prompt.ga_generations: int`: Number of generations for GA prompt optimization. Replaces `n_trials_prompt`. Defaults to `5`.


src/nat/data_models/optimizer.py
```diff
@@ -34,6 +34,10 @@ class NumericOptimizationConfig(BaseModel):
     """
     enabled: bool = Field(default=True, description="Enable numeric optimization")
     n_trials: int = Field(description="Number of trials for numeric optimization.", default=20)
+    sampler: str | None = Field(
```
willkill07: Nit: maybe prefer an enum or list of potential literals?

This would eliminate a breaking change in the future if you ever wanted another sampler.

src/nat/data_models/optimizable.py
```diff
@@ -65,6 +65,57 @@ def suggest(self, trial: Trial, name: str):
...
+                f"Grid search with range (low={self.low}, high={self.high}) requires 'step' to be specified. "
+                "Please define the step size to discretize the range, for example: step=0.1")
+
+        # Generate grid values from range with step
+        if isinstance(self.low, int) and isinstance(self.high, int):
+            step = int(self.step)
+            if self.log:
+                raise ValueError("Log scale is not supported for integer ranges in grid search. "
+                                 "Please use linear scale or provide explicit values.")
+            return list(range(self.low, self.high + 1, step))
```
willkill07: It is perfectly valid and reasonable for me to use integral low + high with a floating-point step. For instance: `[low=0, high=10, step=0.5]` is a valid grid search. Unless I am misunderstanding something more fundamental.

I don't think the specialization here for using a list is needed. I would just cast to from int and go from there.

Note: the check for `self.log` does make sense though

src/nat/profiler/parameter_optimization/parameter_optimizer.py
```diff
@@ -59,7 +59,22 @@ def optimize_parameters(
     eval_metrics = [v.evaluator_name for v in metric_cfg.values()]
     weights = [v.weight for v in metric_cfg.values()]
 
-    study = optuna.create_study(directions=directions)
+    # Create appropriate sampler based on configuration
+    sampler_type = optimizer_config.numeric.sampler
+    if sampler_type is not None:
+        sampler_type = sampler_type.lower()
+
+    if sampler_type == "grid":
```
dnandakumar-nv: This could also be simplified with an `enum`

src/nat/data_models/optimizable.py
bbednarski9: yeah agree let's just go to float

src/nat/profiler/parameter_optimization/parameter_optimizer.py
bbednarski9: addressed in: [8c89bca](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1076/commits/8c89bca3030fab77f5763a5efecf5c4d24d2d902)

src/nat/data_models/optimizer.py
bbednarski9: addressed in [8c89bca](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1076/commits/8c89bca3030fab77f5763a5efecf5c4d24d2d902)

src/nat/data_models/optimizable.py
```diff
@@ -65,6 +65,57 @@ def suggest(self, trial: Trial, name: str):
...
+
```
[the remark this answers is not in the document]
bbednarski9: non issue after changes in [8c89bca](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1076/commits/8c89bca3030fab77f5763a5efecf5c4d24d2d902) we are now casting to float

src/nat/data_models/optimizable.py
bbednarski9: [8c89bca](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1076/commits/8c89bca3030fab77f5763a5efecf5c4d24d2d902)

src/nat/data_models/optimizable.py
```diff
@@ -95,26 +95,39 @@ def to_grid_values(self) -> list:
                 f"Grid search with range (low={self.low}, high={self.high}) requires 'step' to be specified. "
                 "Please define the step size to discretize the range, for example: step=0.1")
 
+        # Validate step is positive
+        step_float = float(self.step)
+        if step_float <= 0:
```
willkill07, line 126: also make sure low <= high ?

src/nat/data_models/optimizer.py
```diff
@@ -28,15 +29,22 @@ class OptimizerMetric(BaseModel):
...
+
+
 class NumericOptimizationConfig(BaseModel):
     """
     Configuration for numeric/enum optimization (Optuna).
     """
     enabled: bool = Field(default=True, description="Enable numeric optimization")
     n_trials: int = Field(description="Number of trials for numeric optimization.", default=20)
-    sampler: str | None = Field(
+    sampler: SamplerType | None = Field(
```
willkill07, line 43: Do we need this as an optional? wouldn't a default value of `'bayesian'` work?

src/nat/data_models/optimizable.py
bbednarski9: we can add this to the search space model_validator. We will also not accept low/high for prompt fields

src/nat/data_models/optimizer.py
bbednarski9: It would work the same using bayesian default. However, this way leaves the door open for us to re-define a preferred bayesian appraoch. I'm not sure how much validation has been done against search methods to date. But it might be beneficial to re-evaluate these with respect to LLM params and tweak our in house "bayesian" approach compared to Optuna's default

src/nat/data_models/optimizable.py
bbednarski9: fixing this thanks

docs/source/reference/optimizer.md
```diff
@@ -113,6 +113,7 @@ This is the main configuration object for the optimizer.
 -   `eval_metrics: dict[str, OptimizerMetric] | None`: A dictionary of evaluation metrics to optimize. The keys are custom names for the metrics, and the values are `OptimizerMetric` objects.
 -   `numeric.enabled: bool`: Enable numeric optimization (Optuna). Defaults to `true`.
 -   `numeric.n_trials: int`: Number of numeric trials. Defaults to `20`.
+-   `numeric.sampler: str | None`: Sampling strategy for numeric optimization. `None` and `bayesian` use Optuna default: TPE for single-objective, NSGA-II for multi-objective. `grid` performs exhaustive grid search over parameter combinations. For grid search, optimizable parameters must either specify explicit `values` or provide `low`, `high`, and `step` to create the range. Defaults to `None`.
```
willkill07: Last piece of feedback (a micronit)

the type is technically a Enum -- unsure if we want to explicitly list that type here. Since its in the docs you can link to the type

willkill07 (approved):

PR #1132, opened 2025-10-31
Add default Phoenix session tracking support

ericevans-nv:
## Description

Adds `session.id` attribute to spans to enable Phoenix session grouping.

All traces from a conversation are now grouped together in Phoenix's Sessions view. This enables better visibility into multi-turn interactions.

## Summary by CodeRabbit

* **New Features**
  * Session-level grouping has been added to observability traces: a distinct session identifier is now attached to related spans, improving the ability to correlate and navigate session-related operations within trace data for clearer end-to-end visibility.

### DIFF
--- a/src/nat/observability/exporter/span_exporter.py
+++ b/src/nat/observability/exporter/span_exporter.py
@@ -196,7 +196,7 @@ class SpanExporter(ProcessingExporter[InputSpanT, OutputSpanT], SerializeMixin):
         span_kind = event_type_to_span_kind(event.event_type)
         sub_span.set_attribute(f"{self._span_prefix}.span.kind", span_kind.value)
 
-        # Enable Phoenix session grouping by setting session.id from conversation_id
+        # Enable session grouping by setting session.id from conversation_id
         try:
             conversation_id = self._context_state.conversation_id.get()
             if conversation_id:


src/nat/observability/exporter/span_exporter.py
```diff
@@ -196,6 +196,14 @@ def _process_start_event(self, event: IntermediateStep):
         span_kind = event_type_to_span_kind(event.event_type)
         sub_span.set_attribute(f"{self._span_prefix}.span.kind", span_kind.value)
 
+        # Enable Phoenix session grouping by setting session.id from conversation_id
```
mpenn: Suggest making this comment a little more general. This attribute will be available to all exporter that subclass the `SpanExporter` including but not limited to Phoenix.

src/nat/observability/exporter/span_exporter.py
```diff
@@ -196,6 +196,14 @@ def _process_start_event(self, event: IntermediateStep):
...
+        try:
+            conversation_id = self._context_state.conversation_id.get()
+            if conversation_id:
+                sub_span.set_attribute("session.id", conversation_id)
+        except (AttributeError, LookupError):
+            pass
+
```
mpenn, line 206: Should we instead always add the `session.id` attribute with a default value (e.g. `unknown`) it doesn't exist?

mpenn (changes requested):

src/nat/observability/exporter/span_exporter.py
ericevans-nv: I think we should only set it when a valid value is provided since it is optional. That way, the sessions view remains clean and accurate.

mpenn (approved):

PR #1143, opened 2025-11-03
feat!(llm): exclude unset fields in model dump for all LLMs and Embedders

willkill07:
## Description
Make it so we only forward what has been specified to LLM and Embedder providers.
This gives the flexibility and control to end-users rather than the toolkit being naturally unable to support newer models with added imposed restrictions (such as GPT-5 and Claude Sonnet 4.5)

- Remove `TopPMixin` and `TemperatureMixin` due to not being needed
- Add `exclude_unset=True` to all LLM and Embedder `model_dump`

## Summary by CodeRabbit

* **Documentation**
  * Removed deprecated gated-field mixin examples and simplified gated-fields docs.

* **Refactor**
  * Exposed temperature and top-p as explicit, consistent configurable model parameters across providers.
  * Standardized client serialization to omit unset configuration fields when building LLM clients.

* **Tests**
  * Removed legacy mixin-specific tests and updated assertions to align with new defaults.

* **Examples**
  * Added explicit temperature settings in several example configs.

### DIFF
--- a/examples/frameworks/agno_personal_finance/src/nat_agno_personal_finance/configs/config.yml
+++ b/examples/frameworks/agno_personal_finance/src/nat_agno_personal_finance/configs/config.yml
@@ -25,6 +25,7 @@ llms:
     model_name: gpt-4o
     max_tokens: 2000
     api_key: ${OPENAI_API_KEY}
+    temperature: 0.0
 
 workflow:
   _type: agno_personal_finance
--- a/examples/frameworks/haystack_deep_research_agent/src/nat_haystack_deep_research_agent/configs/config.yml
+++ b/examples/frameworks/haystack_deep_research_agent/src/nat_haystack_deep_research_agent/configs/config.yml
@@ -17,9 +17,11 @@ llms:
   rag_llm:
     _type: nim
     model: nvidia/llama-3.3-nemotron-super-49b-v1
+    temperature: 0.0
   agent_llm:
     _type: nim
     model: nvidia/llama-3.3-nemotron-super-49b-v1
+    temperature: 0.0
 
 workflow:
   _type: haystack_deep_research_agent
--- a/examples/frameworks/semantic_kernel_demo/src/nat_semantic_kernel_demo/configs/config.yml
+++ b/examples/frameworks/semantic_kernel_demo/src/nat_semantic_kernel_demo/configs/config.yml
@@ -51,6 +51,7 @@ llms:
   openai_llm:
     _type: openai
     model_name: gpt-3.5-turbo
+    temperature: 0.0
 
 workflow:
   _type: semantic_kernel


docs/source/extend/adding-an-llm-provider.md
```diff
@@ -84,55 +84,12 @@ class OpenAIModelConfig(LLMBaseConfig, RetryMixin, name="openai"):
 
 Some configuration parameters are only valid for certain models or may be dependent on other parameters. The toolkit provides built-in mixins that automatically validate and default these parameters based on a specified field. For details on the mechanism, see [Gated Fields](./gated-fields.md).
 
-- `TemperatureMixin`: adds a `temperature` field in [0, 1], with a default of `0.0` when supported by a model
-- `TopPMixin`: adds a `top_p` field in [0, 1], with a default of `1.0` when supported by a model
 - `ThinkingMixin`: adds a `thinking` field, with a default of `None` when supported by a model. If supported, the `thinking_system_prompt` property will return the system prompt to use for thinking.
```
bbednarski9, line 87: If ThinkingMixin is the only Gated Field Mixin after this change, should we include that as an explicit comment in this README? Folks may go looking for others. Perhaps we should also include a note about roadmapping for others that are on the way?

bbednarski9: link to gated-fields.md is in a good spot though and denotes the above.

src/nat/llm/azure_openai_llm.py
```diff
@@ -50,6 +48,16 @@ class AzureOpenAIModelConfig(
                                   serialization_alias="azure_deployment",
                                   description="The Azure OpenAI hosted model/deployment name.")
     seed: int | None = Field(default=None, description="Random seed to set for generation.")
+    temperature: float | None = OptimizableField(default=None,
+                                                 ge=0.0,
+                                                 le=1.0,
```
bbednarski9, line 53: Azure OpenAI models allow temperature values between 0.0-2.0.

Generally 1.0 is recommended highest, but they should accept up to 2.0, users might have a UC for this.

https://learn.microsoft.com/en-us/azure/ai-foundry/openai/latest

src/nat/llm/nim_llm.py
```diff
@@ -51,6 +43,16 @@ class NIMModelConfig(LLMBaseConfig,
     max_tokens: PositiveInt = OptimizableField(default=300,
                                                description="Maximum number of tokens to generate.",
                                                space=SearchSpace(high=2176, low=128, step=512))
+    temperature: float | None = OptimizableField(default=None,
+                                                 ge=0.0,
+                                                 le=1.0,
```
bbednarski9, line 48: same comment as OpenAI models. What if a user has a UC for temperature = 1.5 for some model?

I think this comment pertains to all providers except for AWS

src/nat/llm/azure_openai_llm.py
willkill07: This PR retains all existing behavior of temperature and top_p fields

The only difference is rather than inheriting from a Mixin, we simply copy and paste the fields.

This discrepancy with models provided by third-party providers already existed, and I consider only accepting a valid subrange as out-of-scope for this PR.

src/nat/llm/nim_llm.py
willkill07: This already would have failed prior to this PR since we actively limit between 0.0 and 1.0

bbednarski9: for sure. out of scope of this PR then

src/nat/llm/azure_openai_llm.py
bbednarski9: _"I consider only accepting a valid subrange as out-of-scope for this PR."_

Rather than only accepting a valid range, im actually making a case that it's not our responsibility. especially without the mixins, the range can be anywhere from 0-5.0 depending on the model and may vary intra-provider

bbednarski9: I'll approve PR because its out of scope but would be interested in your thoughts looking forward.

bbednarski9 (approved):

src/nat/llm/azure_openai_llm.py
willkill07: Do you think just having a constraint >= 0.0 would be broad enough to encapsulate any potential model?

PR #1185, opened 2025-11-17
fix(azure-openai): ensure api_version is specified

willkill07:
## Description
#1143 incorrectly restricted model_dump to exclude defaults.
We still need to emit defaults where appropriate.

Closes AIQ-2359

## Summary by CodeRabbit

* **Improvements**
  * Explicit API version support added across Azure OpenAI LLM and embedding integrations so configured api_version values are consistently forwarded to clients and preserved in model/state construction.
  * Reduces mismatches between configuration and runtime client behavior for different Azure deployments.

* **Tests**
  * Tests parameterized to validate behavior with multiple api_version values to ensure compatibility.

### DIFF
--- a/packages/nvidia_nat_llama_index/src/nat/plugins/llama_index/embedder.py
+++ b/packages/nvidia_nat_llama_index/src/nat/plugins/llama_index/embedder.py
@@ -29,7 +29,9 @@ async def azure_openai_llama_index(embedder_config: AzureOpenAIEmbedderModelConf
     from llama_index.embeddings.azure_openai import AzureOpenAIEmbedding
 
     client = AzureOpenAIEmbedding(
-        **embedder_config.model_dump(exclude={"type"}, by_alias=True, exclude_none=True, exclude_unset=True))
+        **embedder_config.model_dump(exclude={"type"}, by_alias=True, exclude_none=True, exclude_unset=True),
+        api_version=embedder_config.api_version,
+    )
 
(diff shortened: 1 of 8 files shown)


packages/nvidia_nat_llama_index/src/nat/plugins/llama_index/embedder.py
```diff
@@ -29,7 +29,9 @@ async def azure_openai_llama_index(embedder_config: AzureOpenAIEmbedderModelConf
     from llama_index.embeddings.azure_openai import AzureOpenAIEmbedding
 
     client = AzureOpenAIEmbedding(
-        **embedder_config.model_dump(exclude={"type"}, by_alias=True, exclude_none=True, exclude_unset=True))
+        **embedder_config.model_dump(exclude={"type"}, by_alias=True, exclude_none=True, exclude_unset=True),
+        api_version=embedder_config.api_version,
```
dagardner-nv, line 36: There's a problem here in that when `api_version` is explicitly set it will be included in the model dump resulting in a `TypeError`. 

The fix I believe is to mimic what you did in `packages/nvidia_nat_adk/src/nat/plugins/adk/llm.py` to perform a model dump, and then explicitly set `api_version` in the dumped dict.

dagardner-nv (changes requested):

dagardner-nv (approved):

PR #1216, opened 2025-12-02
Extend Middleware interface with pre/post invoke hooks and add DynamicFunctionMiddleware

ericevans-nv:
## Dynamic Middleware and Pre/Post Invoke Hooks

This PR introduces **DynamicFunctionMiddleware** for automatic component discovery and interception, and enhances the middleware interface with **pre/post invoke hooks** using structured context objects.

### DynamicFunctionMiddleware

A new middleware base class that extends `FunctionMiddleware` to provide:

- **Automatic component discovery** — discovers and wraps functions from LLMs, embedders, retrievers, memory, object stores, and auth providers at runtime
- **Builder patching** — intercepts `builder.get_llm()`, `builder.get_embedder()`, etc. to wrap component methods
- **Allowlist-based security** — only configured component methods are wrapped (secure by default)
- **Configurable discovery** — enable/disable discovery per component type via `register_*` flags
- **Explicit component references** — specify exact components to wrap
- **Workflow function registration** — automatically intercept user-defined workflow functions
- **Runtime unregistration** — support for unregistering discovered functions at runtime

### Pre/Post Invoke Hooks

Enhanced the middleware interface with structured context objects:

- **`pre_invoke(context: PreInvokeContext)`** — hook called before function execution
- **`post_invoke(context: PostInvokeContext)`** — hook called after function execution
- **`PreInvokeContext`** — contains frozen `original_args`/`original_kwargs` and mutable `modified_args`/`modified_kwargs`
- **`PostInvokeContext`** — contains frozen invocation data and mutable `output`

### LoggingMiddleware

Updated the logging middleware example to demonstrate pre/post invoke hook logic:

```python
class LoggingMiddleware(DynamicFunctionMiddleware):
    async def pre_invoke(self, context: PreInvokeContext) -> PreInvokeContext | None:
        logger.info(f"Calling {context.function_context.name} with args: {context.modified_args}")
        return None

    async def post_invoke(self, context: PostInvokeContext) -> PostInvokeContext | None:
        logger.info(f"{context.function_context.name} returned: {context.output}")
        return None
```
### WorkflowInventory

Added utility for tracking discovered components and their callable functions during middleware initialization.

### Auto-Discovery Flags

- `register_llms` (bool, default=`False`)
- `register_embedders` (bool, default=`False`)
- `register_retrievers` (bool, default=`False`)
- `register_memory` (bool, default=`False`)
- `register_object_stores` (bool, default=`False`)
- `register_auth_providers` (bool, default=`False`)
- `register_workflow_functions` (bool, default=`False`)

### Explicit Component References

- `llms`, `embedders`, `retrievers`, `memory`, `object_stores`, `auth_providers`: Lists of specific component names to intercept

### Function Allowlists

- `allowed_component_functions`: Extend default allowlists with additional methods

## Default Allowed Functions by Component Type

| Component Type | Default Allowed Methods |
|----------------|------------------------|
| **LLMs** | `invoke`, `ainvoke`, `stream`, `astream` |
| **Embedders** | `embed_query`, `aembed_query` |
| **Retrievers** | `search` |
| **Memory** | `search`, `add_items`, `remove_items` |
| **Object Stores** | `put_object`, `get_object`, `delete_object`, `upsert_object` |
| **Authentication** | `authenticate` |

## Summary by CodeRabbit

* **New Features**
  * Runtime dynamic middleware for auto-discovery, registration, unregistration, and per-call interception of workflow components and functions.
  * Logging middleware with configurable log level and pre/post-invoke logging.
  * Unified invocation context and pre-/post-invoke lifecycle hooks for middleware.

* **Bug Fixes**
  * Cache middleware: improved enablement modes, input handling, serialization fallbacks, and argument forwarding.

* **Refactor**
  * Middleware API moved to config-driven registration and broader invocation signatures.

* **Tests**
  * Extensive new tests for dynamic middleware, discovery, policies, registration, and caching.


src/nat/data_models/middleware.py
```diff
@@ -33,3 +47,110 @@ class MiddlewareBaseConfig(TypedBaseModel, BaseModelRegistryTag):
...
+
+    register_auth_providers: bool | None = Field(
+        default=False, description="Auto-discover and register all authentication providers component functions")
+
+    register_workflow_functions: bool | None = Field(default=False,
+                                                     description="Auto-discover and register all workflow functions")
+
+    # === Policy Configuration ===
+
+    pre_invoke_policy: list[FunctionPolicyRef] | None = Field(
```
kshiarl1: You have created the dynamic function middleware so that it can be applied to 
1. A huge range of function and components. Which is great. 
2. Potentially multiple functions or components at the same time. Also great. 

So one limitation I foresee is that if these components have different input and output schemas we would have one policy per intercepted component and hence one dynamic middleware per intercepted component. Do you see this as a big issue? Any ideas for resolving it?

dnandakumar-nv:
What is the difference between function policy and function middleware? Seems to me like the do the same thing: 
1. Wrap the I/O of a function with custom logic (optionally with the ability to modify the content)
2. Chain multiple instances to create cascading policies or middleware 

Is there a specific functionality policies provide that middleware cannot? If so, would it be worth extending a function middleware instead so we don't have duplicated functionality?

ericevans-nv:
Function policies are a higher-level abstraction that runs within DynamicFunctionMiddleware, which handles auto-discovery, allowlist security, and policy orchestration for both sync and streaming functions. Policies provide a simple on_pre_invoke/on_post_invoke interface for inspecting or transforming inputs and outputs, users just receive a context object and return a value. Unlike middleware, policies don't require knowing which functions to intercept at compile time; the dynamic middleware discovers and wraps components at runtime, then applies policies to them. This lets users build and share reusable policy logic without writing custom middleware or manually registering functions. @dnandakumar-nv

src/nat/data_models/middleware.py
ericevans-nv: The input and output schemas are standardized through the policy pre/post-invoke context, which forwards all positional arguments and keyword arguments. This means policies operate on the function inputs, not typed schemas, so a single policy can handle multiple components with different schemas.

src/nat/middleware/middleware.py
```diff
@@ -66,31 +69,75 @@ class FunctionMiddlewareContext:
...
+        frozen=True, description="The original function input keyword arguments before any middleware processing.")
+
+    # Mutable fields - modify these to transform inputs
+    modified_args: tuple[Any,
+                         ...] = Field(description="The modified function input arguments after middleware processing.")
+    modified_kwargs: dict[str, Any] = Field(
+        description="The modified function input keyword arguments after middleware processing.")
+
+
+class PostInvokeContext(BaseModel):
```
dnandakumar-nv: I don't think we should have two different contexts for pre and post invoke. I think we should standardize around a include `InvocationContext` and have the `output` field be `Any | None`

docs/source/build-workflows/advanced/middleware.md
```diff
@@ -80,95 +80,165 @@ async def my_function(config, builder):
...
+**Explicit Component References:**
+
+For fine-grained control, specify exactly which components to intercept (alternative to auto-discovery):
+
+- `llms` (list, default=`[]`): Specific LLM component names to intercept
+- `embedders` (list, default=`[]`): Specific embedder component names to intercept
+- `retrievers` (list, default=`[]`): Specific retriever component names to intercept
+- `memory` (list, default=`[]`): Specific memory provider component names to intercept
+- `object_stores` (list, default=`[]`): Specific object store component names to intercept
+- `auth_providers` (list, default=`[]`): Specific authentication provider component names to intercept
```
AnuradhaKaruppiah, line 130: Only components used bye the workflow can have middleware? I would have liked to see evaluators in this list

AnuradhaKaruppiah (approved):

docs/source/build-workflows/advanced/middleware.md
ericevans-nv: Interesting point. I hadn’t considered intercepting evaluator components before. I’d love to hear what use cases you have in mind. @AnuradhaKaruppiah

PR #1227, opened 2025-12-05
Defense middleware

soumilinandi:
## Description

Closes


ericevans-nv (approved):

ericevans-nv (approved):

src/nat/middleware/defense_middleware.py
```diff
@@ -0,0 +1,171 @@
...
+    
+    This base class provides:
+    - Common configuration fields (action, check_input, check_output, llm_wrapper_type)
+    - Helper methods for LLM loading (for LLM-based defenses)
+    - Access to builder for any resources needed
+    
+    Unlike an abstract base class, this does NOT enforce a specific pattern.
+    Each defense middleware implements its own invoke/stream logic based on
+    its specific defense strategy:
+    - LLM-based analysis (guard models, verifiers)
```
shaonag, line 111: nit: Rules and Heuristics can be combined into one. 
We will also have ML models that are not LLMs. For example a topic detection model, intent detection ML Model may or may not be a LLM based model. Perhaps we could have LLM based, ML based, Rules / Heuristics based...

src/nat/middleware/defense_middleware_content_guard.py
```diff
@@ -0,0 +1,333 @@
...
+
+class ContentSafetyGuardMiddlewareConfig(DefenseMiddlewareConfig, name="content_safety_guard"):
+    """Configuration for Content Safety Guard middleware.
+    
+    This middleware uses guard models to classify content as safe or harmful.
+    
+    Actions:
+    - 'log': Pass through with warning
+    - 'block': Return error dict (hard stop)
+    - 'sanitize': Return polite refusal message (user-friendly)
```
shaonag: For this, the equivalent safety domain terms are 
- Partial Compliance : Comply with user request with warning.
- Refusal: Refuse user request (Hard refusal)
- Redirection: Redirect user request to a safe place; provide a safer response by redirection. 

We may need to include Full Compliance as well in future as based on certain runtime policy and governance directives or user requirements, we should let the model respond even if the guard model detected something unsafe. For example if the user is designing a game bot for a crime game, they game bot may be allowed to respond about certain unsafe concepts pertaining to the crime game but block other content safety harm categories.

src/nat/middleware/defense_middleware_content_guard.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        try:
+            # Get the guard model LLM
+            llm = await self._get_llm()
+            
+            # Prepare simple prompt - just ask if content is harmful
+            content_str = str(content)
+            
+            if content_type == "output" and original_input is not None:
+                prompt = f"""Is this AI response harmful or dangerous?
```
shaonag: Depending on the guard model used, this prompt will be different for example NVIDIA's aegis model will not work with this.  In future, we would want to be able to either have a dynamic way to setting this. Also, something to think about for future is user may want to bring in their own guard model in which case they can set the prompt in the config file?

src/nat/middleware/defense_middleware_output_verifier.py
```diff
@@ -0,0 +1,332 @@
...
+        description="System prompt for the LLM"
+    )
+
+
+class OutputVerifierMiddleware(DefenseMiddleware):
+    """Verification middleware using an LLM for correctness and security.
+    
+    This middleware uses NAT's LLM system to verify function outputs for:
+    - Mathematical correctness
+    - Security threats and malicious content
```
shaonag: Security (CIAAN)

src/nat/middleware/defense_middleware_pii.py
```diff
@@ -0,0 +1,252 @@
...
+from nat.middleware.middleware import FunctionMiddlewareContext
+
+logger = logging.getLogger(__name__)
+
+
+class PIIDefenseMiddlewareConfig(DefenseMiddlewareConfig, name="pii_defense"):
+    """Configuration for PII Defense Middleware using Microsoft Presidio.
+
+    Detects PII in function outputs. Actions: 'log', 'block', or 'sanitize'.
+    Uses Presidio's rule-based entity recognition (no LLM required).
```
shaonag: We need to add link to Microsoft's presidio library either here or where we are loading the library at startup. @ericevans-nv  highlighting to you that this is a partner (Microsoft's) open source PII library, we have approval to use and ship.

shaonag (commented):
LG. Thanks for the hard work!

src/nat/middleware/defense_middleware_content_guard.py
soumilinandi: Done. Added `Partial_Compliance`, `Refusal`, and `Redirection` as action now

src/nat/middleware/defense_middleware_content_guard.py
soumilinandi: great suggestion @shaonag ,  most guard models don't use a custom prompt like qwen or nemoguard - rather it takes just the Input string and classifies it. Output format can differ per model. I have improved the logic to just use tool input directly to model and added parsing logic for different output formats. Tested with `llama-3.1-nemoguard-8b-content-safety` using nim and `Qwen3Guard` using huggingace. Both can be used via config now. Please let me know if you have suggestions

src/nat/middleware/defense_middleware_pii.py
soumilinandi: Done. added the URL

src/nat/middleware/defense_middleware_content_guard.py
```diff
@@ -0,0 +1,425 @@
...
+                        categories = [cat.strip() for cat in category_text.split(",")]
+                        break
+            
+            return categories
+        except Exception:
+            # If any error occurs during category extraction, return empty list
+            logger.debug("Failed to extract categories from guard response, returning empty list")
+            return []
+    
+    def _parse_guard_response(self, response_text: str) -> dict:
```
kshiarl1, line 152: We could use the same configuration logic that we used for finding fields in inputs and outputs within the middleware. Also allow the user to define which fields are safe and unsafe? It depends on what the variability is on the outputs of these models

PR #1206, opened 2025-11-25
Implement Per-User Function Instantiation

yczhang-nv:
## Description

This PR implements a new feature that enables users to register `Function`/`FunctionGroup` as `per-user` scope, using the `@register_per_user_function()` decorator, which will make the component has a separate instance per-user and built lazily as the first request of the user being processed by the `FastAPI` server.

- Enable Lazy Workflow Initialization
    - On-demand per-user building: Defer per-user function and workflow creation until a user's first request arrives, eliminating per-user workflow building during server startup
    - Configurable component scope: Support explicit declaration of sharing scope for functions via `@register_function` (shared) or `@register_per_user_function` (per-user)
    - Schema availability at startup: Enable `FastAPI` OpenAPI documentation generation by requiring explicit schema declaration for per-user workflows (to keep the documentation consistent as shared workflows)
- Enable Full User Isolation
    - Per-user workflow instances: Each user gets their own `PerUserWorkflowBuilder` with isolated per-user functions
    - Stateful per-user functions: Support functions that maintain state across requests within the same user session
    - Automatic lifecycle management: Per-user builders are created on-demand, cached during active use, and automatically cleaned up after configurable inactivity timeout
    - User identification: Extract user identity from nat-session cookie
- Improve Architecture Clarity
    - Clear component ownership: Shared components (LLMs, embedders, shared functions) are built once in `WorkflowBuilder`; per-user components are built separately in `PerUserWorkflowBuilder`
    - Dependency validation: Enforce at build time that shared components cannot depend on per-user components
    - Route isolation: Multiple FastAPI routes (default, eval, custom endpoints) each have their own `SessionManager` with isolated per-user caches

Closes AIQ-2342

## By Submitting this PR I confirm:
- I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/docs/source/resources/contributing.md).
- We require that all contributors "sign-off" on their commits. This certifies that the contribution is your original work, or you have rights to submit it under the same license, or a compatible license.
  - Any contribution which contains commits that are not Signed-Off will not be accepted.
- When the PR is ready for review, new or existing tests cover these changes.
- When the PR is ready for review, the documentation is up to date with these changes.

## Summary by CodeRabbit

* **New Features**
  * Per-user workflows with isolated per-user state, lazy per-user component instantiation, per-request sessions, and CLI/web support for specifying user IDs.
  * Register per-user functions and function groups with explicit schemas.

* **Enhancements**
[...]

PR #1253, opened 2025-12-15
Implement Non-session-aware Per-user `MCPClient`

yczhang-nv:
## Description

As the Per-user function feature ([PR-1206](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1206)) becomes available, the `MCPClient` component can be initialized per-user so every user gets their own separate instance. Thus the `MCPClient` implementation can be significantly simplified (`PerUserMCPClient` in this PR) by removing the session management and `default_user_id` (since users will get themselves authenticated lazily upon the first request).

Note that the `PerUserMCPClient` only works with per-user workflow. This requires updating any existing workflow that's not per-user to be registered with `@register_per_user_function`. To making it a smooth migration, the previous implementation was kept in this PR and a migration guide is provided.

This PR needs to be merged after [PR-1206](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1206).

Closes AIQ-2507

## Release Notes

* **New Features**
  * Added support for per-user MCP authentication, enabling multi-user isolation and per-user workflow execution.

* **Documentation**
  * Added comprehensive guidance for running per-user workflows with MCP authentication, including setup instructions and environment configuration steps.

* **Examples**
  * Added per-user configuration examples for Kaggle and Jira integrations with multi-user server support and OAuth authentication.


src/nat/authentication/oauth2/oauth2_auth_code_flow_provider.py
```diff
@@ -105,11 +105,14 @@ async def authenticate(self, user_id: str | None = None, **kwargs) -> AuthResult
...
                     return refreshed_auth_result
 
-        # Try getting callback from the context if that's not set, use the default callback
-        try:
-            auth_callback = Context.get().user_auth_callback
-        except RuntimeError:
-            auth_callback = self._auth_callback
+        # Prefer custom auth callback if explicitly set (e.g., MCP-specific handler),
+        # otherwise fall back to context's callback
+        auth_callback = self._auth_callback
```
AnuradhaKaruppiah: Why do you need to use the custom callback? The Console/WebSockets ones should be available at the time of lazy init, right?

PS: custom callback is largely a hack done for early init (with shared client)

yczhang-nv: Thanks for pointing that out, just verified that the change is not needed

AnuradhaKaruppiah (approved):

PR #1271, opened 2025-12-16
Fix MCP integration issues with parameter handling and async context management

### LINKED ISSUE #1269: MCP tool calls fail when optional parameters are not provided - incorrect null handling
### Issue
When NAT agent calls MCP tools (e.g., Sourcegraph's `sg_list_repos`) without providing optional parameters, the tool invocation includes `null` values for those parameters, causing validation errors from the MCP server.

### Error Message
```
invalid params: validating "arguments": validating root: 
validating /properties/after: type: <invalid reflect.Value> has type "null", want "string"
```

### Root Cause
`packages/nvidia_nat_mcp/src/nat/plugins/mcp/tool.py` line 92 uses `model_dump()` without `exclude_none=True`, which includes all Pydantic fields even those with `None` values.

The MCP protocol expects optional parameters to be **omitted entirely** when not provided, not set to `null`.

### Expected Behavior
- Agent calls: `sg_list_repos(query="authenticate")`
- NAT should send: `{"query": "authenticate"}`
- Currently sends: `{"query": "authenticate", "after": null, "limit": null}`

### Reproduction
Any MCP tool call with optional parameters not provided will fail validation.

### Affected Versions
NAT 1.3.1 and prior

### LINKED ISSUE #1270: MCP client cleanup fails with cancel scope context violation in stdio transport
### Issue
When NAT's MCP client (using stdio transport) exits its async context manager, it crashes with:

```
RuntimeError: Attempted to exit cancel scope in a different task than it was entered in
```

This occurs during normal cleanup, even with `reconnect_enabled: false`.

### Root Cause
`packages/nvidia_nat_mcp/src/nat/plugins/mcp/client_base.py` lines 195-217 use `AsyncExitStack` to manage the MCP session lifecycle. When the context manager yields to return control to the caller, execution switches to a different task. Later, when `AsyncExitStack.aclose()` is called during context exit, it attempts to clean up resources (particularly anyio cancel scopes) from a different task than the one that created them.

The MCP SDK's `stdio_client()` internally uses `anyio.fail_after()` which creates task-local cancel scopes. This violates anyio's requirement that cancel scopes be exited in the same task that created them.

### Error Details
- Entry: `AsyncExitStack.enter_async_context()` in Task A (connect_to_server)
- Exit: `AsyncExitStack.aclose()` in Task B (caller task)
- Violation: anyio cancel scope task-local state assertion fails

### Expected Behavior
MCP client session cleanup completes cleanly without task context errors.

### Reproduction
```python
async with MCPStdioClient(...) as client:
    tools = await client.get_tools()
    # ... use tools ...
# ❌ RuntimeError during implicit __aexit__
```

### Affected Versions
NAT 1.3.1 and prior with stdio transport

sjarmak:
## Summary
Fixes two critical bugs preventing NAT from working correctly with MCP tools via stdio transport.

## Issues Closed
Closes #1269 (optional parameter validation error)
Closes #1270 (async context cleanup crash)

### 1. Fix Optional Parameter Handling (tool.py)
- Changed: `model_dump()` → `model_dump(exclude_none=True)`
- Rationale: MCP protocol expects optional params to be omitted when not provided
- Impact: Tool calls with partial parameters now succeed

### 2. Fix Async Context Management (client_base.py)
- Changed: AsyncExitStack pattern → explicit context manager
- Rationale: Ensures entry/exit in same task to comply with anyio requirements
- Impact: Session cleanup no longer crashes with cancel scope errors

## Testing
- Verified tool calls with optional parameters omitted
- Verified MCP client cleanup without async context errors
- No behavior changes from caller perspective

## Commits
1. fix(mcp): exclude None values from optional tool parameters
2. fix(mcp): use explicit context manager for task-safe cleanup

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling and resource cleanup during client startup to avoid partially-initialized state and ensure reliable teardown.
  * Fixed handling of task/cancellation boundary errors to continue safe cleanup.

* **Refactor**
  * Reworked internal context management for better task isolation and reliability.

* **Enhancements**
  * Stronger input validation and omission of null fields when invoking tools; tool metadata now includes input schema and converters.

### DIFF
--- a/packages/nvidia_nat_mcp/src/nat/plugins/mcp/client_base.py
+++ b/packages/nvidia_nat_mcp/src/nat/plugins/mcp/client_base.py
@@ -163,6 +163,7 @@ class MCPBaseClient(ABC):
             raise ValueError("transport must be either 'sse', 'stdio' or 'streamable-http'")
 
         self._exit_stack: AsyncExitStack | None = None
+        self._context_manager = None  # Explicit context manager for proper task isolation
         self._session: ClientSession | None = None  # Main session
         self._connection_established = False
         self._initial_connection = False
@@ -193,26 +194,51 @@ class MCPBaseClient(ABC):
         return self._transport
 
     async def __aenter__(self):
-        if self._exit_stack:
+        if self._context_manager:
             raise RuntimeError("MCPBaseClient already initialized. Use async with to initialize.")
-
-        self._exit_stack = AsyncExitStack()
-
-        # Establish connection with httpx.Auth
-        self._session = await self._exit_stack.enter_async_context(self.connect_to_server())
-
-        self._initial_connection = True
-        self._connection_established = True
-
-        return self
-
+    
+        try:
+            # Instead of using AsyncExitStack, directly manage the context manager.
+            # This ensures entry and exit happen in the same task context, avoiding
+            # anyio's "cancel scope entered in different task" error.
+            self._context_manager = self.connect_to_server()
+            self._session = await self._context_manager.__aenter__()
+            
+            self._initial_connection = True
+            self._connection_established = True
+            
+            return self
+        except Exception:
+            # If initialization fails, clean up the context manager
+            if hasattr(self, '_context_manager') and self._context_manager:
+                try:
+                    await self._context_manager.__aexit__(None, None, None)
+                except Exception:
+                    pass  # Suppress cleanup errors on init failure
+            raise
+    
     async def __aexit__(self, exc_type, exc_value, traceback):
-        if self._exit_stack:
-            # Close session
-            await self._exit_stack.aclose()
-            self._session = None
-            self._exit_stack = None
-
+        if hasattr(self, '_context_manager') and self._context_manager:
+            try:
+                # Exit the context manager directly.
+                # This ensures exit happens in the same task as entry.
+                await self._context_manager.__aexit__(exc_type, exc_value, traceback)
+            except RuntimeError as e:
+                if "cancel scope" in str(e).lower():
+                    # Known issue with MCP SDK's stdio transport crossing task boundaries.
+                    # Log the error but don't crash - the context is still cleaned up.
+                    logger.warning(
+                        "MCP client cleanup: Ignoring cancel scope task mismatch error "
+                        "(upstream MCP SDK issue with anyio context switching): %s", 
+                        e
+                    )
+                else:
+                    # Re-raise other RuntimeErrors
+                    raise
+            finally:
+                self._context_manager = None
+                self._session = None
+    
         self._connection_established = False
         self._tools = None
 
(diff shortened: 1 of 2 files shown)


willkill07:
/ok to test [4fd5bc5](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1271/commits/4fd5bc591fb49db622b7b097a59d0e75699a12d4)

packages/nvidia_nat_mcp/src/nat/plugins/mcp/tool.py
```diff
@@ -85,17 +85,21 @@ def _convert_from_str(input_str: str) -> tool.input_schema:
...
-            return await tool.acall(kwargs)
-        except Exception as e:
-            logger.warning("Error calling tool %s", tool.name, exc_info=True)
-            return str(e)
+         # Run the tool, catching any errors and sending to agent for correction
+         try:
+             if tool_input:
+                 # Exclude None values for optional parameters.
+                 # MCP server will use server-side defaults for omitted optional fields.
+                 args = tool_input.model_dump(exclude_none=True)
```
AnuradhaKaruppiah, line 93: This fix makes sense and has already been applied to the mcp_client function group. It wasn't added to mcp_tool wrapper as this function has been deprecated and is no longer maintained.

AnuradhaKaruppiah:
@sjarmak Thx for sending this PR. I will recommend switching mcp_client function group which shouldn't have either issue. mcp_tool_wrapper function has been deprecated and is no longer maintained.

packages/nvidia_nat_mcp/src/nat/plugins/mcp/client_base.py
```diff
@@ -193,26 +194,51 @@ def transport(self) -> str:
         return self._transport
 
     async def __aenter__(self):
-        if self._exit_stack:
+        if self._context_manager:
```
AnuradhaKaruppiah, line 197: AsyncExitStack is still needed to handled mcp_client function group in its current form. @sjarmak can you pls migrate your agent from mcp_tool_wrapper to mcp_client and let us know if this is an issue there?

sjarmak: ah okay the fix wasn't present in a pip installed version I was working off of; built from source and migrated to mcp_cilent and can confirm it works there, thanks!

AnuradhaKaruppiah: awesome, thx for confirming. I will pull the plug on mcp_tool_wrapper in 1.5 (maybe even in 1.4 after checking with users) to avoid this confusion.
Stephanie, is there any other change in this PR that is still needed or is it OK to close?

sjarmak: sounds good, thanks, I think we're good so I'll close the PR

PR #1277, opened 2025-12-16
Fix MCP workflow entry function handling

mpenn:
## Description

Closes

This PR fixes an issue where `nat mcp serve` would fail with `Error: Function '<workflow_name>' not found` when using any workflow type in the configuration, while `nat serve` worked correctly with the same config. The root cause was that the MCP front-end worker was incorrectly passing the workflow name as the `entry_function` parameter to `SessionManager.create()`, causing the builder to search for a function with that name instead of using the workflow itself. Additionally, where nat mcp serve would fail with isinstance() argument 2 cannot be a parameterized generic when calling functions that return dict or list types, while nat serve (FastAPI) worked correctly with the same config. The root cause was that the TypeConverter was using isinstance() directly with parameterized generic types (like dict[str, Any]), which is not supported in Python, and the MCP tool wrapper was unnecessarily forcing type conversion to str before the result serialization logic could handle it properly.

- Modified `mcp_front_end_plugin_worker.py` to pass `entry_function=None` when creating SessionManagers for workflows (e.g., `react_agent`, `rewoo_agent`, etc.)
- Added unit test `test_session_manager_creation_for_workflow_vs_function` to verify correct parameter handling for both workflows and regular functions
- Modified src/nat/utils/type_converter.py to use `DecomposedType.is_instance()` instead of direct `isinstance()` calls in `_try_indirect_conversion()`, allowing safe handling of parameterized generics
- Modified `src/nat/front_ends/mcp/tool_converter.py` to remove `to_type=str` parameter from `runner.result()` call, letting the existing JSON serialization logic handle dict/list conversion
- Added tests in `tests/nat/utils/test_converter.py` to verify correct handling of parameterized generic types (dict[str, Any], list[int], etc.)
- Added tests in `tests/nat/front_ends/mcp/test_tool_converter.py` to verify correct result handling without type coercion and proper dict/list JSON serialization

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced support for parameterized generic types (e.g., dict[str, Any], list[int]) in type conversions
  * Improved result type handling for workflows with proper type inference

* **Tests**
  * Added comprehensive test coverage for generic type handling and result conversions


dagardner-nv (approved):
Thanks for adding all those tests!

AnuradhaKaruppiah (approved):

PR #1282, opened 2025-12-17
Remove `default_user_id` from `GeneralConfig` to prevent unsafe per-user workflow sharing

yczhang-nv:
## Description

The `default_user_id` field in `GeneralConfig` was designed to support single-user use cases like `nat run` and `nat eval` with per-user workflow lazy instantiation. However, when used with `nat serve`, this configuration is unsafe—if enabled without careful attention, all users would share the same `default_user_id`, causing requests from multiple users to unintentionally use the same workflow instance.

- For `nat run` and `nat eval`: Add a `user_id` field to their respective configs (`ConsoleFrontEndConfig` and `EvaluationRunConfig`) with explicit default values (`nat_run_user_id` and `nat_eval_user_id`), making it clear these are single-user contexts.
- For nat serve (per-user lazy init): Extract the `user_id` from the request context (via `nat-session` cookie). If no `user_id` can be determined, fail the per-user function/function group initialization and reject the request.

With this change, HTTP requests to per-user workflows via nat serve will be rejected if no `nat-session` cookie is set, since the `user_id` cannot be determined. This ensures multi-user deployments don't accidentally share workflow state.

This PR is breaking since it removes the `default_user_id` field from `GeneralConfig`. The impact should be small since the field was just added in recent [PR-1206](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1206).

Closes

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearer error message when a workflow user ID cannot be determined.

* **Improvements**
  * User ID handling standardized: fields now have concrete defaults for CLI/evaluation runs, reducing ambiguous missing values and simplifying initialization.
  * Session resolution prefers explicit user_id and fails fast if still unset.

* **Chores**
  * Removed legacy fallback default user ID from configuration to streamline settings.

### DIFF
--- a/src/nat/cli/commands/evaluate.py
+++ b/src/nat/cli/commands/evaluate.py
@@ -195,16 +195,19 @@ def process_nat_eval(
                                "have a partially completed dataset.")
 
     # Create the configuration object
-    config = EvaluationRunConfig(
-        config_file=config_file,
-        dataset=str(dataset) if dataset else None,
-        result_json_path=result_json_path,
-        skip_workflow=skip_workflow,
-        skip_completed_entries=skip_completed_entries,
-        endpoint=endpoint,
-        endpoint_timeout=endpoint_timeout,
-        reps=reps,
-        override=override,
-        user_id=user_id,
-    )
+    # Only include user_id if explicitly provided via CLI, otherwise use the default
+    config_kwargs = {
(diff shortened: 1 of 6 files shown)


AnuradhaKaruppiah (approved):
Thanks for making this change

PR #1313, opened 2025-12-30
Fix base64 image support in ChatRequest to match OpenAI API specification

mpenn:
## Description

Closes

This PR fixes NAT's OpenAI Chat Completions API implementation to properly support base64-encoded images in the `data:image/*;base64,...` format. The root issue was that `ImageUrl.url` was typed as Pydantic's `HttpUrl`, which only accepts http/https schemes and rejects data URIs. This has been changed to plain `str` to match OpenAI's official implementation, which accepts both regular URLs and base64-encoded data URIs. The fix enables full compatibility with OpenAI's vision API and allows the NAT UI's base64 image encoding to work end-to-end. All changes are backward compatible - HTTP/HTTPS URLs continue to work as before.

## Summary by CodeRabbit

* **Refactor**
  * Modified image URL field structure to improve handling and default value management.

### DIFF
--- a/src/nat/data_models/api_server.py
+++ b/src/nat/data_models/api_server.py
@@ -24,7 +24,6 @@ from pydantic import BaseModel
 from pydantic import ConfigDict
 from pydantic import Discriminator
 from pydantic import Field
-from pydantic import HttpUrl
 from pydantic import conlist
 from pydantic import field_serializer
 from pydantic import field_validator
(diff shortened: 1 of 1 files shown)


src/nat/data_models/api_server.py
```diff
@@ -90,7 +89,8 @@ class AudioContent(BaseModel):
 
 
 class ImageUrl(BaseModel):
-    url: HttpUrl = HttpUrl(url="http://default.com")
+    url: str = Field(default="http://default.com",
```
willkill07, line 92: Nit: could this be a Pydantic `AnyUrl` instead of a `str`? Otherwise some amount of validation would also need performed.

willkill07 (approved):

PR #1357, opened 2026-01-07
bug fix: Dynamo SGLang Startup Script Cleanup

bbednarski9:
## Description
This PR addresses a bug in the /external/dynamo startup scripts for unified and disaggregated serving that was preventing the router from starting up. Router used deprecated python methods from a previous version of NVIDIA dynamo that are not supported in Dynamo 0.6.1 (docker image nvcr.io/nvidia/ai-dynamo/sglang-runtime:0.6.1). Removing these deprecated pycls from the dynamo components allows startup.

We also extend the timeout period for API model cold start from 1.5 to 15 minutes to account for potential increased API cold start latency (early exit when /health GET endpoint is up). We have also bumped up the max timeout period for worker initialization from 2 to 5 minutes with early exit.

I have also added polling with exponential backoff to the startup of generalized/router.py, updated copywrite/license years, and the demo instructions.
Closes

## Summary by CodeRabbit

* **Documentation**
  * Expanded setup and download docs with a Python venv workflow, new model-download steps, metrics CSV appendix, KV-overlap routing guidance, and clearer startup paths.

* **Bug Fixes**
  * Improved startup readiness and validation with ETCD-backed worker registration, model-directory checks, endpoint probing, and extended timeouts for large models.

* **New Features**
  * Frontend accepts configurable model-path mappings; router adds KV-overlap fallbacks and tolerates missing metrics while preserving CSV tracing.

* **Chores**
  * Activated example GPU device list, added router metrics CSV setting, and bumped runtime image versions.

### DIFF
--- a/external/dynamo/generalized/router.py
+++ b/external/dynamo/generalized/router.py
@@ -311,7 +311,7 @@ class WorkloadAwareRouter:
         if min_workers < 0:
             raise ValueError(f"min_workers must be >= 0, got {min_workers}")
 
-        timeout_s = float(os.environ.get("DYNAMO_ROUTER_WAIT_FOR_WORKERS_TIMEOUT_S", "300"))
+        timeout_s = float(os.environ.get("DYNAMO_ROUTER_WAIT_FOR_WORKERS_TIMEOUT_S", "600"))
         if not math.isfinite(timeout_s) or timeout_s <= 0:
             raise ValueError("DYNAMO_ROUTER_WAIT_FOR_WORKERS_TIMEOUT_S must be a finite number > 0 "
                              f"(got {timeout_s!r})")
--- a/external/dynamo/generalized/router.py
+++ b/external/dynamo/generalized/router.py
@@ -1,5 +1,17 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA
+# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
 import argparse
 import asyncio
@@ -9,50 +21,297 @@ import logging
 import math
 import os
 import random
+import re
 import threading
 import time
 import uuid
 from collections import deque
+from dataclasses import dataclass
 from functools import wraps
 from typing import Any
 
+import aiohttp
 import numpy as np
 import uvloop
+from dynamo.llm import KvIndexer
+from dynamo.llm import OverlapScores
 from dynamo.runtime import DistributedRuntime
 from dynamo.runtime import dynamo_worker
 from dynamo.runtime.logging import configure_dynamo_logging
 from pydantic import BaseModel
 
-# Try to import KV routing classes from dynamo.llm, fallback to stubs if unavailable
-try:
-    from dynamo.llm import KvIndexer
-    from dynamo.llm import OverlapScores
-except ImportError:
-    logger_init = logging.getLogger(__name__)
-    logger_init.warning("dynamo.llm KV classes not available, using fallback implementations")
-
-    class OverlapScores:
-        """Fallback: KV cache overlap scores between a request and workers."""
-
-        def __init__(self, scores: dict[int, float] | None = None):
-            self.scores = scores if scores is not None else {}
-
-    class KvIndexer:
-        """Fallback: KV cache indexer for finding overlap between requests and workers."""
-
-        def __init__(self, engine: Any, block_size: int):
-            self.engine = engine
-            self.block_size = block_size
-
-        async def find_matches_for_request(self, tokens: list[int], min_overlap: int) -> OverlapScores:
-            """Find overlap scores for each worker. Returns empty scores (round-robin fallback)."""
-            return OverlapScores({})
-
-
 configure_dynamo_logging()
 logger = logging.getLogger(__name__)
 
 WorkerId = int
+
+
+# ---------------------- metrics shim for 0.7.1+ ---------------------- #
+# Replaces deprecated KvMetricsAggregator/AggregatedMetrics with Prometheus scraping
+@dataclass
+class EndpointMetrics:
+    """Per-worker metrics, compatible with old AggregatedMetrics.endpoints shape."""
+
+    worker_id: int
+    gpu_cache_usage_perc: float  # 0-100 scale
+    num_requests_waiting: int
+
+
+@dataclass
+class AggregatedMetricsShim:
+    """Shim to maintain API compatibility with old AggregatedMetrics."""
+
+    endpoints: list[EndpointMetrics]
+
+
+class PrometheusMetricsClient:
+    """
+    Scrapes Dynamo's Prometheus /metrics endpoint and provides per-worker
+    KV cache usage and queue depth in the format expected by WorkloadAwareRouter.
+
+    Metrics used (0.7.1+):
+      - dynamo_frontend_queued_requests{worker_id=...}
+      - dynamo_frontend_inflight_requests{worker_id=...}
+      - dynamo_frontend_model_total_kv_blocks{worker_id=...}
+      - dynamo_component_kv_blocks_used{worker_id=...} (if available)
+
+    Falls back to estimating KV usage from inflight requests if direct metrics unavailable.
+    """
+
+    # Prometheus metric patterns
+    _METRIC_PATTERN = re.compile(
+        r'^(?P<name>[a-zA-Z_:][a-zA-Z0-9_:]*)'
+        r'(?:\{(?P<labels>[^}]*)\})?\s+'
+        r'(?P<value>[+-]?(?:\d+\.?\d*|\.\d+)(?:[eE][+-]?\d+)?|NaN|[+-]Inf)$'
+    )
+    _LABEL_PATTERN = re.compile(r'(\w+)="([^"]*)"')
+
+    def __init__(
+        self,
+        metrics_url: str | None = None,
+        scrape_interval_seconds: float = 1.0,
+        request_timeout_seconds: float = 5.0,
+        default_total_kv_blocks: int = 8192,
+        estimated_kv_blocks_per_request: int = 128,
+    ):
+        """
+        Args:
+            metrics_url: Prometheus metrics endpoint URL (e.g., http://localhost:9090/metrics).
+                         If None, uses DYNAMO_METRICS_URL env var or defaults to localhost:9090.
+            scrape_interval_seconds: Minimum interval between scrapes to avoid hammering.
+            request_timeout_seconds: HTTP request timeout.
+            default_total_kv_blocks: Fallback total KV blocks if metric unavailable.
+            estimated_kv_blocks_per_request: Estimated KV blocks per inflight request for usage calc.
+        """
+        self.metrics_url = metrics_url or os.environ.get(
+            "DYNAMO_METRICS_URL", "http://localhost:9090/metrics"
+        )
+        self.scrape_interval = float(scrape_interval_seconds)
+        self.request_timeout = float(request_timeout_seconds)
+        self.default_total_kv_blocks = int(default_total_kv_blocks)
+        self.estimated_kv_per_request = int(estimated_kv_blocks_per_request)
+
+        # Cache for rate limiting
+        self._cache_lock = threading.Lock()
+        self._cached_metrics: AggregatedMetricsShim | None = None
+        self._last_scrape_time: float = 0.0
+
+        # Persistent session (created lazily)
+        self._session: aiohttp.ClientSession | None = None
+
+    async def _get_session(self) -> aiohttp.ClientSession:
+        if self._session is None or self._session.closed:
+            timeout = aiohttp.ClientTimeout(total=self.request_timeout)
+            self._session = aiohttp.ClientSession(timeout=timeout)
+        return self._session
+
+    async def close(self):
+        """Close the HTTP session."""
+        if self._session and not self._session.closed:
+            await self._session.close()
+            self._session = None
+
+    def _parse_prometheus_text(self, text: str) -> dict[str, list[tuple[dict[str, str], float]]]:
+        """
+        Parse Prometheus text format into structured data.
+        Returns: {metric_name: [(labels_dict, value), ...]}
+        """
+        result: dict[str, list[tuple[dict[str, str], float]]] = {}
+
+        for line in text.splitlines():
+            line = line.strip()
+            if not line or line.startswith('#'):
+                continue
+
+            match = self._METRIC_PATTERN.match(line)
+            if not match:
+                continue
+
+            name = match.group('name')
+            labels_str = match.group('labels') or ''
+            value_str = match.group('value')
+
+            # Parse labels
+            labels = dict(self._LABEL_PATTERN.findall(labels_str))
+
+            # Parse value
+            try:
+                if value_str == 'NaN':
+                    value = float('nan')
+                elif value_str in ('+Inf', 'Inf'):
+                    value = float('inf')
+                elif value_str == '-Inf':
+                    value = float('-inf')
+                else:
+                    value = float(value_str)
+            except ValueError:
+                continue
+
+            if name not in result:
+                result[name] = []
+            result[name].append((labels, value))
+
+        return result
+
+    def _extract_worker_id(self, labels: dict[str, str]) -> int | None:
+        """Extract worker_id from Prometheus labels."""
+        # Try various label names that might contain worker ID
+        for key in ('worker_id', 'instance_id', 'worker', 'instance', 'replica'):
+            if key in labels:
+                try:
+                    return int(labels[key])
+                except ValueError:
+                    # Try extracting number from string like "worker-0"
+                    match = re.search(r'(\d+)', labels[key])
+                    if match:
+                        return int(match.group(1))
+        return None
+
+    def _build_metrics_from_prometheus(
+        self, parsed: dict[str, list[tuple[dict[str, str], float]]]
+    ) -> AggregatedMetricsShim:
+        """Build AggregatedMetricsShim from parsed Prometheus data."""
+
+        # Collect per-worker data
+        worker_data: dict[int, dict[str, float]] = {}
+
+        # Helper to ensure worker entry exists
+        def ensure_worker(wid: int):
+            if wid not in worker_data:
+                worker_data[wid] = {
+                    'queued': 0.0,
+                    'inflight': 0.0,
+                    'total_kv_blocks': float(self.default_total_kv_blocks),
+                    'used_kv_blocks': 0.0,
+                }
+
+        # Process queued requests
+        for labels, value in parsed.get('dynamo_frontend_queued_requests', []):
+            wid = self._extract_worker_id(labels)
+            if wid is not None:
+                ensure_worker(wid)
+                worker_data[wid]['queued'] = max(0.0, value)
+
+        # Process inflight requests
+        for labels, value in parsed.get('dynamo_frontend_inflight_requests', []):
+            wid = self._extract_worker_id(labels)
+            if wid is not None:
+                ensure_worker(wid)
+                worker_data[wid]['inflight'] = max(0.0, value)
+
+        # Process total KV blocks
+        for labels, value in parsed.get('dynamo_frontend_model_total_kv_blocks', []):
+            wid = self._extract_worker_id(labels)
+            if wid is not None:
+                ensure_worker(wid)
+                worker_data[wid]['total_kv_blocks'] = max(1.0, value)
+
+        # Process used KV blocks (if available)
+        for labels, value in parsed.get('dynamo_component_kv_blocks_used', []):
+            wid = self._extract_worker_id(labels)
+            if wid is not None:
+                ensure_worker(wid)
+                worker_data[wid]['used_kv_blocks'] = max(0.0, value)
+
+        # Also check gpu_cache_usage_perc if directly available
+        for labels, value in parsed.get('dynamo_frontend_gpu_cache_usage_perc', []):
+            wid = self._extract_worker_id(labels)
+            if wid is not None:
+                ensure_worker(wid)
+                worker_data[wid]['gpu_cache_usage_perc_direct'] = max(0.0, min(100.0, value))
+
+        # Build endpoint metrics
+        endpoints: list[EndpointMetrics] = []
+        for wid, data in worker_data.items():
+            # Calculate GPU cache usage percentage
+            if 'gpu_cache_usage_perc_direct' in data:
+                gpu_cache_usage = data['gpu_cache_usage_perc_direct']
+            elif data['used_kv_blocks'] > 0:
+                # Use actual used KV blocks if available
+                gpu_cache_usage = 100.0 * data['used_kv_blocks'] / data['total_kv_blocks']
+            else:
+                # Estimate from inflight requests
+                estimated_used = data['inflight'] * self.estimated_kv_per_request
+                gpu_cache_usage = 100.0 * min(estimated_used, data['total_kv_blocks']) / data['total_kv_blocks']
+
+            gpu_cache_usage = max(0.0, min(100.0, gpu_cache_usage))
+
+            endpoints.append(EndpointMetrics(
+                worker_id=wid,
+                gpu_cache_usage_perc=gpu_cache_usage,
+                num_requests_waiting=int(data['queued']),
+            ))
+
+        return AggregatedMetricsShim(endpoints=endpoints)
+
+    async def get_metrics(self) -> AggregatedMetricsShim:
+        """
+        Fetch and return aggregated metrics. Uses caching to rate-limit scrapes.
+        """
+        now = time.time()
+
+        # Check cache
+        with self._cache_lock:
+            if (
+                self._cached_metrics is not None
+                and (now - self._last_scrape_time) < self.scrape_interval
+            ):
+                return self._cached_metrics
+
+        # Scrape Prometheus
+        try:
+            session = await self._get_session()
+            async with session.get(self.metrics_url) as response:
+                if response.status != 200:
+                    logger.warning(
+                        "Prometheus scrape failed: HTTP %d from %s",
+                        response.status,
+                        self.metrics_url,
+                    )
+                    # Return cached or empty
+                    with self._cache_lock:
+                        return self._cached_metrics or AggregatedMetricsShim(endpoints=[])
+
+                text = await response.text()
+                parsed = self._parse_prometheus_text(text)
+                metrics = self._build_metrics_from_prometheus(parsed)
+
+                with self._cache_lock:
+                    self._cached_metrics = metrics
+                    self._last_scrape_time = now
+
+                return metrics
+
+        except asyncio.TimeoutError:
+            logger.warning("Prometheus scrape timed out: %s", self.metrics_url)
+        except aiohttp.ClientError as e:
+            logger.warning("Prometheus scrape error: %s - %s", self.metrics_url, e)
+        except Exception as e:
+            logger.warning("Unexpected error scraping metrics: %s", e)
+
+        # Return cached or empty on error
+        with self._cache_lock:
+            return self._cached_metrics or AggregatedMetricsShim(endpoints=[])
 
 
 # ---------------------- request / response models ---------------------- #
@@ -169,6 +428,7 @@ class WorkloadAwareRouter:
         # clients / helpers (initialized later)
         self.engine_client = None
         self.indexer: KvIndexer | None = None
+        self.metrics: PrometheusMetricsClient | None = None
 
         # concurrency primitives
         self._init_lock = threading.Lock()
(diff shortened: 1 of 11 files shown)


external/dynamo/start_dynamo_unified_thompson_hints.sh
```diff
@@ -272,41 +273,41 @@ docker run -d \
...
-    
+
     echo ''
-    
+
     # Function to wait for worker initialization
     wait_for_worker() {
         local worker_type=\$1
         local pid=\$2
         local max_wait=120
```
mnajafian-nv: PR description states "5 minutes maximum" but this is 2 minutes. Which is correct? 

Suggestion: `local max_wait=300  # 5 minutes for cold starts`

external/dynamo/start_dynamo_unified_thompson_hints.sh
```diff
@@ -272,41 +273,41 @@ docker run -d \
...
-            
+
             sleep 5
             elapsed=\$((elapsed + 5))
             if [ \$((elapsed % 15)) -eq 0 ]; then
                 echo \"  ... \${elapsed}s / \${max_wait}s\"
             fi
-            
+
             # After 60s, assume it's initialized (model loading takes time for 70B)
```
mnajafian-nv: The sleep 60 waits for time rather than checking worker readiness. Large models (70B) can take 3-5 minutes to load on cold start. As a result router starts at T+90s, but worker is still loading, hence  connection fails.
My suggestion is to poll the worker endpoint until it's ready instead of using a fixed delay, what do you think? Happy to discuss! 😊

Suggestion
```
# Don't return early - let it wait full max_wait
# OR add verification:
if [ $elapsed -ge 60 ]; then
    # Check actual readiness
    if python3 -c "import requests; requests.get('http://localhost:30000/health', timeout=2)" 2>/dev/null; then
        echo "✓ Worker verified via health check"
        return 0
    fi
    # Otherwise keep waiting
fi
```

external/dynamo/start_dynamo_unified_thompson_hints.sh
```diff
@@ -322,15 +323,15 @@ docker run -d \
     WORKER_PID=\$!
     echo \"Unified Worker PID: \$WORKER_PID\"
     echo \"\"
-    
+
     # Wait for unified worker to initialize
     wait_for_worker \"Unified\" \$WORKER_PID || exit 1
-    
+
     # Give worker extra time to register with ETCD
```
mnajafian-nv: This comment says "Give worker extra time to register with ETCD" but doesn't verify registration happened. I personally think this creates a false sense of security, from what i can see here the script continues regardless of actual ETCD state.  What happens if ETCD is unreachable? I am concerned that the script would silently proceed and fail later with cryptic errors

external/dynamo/start_dynamo_disagg.sh
```diff
@@ -236,41 +236,41 @@ docker run -d \
...
-    
+
     echo ''
-    
+
     # Function to wait for worker initialization
     wait_for_worker() {
         local worker_type=\$1
         local pid=\$2
         local max_wait=120
```
mnajafian-nv: From what i see here disaggregated mode has 2 workers (prefill + decode). Both get 120s, but they run sequentially, not in parallel, so the total time before frontend starts: ~240s. Is this intentional? Can we consider monitoring in parallel to reduce total startup time?

external/dynamo/start_dynamo_disagg.sh
```diff
@@ -236,41 +236,41 @@ docker run -d \
...
             elapsed=\$((elapsed + 5))
             if [ \$((elapsed % 15)) -eq 0 ]; then
                 echo \"  ... \${elapsed}s / \${max_wait}s\"
             fi
-            
+
             # After 60s, assume it's initialized (model loading takes time for 70B)
             if [ \$elapsed -ge 60 ]; then
                 echo \"✓ \$worker_type worker should be initialized\"
                 return 0
```
mnajafian-nv: Is the 60s assumption valid? I think assuming success after 60s for each increases the chance of race conditions, I'm extra concerned for disagg mode: Prefill and decode workers must BOTH be ready.

external/dynamo/start_dynamo_disagg.sh
```diff
@@ -311,15 +311,15 @@ docker run -d \
...
     echo \"\"
-    
+
     # Wait for decode worker to initialize
     wait_for_worker \"Decode\" \$DECODE_PID || exit 1
-    
+
     # Give workers extra time to register with ETCD
     echo ''
     echo 'Waiting for workers to register with ETCD (30s)...'
```
mnajafian-nv: After starting 2 workers sequentially + 2×30s sleeps, we're at ~180s before frontend starts.

Q: For disaggregated mode, do both workers need to register in ETCD before frontend starts? Current code assumes yes but doesn't verify.

I recommend adding a check that both workers are registered:

```
if etcdctl get /dynamo/workers/ --prefix | grep -c "worker" | grep -q "^2$"; then
    echo "✓ Both workers registered"
else
    echo "Expected 2 workers, frontend may not work correctly"
fi
```

mnajafian-nv (changes requested):
Thanks for tackling NVBug-5756833! I've left detailed inline comments on specific lines.
I appreciate the PID monitoring, consistent implementation across scripts. Here are my main areas of concern
-  Issue: 2min vs 5min discrepancy  
-  Critical: Assumes success after 60s without verification - won't handle cold starts

My recommendation is to fix the timeout discrepancy and remove the 60s early return. I'll push a follow-up PR adding retry logic to `router.py` to handle the Python-side race condition.

With your bash improvements + my Python retry in my next PR we can have a  comprehensive solution.

dnandakumar-nv:
Without overlap scores, the router loses any insight into KV cache match, so it won't take that into account when routing.
Suggest we add to documentation this failure mode and in what configuration the KV cache won't be accounted for.

external/dynamo/start_dynamo_unified_thompson_hints.sh
```diff
@@ -552,15 +553,15 @@ if docker ps --format '{{.Names}}' | grep -q "^${CONTAINER_NAME}$"; then
...
-    echo "Waiting for SGLang to initialize (this may take 90-120 seconds for 70B model)..."
+    echo "Waiting for SGLang to initialize (this may take 5-10 minutes for a 70B model)..."
     echo "Monitoring logs (Ctrl+C to exit, container continues)..."
     echo ""
-    
+
     # Wait for server to be ready
     echo "Checking for API availability..."
-    max_attempts=90
+    max_attempts=900
```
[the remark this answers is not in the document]
bbednarski9: changed sleep from 2->1 so that max_attempts=timeout(in seconds). Edited print statements for consistency

external/dynamo/start_dynamo_unified_thompson_hints.sh
bbednarski9: @mnajafian-nv  This is a different timeout than what I was referring to in the PR description. I was referring to the max wait time on the endpoint: `curl -s http://localhost:$HTTP_PORT/health` at the end of each dynamo startup script, which is where I saw the startup failure.

This is also valid through, see commit 6a12a21. I'll increase timeout periods for worker startup in the docker container with early stopping as well.

external/dynamo/start_dynamo_unified_thompson_hints.sh
bbednarski9: check out [84b3319](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1357/commits/84b3319c1da9c178ba6c2380bc6e1bb6449b8ab9)

For the non-custom dynamo frontend servers we can request the /v1/models endpoint for availability. For the custom frontend we dont have that endpoint (yet) but we can use the /health endpoint and improve polling.

external/dynamo/start_dynamo_unified_thompson_hints.sh
bbednarski9: See `start_dynamo_unified_thompson_hints.sh` line 302, I now verify dynamo worker registration directory on ETCD:

```
local etcd_response=\$(curl -s --max-time 2 http://localhost:2379/v3/kv/range \
                -X POST \
                -H \"Content-Type: application/json\" \
                -d '{\"key\":\"AA==\",\"range_end\":\"AA==\",\"keys_only\":true}' 2>&1)
```

external/dynamo/start_dynamo_disagg.sh
bbednarski9: Added parallel startup polling to start_dynamo_disagg.sh

external/dynamo/start_dynamo_disagg.sh
bbednarski9: This has been removed in favor of parallel startup polling

external/dynamo/start_dynamo_disagg.sh
bbednarski9: wait_for_workers_parallel() function (lines 384-423) actively polls ETCD via the v3 API, checking that current_count >= 2 before returning success. The frontend only starts after wait_for_workers_parallel || exit 1 succeeds (line 425), ensuring both workers are registered. This replaces the arbitrary sleep-based approach with explicit verification — the script will either confirm "✓ Both workers registered in ETCD" or fail with a timeout/error before attempting to start the frontend.

mnajafian-nv:
Thanks for the great work @bbednarski9 @dnandakumar-nv . Just confirming my understanding of the root cause: `router.py:initialize()` calls `await self.engine_client.wait_for_instances()` without timeout/retry, which blocks indefinitely when workers aren't immediately registered in ETCD. Your bash improvements (5-min timeout, ETCD polling, health checks) give workers adequate time to initialize before the router starts, which handles most cases. However, the router code itself still has no resilience if workers take longer than 5 minutes or register just after your bash timeout completes. To fully resolve the NVBug, we need to add retry logic with exponential backoff to that `wait_for_instances()` call in `router.py`. If my understanding is accurate, since I drafted this yesterday I can either submit a follow-up PR after yours merges, or if you prefer, you can add it to this PR and I'll review. Let me know which approach you'd like :)

bbednarski9:
Hi @mnajafian-nv, the original root cause failure was just that the environment variables for paths to the model endpoint needed some patching. However, after your review, I started digging into some of the startup reliability issues. I pushed some changes that reflect your comments regarding polling (for both unified scripts) and parallel polling (for the disagg script) that should greatly improve the startup reliability of the servers. This also includes improved worker registration checking to validate that backend components come up before we try to bring up the frontend and API. Overall, its starting up noticeably faster now.

Therefore, the startup issues from the reported bug were not from the generalized/router.py code. I discovered that while doing additional testing. I'm going to look into that next because the metrics are used by Dhruv's learner for intelligent routing, and some of the SDK classes that we're used in Dhruv's earlier version have been deprecated. So we need to find alternatives. Per Dhruv's comment, if we use dummy data or stubs for these, the features will be useless. Ill follow up soon. It's possible that we need to upgrade to the dynamo-sglang docker image for version 0.7.1.

external/dynamo/start_dynamo_unified_thompson_hints.sh
```diff
@@ -61,8 +61,9 @@ LOCAL_MODEL_DIR="${DYNAMO_MODEL_DIR}"
 
 # Repository directory - auto-detect from script location or use env var
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-LOCAL_REPO_DIR="${DYNAMO_REPO_DIR:-$(cd "$SCRIPT_DIR/../.." && pwd)}"
-CUSTOM_DYNAMO_DIR="$LOCAL_REPO_DIR/external/dynamo/generalized"
+LOCAL_REPO_DIR="${DYNAMO_REPO_DIR:-$SCRIPT_DIR}"
+# Custom dynamo components are always relative to the script location (external/dynamo/generalized)
+CUSTOM_DYNAMO_DIR="${SCRIPT_DIR}/generalized"
```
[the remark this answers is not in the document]
bbednarski9: resolved

bbednarski9:
Discussed directly with @mnajafian-nv

I will take on the router learner comments and also look into startup of router.py, then request review

external/dynamo/start_dynamo_unified.sh
```diff
@@ -37,7 +37,7 @@ TP_SIZE="${DYNAMO_TP_SIZE:-4}"
 HTTP_PORT="${DYNAMO_HTTP_PORT:-8099}"
 MODEL="/workspace/models/Llama-3.3-70B-Instruct"
 SERVED_MODEL_NAME="${DYNAMO_MODEL_NAME:-llama-3.3-70b}"
-IMAGE="nvcr.io/nvidia/ai-dynamo/sglang-runtime:0.6.1"
+IMAGE="nvcr.io/nvidia/ai-dynamo/sglang-runtime:0.7.1"
```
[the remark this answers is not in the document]
bbednarski9: Not correct, ignoring

external/dynamo/generalized/router.py
```diff
@@ -271,17 +280,21 @@ def _iat_factor(iat: str) -> float:
...
+        logger.info("Backend workers discovered: %s", list(self.engine_client.instance_ids()))
 
         self.indexer = KvIndexer(engine, self.block_size)
-        self.metrics = KvMetricsAggregator(engine)
 
         self._initialize_bandits()
         self._initialize_contextual()
-        logger.info("WorkloadAwareRouter initialized")
+        logger.info("WorkloadAwareRouter initialized with %d backend worker(s)",
+                    len(list(self.engine_client.instance_ids())))
```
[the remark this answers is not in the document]
bbednarski9: we can do this in a follow-up PR, but ultimately there is a router refactor needed to account for the frontend metrics collection so this will be duplicate work

bbednarski9 (commented):
Addressed the comments above, deferring further router.py development to a follow-up PR

external/dynamo/generalized/router.py
```diff
@@ -30,15 +30,47 @@
...
+        """
+
+        def __init__(self, engine: Any, block_size: int):
+            self.engine = engine
+            self.block_size = block_size
+
+        async def find_matches_for_request(self, tokens: list[int], min_overlap: int) -> OverlapScores:  # noqa: ARG002
+            """Find overlap scores for each worker. Returns empty scores (round-robin fallback)."""
+            return OverlapScores({})
+
```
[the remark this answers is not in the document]
bbednarski9: This is NA for now, as we will revisit this when the router is refactored. The failure case is already documented

external/dynamo/generalized/router.py
```diff
@@ -271,17 +302,21 @@ def _iat_factor(iat: str) -> float:
...
+        logger.info("Backend workers discovered: %s", list(self.engine_client.instance_ids()))
 
         self.indexer = KvIndexer(engine, self.block_size)
-        self.metrics = KvMetricsAggregator(engine)
 
         self._initialize_bandits()
         self._initialize_contextual()
-        logger.info("WorkloadAwareRouter initialized")
+        logger.info("WorkloadAwareRouter initialized with %d backend worker(s)",
+                    len(list(self.engine_client.instance_ids())))
```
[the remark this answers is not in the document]
bbednarski9: addressed by [a845380](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1357/commits/a845380cffc21c68805d6659cc87061b0faa8382)

external/dynamo/start_dynamo_unified.sh
```diff
@@ -242,41 +242,64 @@ docker run -d \
...
-        
-        echo \"WARNING: \$worker_type worker initialization timeout, proceeding anyway\"
-        return 0
+
+        echo \"ERROR: \$worker_type worker failed to register with ETCD within \${max_wait}s\"
+        echo \"  Image: $IMAGE\"
+        echo \"  The model may require more time to load, or there may be a startup error.\"
+        echo \"  Check worker logs for details.\"
+        return 1
     }
```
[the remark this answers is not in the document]
bbednarski9: This comment is incorrect. I have tested this on the ETCD server and it doesnt work this way. using current payload {"key":"AA==","range_end":"AA==","keys_only":true} returns all keys correctly

external/dynamo/start_dynamo_unified_thompson_hints.sh
```diff
@@ -272,41 +273,64 @@ docker run -d \
...
-        
-        echo \"WARNING: \$worker_type worker initialization timeout, proceeding anyway\"
-        return 0
+
+        echo \"ERROR: \$worker_type worker failed to register with ETCD within \${max_wait}s\"
+        echo \"  Image: $IMAGE\"
+        echo \"  The model may require more time to load, or there may be a startup error.\"
+        echo \"  Check worker logs for details.\"
+        return 1
     }
```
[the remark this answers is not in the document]
bbednarski9: This comment is incorrect. I have tested this on the ETCD server and it doesnt work this way. using current payload {"key":"AA==","range_end":"AA==","keys_only":true} returns all keys correctly

external/dynamo/start_dynamo_disagg.sh
```diff
@@ -311,15 +378,52 @@ docker run -d \
...
+            sleep \$poll_interval
+            elapsed=\$((elapsed + poll_interval))
+        done
+
+        echo \"ERROR: Timeout waiting for workers to register\"
+        return 1
+    }
+
+    wait_for_workers_parallel || exit 1
+
```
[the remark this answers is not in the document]
bbednarski9: This comment is incorrect. I have tested this on the ETCD server and it doesnt work this way. using current payload {"key":"AA==","range_end":"AA==","keys_only":true} returns all keys correctly

external/dynamo/start_dynamo_unified.sh
```diff
@@ -242,41 +242,64 @@ docker run -d \
...
             fi
-            
-            # After 60s, assume it's initialized (model loading takes time for 70B)
-            if [ \$elapsed -ge 60 ]; then
-                echo \"✓ \$worker_type worker should be initialized\"
+
+            # Check if we got any keys back (count > 0 means workers registered)
+            if echo \"\$etcd_response\" | grep -q '\"count\"' && \
+               ! echo \"\$etcd_response\" | grep -q '\"count\":\"0\"'; then
+                echo \"✓ \$worker_type worker is ready (registered with ETCD at \${elapsed}s)\"
```
mnajafian-nv:
```suggestion
# Check ETCD for Dynamo backend workers
# Workers register under the /dynamo/backend/* namespace in ETCD
# We need to query for keys with prefix /dynamo/backend/ to find registered workers
# ETCD v3 uses base64-encoded keys for prefix matching:
#   key="/dynamo/backend/" -> base64="L2R5bmFtby9iYWNrZW5kLw=="
#   range_end="/dynamo/backend0" -> base64="L2R5bmFtby9iYWNrZW5kMA==" (lexicographically next char)
local etcd_response=\$(curl -s --max-time 2 http://localhost:$ETCD_CLIENT_PORT/v3/kv/range \
    -X POST \
    -H \"Content-Type: application/json\" \
    -d '{\"key\":\"L2R5bmFtby9iYWNrZW5kLw==\",\"range_end\":\"L2R5bmFtby9iYWNrZW5kMA==\",\"keys_only\":true}' 2>&1)

# Extract the count of registered workers from ETCD response
local worker_count=\$(echo \"\$etcd_response\" | grep -o '\"count\":\"[0-9]*\"' | grep -o '[0-9]*' || echo \"0\")

# Debug: Show progress every 30s to help diagnose slow model loads
if [ \$((elapsed % 30)) -eq 0 ] && [ \$elapsed -gt 0 ]; then
    echo \"  [DEBUG] Workers registered in ETCD: \${worker_count}, elapsed: \${elapsed}s\"
    if [ \"\$worker_count\" -eq 0 ]; then
        echo \"  [DEBUG] Still waiting for worker to register (model may still be loading)\"
    fi
fi

# Success condition: At least one Dynamo backend worker is registered in ETCD
# Note: Registration happens early, but model loading may still be in progress
if [ \"\$worker_count\" -gt 0 ] 2>/dev/null; then
    echo \"✓ \$worker_type worker is ready (registered with ETCD at \${elapsed}s)\"
    return 0
fi```

File: external/dynamo/start_dynamo_unified.sh [hunk 17]
```diff
@@ -242,41 +242,64 @@ docker run -d \
...
-            if [ \$((elapsed % 15)) -eq 0 ]; then
-                echo \"  ... \${elapsed}s / \${max_wait}s\"
+
+            # Check ETCD for registered workers using v3 API
+            # Query ALL keys to find where Dynamo registers (empty key "" with range_end "\0" = all keys)
+            # Base64: "" -> AA==, "\0" -> AA==  (we use keys_only to reduce response size)
+            local etcd_response=\$(curl -s --max-time 2 http://localhost:$ETCD_CLIENT_PORT/v3/kv/range \
+                -X POST \
+                -H \"Content-Type: application/json\" \
+                -d '{\"key\":\"AA==\",\"range_end\":\"AA==\",\"keys_only\":true}' 2>&1)
```
mnajafian-nv, line 302:
```suggestion
            # Check ETCD for registered workers using v3 API
            # Query all keys from 0x00 onwards to detect any worker registrations
            # Base64: "" -> AA==, empty range_end means "all keys from key to end"
            # When key==range_end, ETCD does single-key lookup (not a range scan)
            local etcd_response=\$(curl -s --max-time 2 http://localhost:$ETCD_CLIENT_PORT/v3/kv/range \
                -X POST \
                -H \"Content-Type: application/json\" \
                -d '{\"key\":\"AA==\",\"range_end\":\"\",\"keys_only\":true}' 2>&1)
```

mnajafian-nv: When key == range_end, ETCD does a single-key lookup, not a range scan. 
Currently code queries for one non-existent key and always gets count=0. Workers could be registered but we'd never detect them.

external/dynamo/start_dynamo_unified.sh
```diff
@@ -242,41 +242,64 @@ docker run -d \
...
-    
-    # Function to wait for worker initialization
+
+    # Function to wait for worker initialization by checking ETCD registration
+    # Dynamo workers register with ETCD, they don't expose HTTP health endpoints
     wait_for_worker() {
         local worker_type=\$1
         local pid=\$2
-        local max_wait=120
+        local max_wait=300
```
mnajafian-nv:
```suggestion
        # 70B models take 5-10 min to load - 600s gives headroom for slower storage
        local max_wait=${DYNAMO_WORKER_INIT_TIMEOUT_S:-600}
```

mnajafian-nv: 70B models take 5-10 minutes to load (per README and NVBug reports). A 5-minute timeout sits right at the lower bound and causes false failures on slower systems.

external/dynamo/generalized/router.py
```diff
@@ -271,17 +302,64 @@ def _iat_factor(iat: str) -> float:
...
-        logger.info("Getting engine client")
+        logger.info("Getting engine client for dynamo/backend/generate")
         self.engine_client = await engine.endpoint("generate").client()
-        await self.engine_client.wait_for_instances()
+
+        min_workers = int(self.min_workers)
+        if min_workers < 0:
+            raise ValueError(f"min_workers must be >= 0, got {min_workers}")
+
+        timeout_s = float(os.environ.get("DYNAMO_ROUTER_WAIT_FOR_WORKERS_TIMEOUT_S", "300"))
```
mnajafian-nv:
```suggestion
        # Must match bash timeout to avoid race: registration != ready to serve
        timeout_s = float(os.environ.get("DYNAMO_ROUTER_WAIT_FOR_WORKERS_TIMEOUT_S", "600"))
```

mnajafian-nv: Race condition: bash script succeeds at 290s when worker registers, but router times out at 300s while worker is still loading the model. Router needs to wait at least as long as bash does.

external/dynamo/start_dynamo_unified.sh
bbednarski9: This works as is. The current implementation gets all keys, versus the exact key "AA==" ~ "\x00"

I have tested within the container and validated that the current implementation finds the key registrations after they are made.

bbednarski9 (commented):
passing CI

external/dynamo/start_dynamo_unified.sh
mnajafian-nv, line 89:
```suggestion
if [ -z "${DYNAMO_MODEL_DIR}" ]; then
    echo "ERROR: DYNAMO_MODEL_DIR environment variable must be set"
    echo ""
    echo "Example:"
    echo "  export DYNAMO_MODEL_DIR=\"/path/to/your/models/Llama-3.3-70B-Instruct\""
    echo ""
    echo "Then run this script again."
    exit 1
fi

# If directory exists, validate it's a proper model directory (NVBug 5756833)
# If it doesn't exist, the download workflow later will handle it
if [ -d "${DYNAMO_MODEL_DIR}" ]; then
    if [ ! -f "${DYNAMO_MODEL_DIR}/config.json" ]; then
        echo "ERROR: ${DYNAMO_MODEL_DIR} exists but is not a valid model directory"
        echo ""
        echo "Missing: config.json"
        echo ""
        echo "Common mistake - pointing to cache root instead of model snapshot:"
        echo "   Wrong: ~/.cache/huggingface/"
        echo "   Right:  ~/.cache/huggingface/hub/models--meta-llama--Llama-3.3-70B-Instruct/snapshots/<hash>"
        echo ""
        echo "Find it: find ~/.cache/huggingface/hub -name config.json -path '*Llama-3.3-70B*'"
        exit 1
    fi
    
    # Verify config.json has model_type field (exact error from NVBug 5756833)
    if ! grep -q '"model_type"' "${DYNAMO_MODEL_DIR}/config.json" 2>/dev/null; then
        echo "ERROR: ${DYNAMO_MODEL_DIR}/config.json is missing 'model_type' field"
        echo ""
        echo "This usually means incomplete/corrupted download. Try:"
        echo "  rm -rf ${DYNAMO_MODEL_DIR}"
        echo "  huggingface-cli download meta-llama/Llama-3.3-70B-Instruct --local-dir ${DYNAMO_MODEL_DIR}"
        exit 1
    fi
fi

LOCAL_MODEL_DIR="${DYNAMO_MODEL_DIR}"
```

mnajafian-nv: Validate model directory early to avoid 5-minute wait

Problem: User set DYNAMO_MODEL_DIR=/home/user/.cache/huggingface/ (the cache root). This directory exists, so the script continues. 5 minutes later, the worker tries to load the model and crashes because there's no config.json at that path.

Fix: When the directory exists, check that it has config.json with a model_type field. Fail immediately with a helpful message instead of after the model loading timeout.
If the directory doesn't exist, skip this check - the existing download prompt (line 189) will handle it.

external/dynamo/start_dynamo_unified_thompson_hints.sh
```diff
@@ -272,41 +273,64 @@ docker run -d \
...
-            if [ \$((elapsed % 15)) -eq 0 ]; then
-                echo \"  ... \${elapsed}s / \${max_wait}s\"
+
+            # Check ETCD for registered workers using v3 API
+            # Query ALL keys to find where Dynamo registers (empty key "" with range_end "\0" = all keys)
+            # Base64: "" -> AA==, "\0" -> AA==  (we use keys_only to reduce response size)
+            local etcd_response=\$(curl -s --max-time 2 http://localhost:2379/v3/kv/range \
+                -X POST \
+                -H \"Content-Type: application/json\" \
+                -d '{\"key\":\"AA==\",\"range_end\":\"AA==\",\"keys_only\":true}' 2>&1)
```
mnajafian-nv, line 333:
```suggestion
                -d '{\"key\":\"AA==\",\"range_end\":\"\",\"keys_only\":true}' 2>&1)
```

mnajafian-nv: Same ETCD query bug here
Same issue - key == range_end does single-key lookup, not range scan.

external/dynamo/start_dynamo_disagg.sh
```diff
@@ -311,15 +378,52 @@ docker run -d \
...
+            if ! kill -0 \$DECODE_PID 2>/dev/null; then
+                echo \"ERROR: Decode worker process died!\"
+                return 1
+            fi
+
+            # Check ETCD for registered workers
+            local etcd_response=\$(curl -s --max-time 2 \
+                -X POST http://localhost:2379/v3/kv/range \
+                -H \"Content-Type: application/json\" \
+                -d '{\"key\":\"AA==\",\"range_end\":\"AA==\",\"keys_only\":true}' 2>&1)
```
mnajafian-nv, line 435:
```suggestion
                -d '{\"key\":\"AA==\",\"range_end\":\"\",\"keys_only\":true}' 2>&1)
```

mnajafian-nv: Same key == range_end issue. This function waits for both prefill + decode workers, so a broken query means we never detect when both are ready.

external/dynamo/start_dynamo_unified_thompson_hints.sh
```diff
@@ -272,41 +273,64 @@ docker run -d \
...
-    
-    # Function to wait for worker initialization
+
+    # Function to wait for worker initialization by checking ETCD registration
+    # Dynamo workers register with ETCD, they don't expose HTTP health endpoints
     wait_for_worker() {
         local worker_type=\$1
         local pid=\$2
-        local max_wait=120
+        local max_wait=300
```
mnajafian-nv:
```suggestion
        # 70B models take 5-10 min to load - use 10 min to avoid false failures
        local max_wait=${DYNAMO_WORKER_INIT_TIMEOUT_S:-600}
```

mnajafian-nv: Same 300s problem. This script waits for the worker before starting router/processor/frontend (line 352), so if the worker times out, nothing else starts.

external/dynamo/start_dynamo_disagg.sh
```diff
@@ -311,15 +378,52 @@ docker run -d \
...
-    # Give workers extra time to register with ETCD
-    echo ''
-    echo 'Waiting for workers to register with ETCD (30s)...'
-    sleep 30
-    
+
+    # Wait for BOTH workers to register (expects 2 workers in ETCD)
+    echo \"Waiting for both workers to initialize in parallel...\"
+    wait_for_workers_parallel() {
+        local max_wait=300
```
mnajafian-nv:
```suggestion
        # 2 workers loading 70B models in parallel - need extra time for contention
        local max_wait=${DYNAMO_WORKER_INIT_TIMEOUT_S:-600}
```

mnajafian-nv: Timeout too short for 2 workers loading in parallel: Waiting for prefill + decode workers, both loading 70B models simultaneously. 300s is too tight when both are competing for disk I/O and CUDA initialization

mnajafian-nv (changes requested):
Thanks for tackling the startup reliability issues! The core fixes look good, removing the 60s early-exit and adding proper synchronization will help.

external/dynamo/start_dynamo_unified.sh
bbednarski9: Added this envar to the .env.example template and to all .sh build scripts

external/dynamo/generalized/router.py
bbednarski9: Good catch, thank you

external/dynamo/start_dynamo_unified_thompson_hints.sh
bbednarski9: addressed

external/dynamo/start_dynamo_disagg.sh
bbednarski9: addressed

external/dynamo/start_dynamo_disagg.sh
bbednarski9: The intention is to query all keys as it's a fresh server. I decoded keys in my testing and verified that it works as intended:

The code itself is correct though. For etcd v3's /v3/kv/range API:
key: "AA==" (null byte \x00) = start from the lowest possible key
range_end: "AA==" (null byte \x00) = special meaning "no upper bound" / query all keys
So {"key":"AA==","range_end":"AA==","keys_only":true} is the correct way to query ALL keys in etcd v3.

external/dynamo/start_dynamo_unified_thompson_hints.sh
bbednarski9: The intention is to query all keys as it's a fresh server. I decoded keys in my testing and verified that it works as intended:

The code itself is correct though. For etcd v3's /v3/kv/range API:
key: "AA==" (null byte \x00) = start from the lowest possible key
range_end: "AA==" (null byte \x00) = special meaning "no upper bound" / query all keys
So {"key":"AA==","range_end":"AA==","keys_only":true} is the correct way to query ALL keys in etcd v3.

external/dynamo/start_dynamo_unified.sh
bbednarski9: addressed in all three and retested

mnajafian-nv (approved):

PR #1350, opened 2026-01-07
Improves finetuning end status logging

dnandakumar-nv:
## Description
Improves logging for finetuning running

## By Submitting this PR I confirm:
- I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/docs/source/resources/contributing/index.md).
- We require that all contributors "sign-off" on their commits. This certifies that the contribution is your original work, or you have rights to submit it under the same license, or a compatible license.
  - Any contribution which contains commits that are not Signed-Off will not be accepted.
- When the PR is ready for review, new or existing tests cover these changes.
- When the PR is ready for review, the documentation is up to date with these changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate fine-tuning job reporting: warns when no jobs ran; reports errors for failed jobs; warns on cancellations; confirms success only when all jobs complete; summarizes mixed states with counts.
  * Ensures resources are cleaned up even after errors; exceptions are logged and then re-raised.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1364, opened 2026-01-08
Fix training cancellation 404 error

mnajafian-nv:
## Description

Closes: nvbugs-5789819

### Fix: Fail-fast on training cancellation and validate LLM endpoints before eval

Adds fail-fast error handling when training jobs are canceled/failed and validates LLM endpoints before starting evaluation to prevent cryptic 404 errors.

When a finetuning job gets canceled (e.g., timeout at 93.3% progress), the current behavior:
- Logs cancellation as status update but doesn't raise error
- Attempts to deploy a non-existent model
- Evaluation starts and fails with "404 page not found" mid-run

This wastes user time debugging unclear errors.

1. Fail-fast on training failure: Raise `RuntimeError` immediately when training status is `CANCELED` or `FAILED` with actionable error messages
2. Pre-validation: Check LLM endpoints are reachable before starting eval workflow
3. Better error messages: Include troubleshooting steps in error output

### Changes
- `trainer_adapter.py` - Added explicit error handling for canceled/failed training jobs
- `evaluate.py` - Integrated LLM validation before workflow execution  
- `llm_validator.py` (new) - Validates OpenAI-compatible endpoints and catches 404s early
- `test_llm_validator.py` (new) - 11 tests covering validation scenarios

### Testing
-  All new tests pass (11/11)
-  No linter errors
-  Tested scenario: canceled training → immediate error with guidance

### Notes
- Related: PR #1350 (merged) improved status logging; this adds fail-fast behavior
- Target: `release/1.4`
- No breaking changes

## Summary by CodeRabbit

* **New Features**
  * Opt-in LLM endpoint validation before evaluations via a new config flag (defaults off). Batched async checks with actionable messages; fatal on unreachable endpoints.

* **Bug Fixes**
  * Improved training terminal-state handling: explicit errors for FAILED/CANCELED when deployment expected, clearer deployment logs, and safer non-deploy paths.

* **Tests**
  * Comprehensive tests for endpoint validation across providers, 404 vs other errors, timeouts, parallel execution, and messaging.

### DIFF
--- a/packages/nvidia_nat_nemo_customizer/src/nat/plugins/customizer/dpo/trainer_adapter.py
+++ b/packages/nvidia_nat_nemo_customizer/src/nat/plugins/customizer/dpo/trainer_adapter.py
@@ -407,38 +407,32 @@ class NeMoCustomizerTrainerAdapter(TrainerAdapter):
                 if status.status == TrainingStatusEnum.FAILED:
                     error_msg = status.message or "Training job failed"
                     logger.error(f"Training job {ref.run_id} FAILED: {error_msg}")
-                    raise RuntimeError(
-                        f"Training job {ref.run_id} failed: {error_msg}. "
-                        f"Model was not trained and cannot be deployed. "
-                        f"Please check the job logs for details and retry the training."
-                    )
-                
+                    raise RuntimeError(f"Training job {ref.run_id} failed: {error_msg}. "
+                                       f"Model was not trained and cannot be deployed. "
+                                       f"Please check the job logs for details and retry the training.")
+
                 if status.status == TrainingStatusEnum.CANCELED:
                     error_msg = status.message or "Training job was canceled"
                     logger.error(f"Training job {ref.run_id} CANCELED: {error_msg}")
-                    
+
                     # Format progress safely
                     progress_str = f"{status.progress:.1f}%" if status.progress is not None else "unknown progress"
-                    
+
                     # If deployment was expected, raise an error
                     if self.adapter_config.deploy_on_completion:
-                        raise RuntimeError(
-                            f"Training job {ref.run_id} was canceled at {progress_str}: {error_msg}. "
-                            f"Model was not trained and will NOT be deployed. "
-                            f"Evaluation will fail because the model does not exist. "
-                            f"\n\nACTION REQUIRED:"
-                            f"\n1. Check if the job was manually canceled or timed out"
-                            f"\n2. Review NeMo MS platform health and resource availability"
-                            f"\n3. Consider increasing deployment_timeout_seconds in config"
-                            f"\n4. Use a fresh namespace to avoid conflicts: namespace: nat-dpo-test-v2"
-                            f"\n5. Retry training: nat finetune --config_file=..."
-                        )
+                        raise RuntimeError(f"Training job {ref.run_id} was canceled at {progress_str}: {error_msg}. "
+                                           f"Model was not trained and will NOT be deployed. "
+                                           f"Evaluation will fail because the model does not exist. "
+                                           f"\n\nACTION REQUIRED:"
+                                           f"\n1. Check if the job was manually canceled or timed out"
+                                           f"\n2. Review NeMo MS platform health and resource availability"
+                                           f"\n3. Consider increasing deployment_timeout_seconds in config"
+                                           f"\n4. Use a fresh namespace to avoid conflicts: namespace: nat-dpo-test-v2"
+                                           f"\n5. Retry training: nat finetune --config_file=...")
                     else:
                         # Just log warning if deployment wasn't expected
-                        logger.warning(
-                            f"Training job {ref.run_id} was canceled at {progress_str}. "
-                            f"No deployment was configured (deploy_on_completion=False)."
-                        )
+                        logger.warning(f"Training job {ref.run_id} was canceled at {progress_str}. "
+                                       f"No deployment was configured (deploy_on_completion=False).")
                         return status
 
                 # Handle successful completion with deployment
@@ -448,10 +442,8 @@ class NeMoCustomizerTrainerAdapter(TrainerAdapter):
                         await self._deploy_model(ref)
                         logger.info(f"Model deployed successfully for job {ref.run_id}")
                     else:
-                        logger.info(
-                            f"Training job {ref.run_id} completed successfully. "
-                            f"Skipping deployment (deploy_on_completion=False)."
-                        )
+                        logger.info(f"Training job {ref.run_id} completed successfully. "
+                                    f"Skipping deployment (deploy_on_completion=False).")
 
                 return status
 
--- a/packages/nvidia_nat_nemo_customizer/src/nat/plugins/customizer/dpo/trainer_adapter.py
+++ b/packages/nvidia_nat_nemo_customizer/src/nat/plugins/customizer/dpo/trainer_adapter.py
@@ -400,12 +400,55 @@ class NeMoCustomizerTrainerAdapter(TrainerAdapter):
                     TrainingStatusEnum.FAILED,
                     TrainingStatusEnum.CANCELED,
             ):
-                # Handle deployment if configured
-                if (status.status == TrainingStatusEnum.COMPLETED and self.adapter_config.deploy_on_completion):
-                    await self._deploy_model(ref)
-
-                # Clean up active job tracking
+                # Clean up active job tracking first
                 self._active_jobs.pop(ref.run_id, None)
+
+                # Handle non-successful completions with clear error messages
+                if status.status == TrainingStatusEnum.FAILED:
+                    error_msg = status.message or "Training job failed"
+                    logger.error(f"Training job {ref.run_id} FAILED: {error_msg}")
+                    raise RuntimeError(
+                        f"Training job {ref.run_id} failed: {error_msg}. "
+                        f"Model was not trained and cannot be deployed. "
+                        f"Please check the job logs for details and retry the training."
+                    )
+                
+                if status.status == TrainingStatusEnum.CANCELED:
+                    error_msg = status.message or "Training job was canceled"
+                    logger.error(f"Training job {ref.run_id} CANCELED: {error_msg}")
+                    
+                    # If deployment was expected, raise an error
+                    if self.adapter_config.deploy_on_completion:
+                        raise RuntimeError(
+                            f"Training job {ref.run_id} was canceled at {status.progress:.1f}% progress: {error_msg}. "
+                            f"Model was not trained and will NOT be deployed. "
+                            f"Evaluation will fail because the model does not exist. "
+                            f"\n\nACTION REQUIRED:"
+                            f"\n1. Check if the job was manually canceled or timed out"
+                            f"\n2. Review NeMo MS platform health and resource availability"
+                            f"\n3. Consider increasing deployment_timeout_seconds in config"
+                            f"\n4. Use a fresh namespace to avoid conflicts: namespace: nat-dpo-test-v2"
+                            f"\n5. Retry training: nat finetune --config_file=..."
+                        )
+                    else:
+                        # Just log warning if deployment wasn't expected
+                        logger.warning(
+                            f"Training job {ref.run_id} was canceled at {status.progress:.1f}% progress. "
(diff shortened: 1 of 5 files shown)


src/nat/eval/llm_validator.py
```diff
@@ -0,0 +1,138 @@
...
+
+    for llm_name, llm_config in config.llms.items():
+        try:
+            # Only validate OpenAI-compatible endpoints
+            if llm_config.type not in ["openai", "nim"]:
+                logger.debug("Skipping validation for LLM '%s' (type: %s)", llm_name, llm_config.type)
+                continue
+
+            base_url = getattr(llm_config, "base_url", None)
+            model_name = getattr(llm_config, "model_name", None)
```
willkill07: I think this is too restrictive and tailored towards OpenAI. It also doesn't handle if a specific API key is included as part of the configuration.

I'm not sure what the right approach would be (perhaps checking if any framework for each model in the configuration can issue a request via ainvoke)

dnandakumar-nv: Perhaps another fix instead of openai-specific validation would be to set a flag in `eval` like `fail_on_first` or something that would immediately raise and stop the evaluation process on its first 404, or other configurable error code. 

That would also cover unavailability of other endpoint types like retrievers, memory, embedders, etc. while still gaining efficiency by failing fast

dnandakumar-nv:
I do like the increased validation on the `trainer_adapter`

mnajafian-nv:
*Proposed Path Forward*:  I see three options:

*Option 1:* Keep current PR + Create follow-up issue
• PRO: Addresses the immediate bug with minimal risk
• PRO: Unblocks users hitting this specific 404 issue
• CON: Leaves the broader solution for later

*Option 2:* Expand current PR to include fail_fast flag
• PRO: More comprehensive solution
• CON: Larger scope, more testing needed
• CON: Delays fixing the immediate bug

*Option 3:* Framework-agnostic validation via ainvoke
• PRO: Works for all LLM types
• CON: More complex implementation
• CON: May not catch 404s before workflow starts (happens during execution)

*My Recommendation:* Ship Option 1 now, implement Option 2 as follow-up
*Rationale:* Current PR solves the immediate pain point (NIM 404 errors after training cancellation). The OpenAI-specific validation is safe (skips other types, handles missing keys). We can design the fail_fast flag properly in a follow-up without rushing.

What do you think? I'm happy to implement either approach based on your preference!

dnandakumar-nv:
If we chose to ship Option 1 now, I think we should make endpoint validation optional with a config so users can enable if wanted / or conversely, disable if they experience regression

willkill07:
Option 1 will cause `nat eval` to fail:

```python
            try:
                from nat.eval.llm_validator import validate_llm_endpoints
                logger.info("Validating LLM endpoints before evaluation...")
                await validate_llm_endpoints(config)
            except Exception as e:
                logger.error(f"LLM endpoint validation failed: {e}")
                raise
```

As written, if any of the validation doesn't work due to API key being unspecified, then it will fail. It will fail on AWS Bedrock models. It will fail on any configuration where such a key is provided as part of the config. That's why I was proposing the solution I did. It's the only way to not guarantee a failure by actually accessing what would actually happen in `nat eval`

My proposed option 3 is being misclassified:
• CON: May not catch 404s before workflow starts (happens during execution)

This is not true. I'm suggesting the logic for `validate_llm_endpoints` to create an artificial config from just the LLMs and test with the various frameworks.

mnajafian-nv:
You're right, let me implement Option 3 properly using the builder + ainvoke approach and get back to you

mnajafian-nv:
@willkill07 @dhruvgairola Thanks for the great feedback. Implemented Option 3 as suggested.

**Changes**

**Framework-agnostic validation using WorkflowBuilder**
- Replaced direct OpenAI client with `WorkflowBuilder + ainvoke()`
- Now validates all LLM types (OpenAI, NIM, Bedrock, vLLM, etc.) uniformly
- Respects NAT's native auth system
- More specific 404 detection (requires HTTP 404 or model-specific context)
- Added 30s timeout to prevent hanging
- Parallel validation with batching (5 LLMs/batch)
- Error message truncation (memory protection)
- System exception safety (KeyboardInterrupt/SystemExit properly re-raised)
- Config structure validation
- Added observability: Duration logging per LLM & Validation summary metrics (total/succeeded/warned/failed)

**Testing**
- 20 tests, all passing
- Covers OpenAI, NIM, Bedrock, timeouts, parallel execution, edge cases

**Impact**

- Catches 404s before eval starts (addresses NVBug 5789819)
- Works across all NAT-supported LLM providers
- Opt-in via `eval.general.validate_llm_endpoints: true`

Ready for your review :)

src/nat/eval/llm_validator.py
```diff
@@ -0,0 +1,324 @@
...
+        Tuple of (llm_name, error_type, error_message):
+        - error_type: "404" for model not found, "warning" for non-critical, None for success
+        - error_message: Description of the error, or None if successful
+    """
+    try:
+        logger.info("Validating LLM '%s' (type: %s)", llm_name, llm_config.type)
+        start_time = time.time()
+
+        # Add LLM to builder (handles all LLM types)
+        await asyncio.wait_for(builder.add_llm(llm_name, llm_config), timeout=VALIDATION_TIMEOUT_SECONDS)
```
willkill07: This will never timeout
```suggestion
        builder.add_llm(llm_name, llm_config)
```

src/nat/eval/llm_validator.py
```diff
@@ -0,0 +1,324 @@
...
+
+        # Get LangChain-wrapped LLM instance
+        llm = await asyncio.wait_for(
+            builder.get_llm(llm_name, LLMFrameworkEnum.LANGCHAIN), timeout=VALIDATION_TIMEOUT_SECONDS
+        )
+
+        # Test with minimal prompt - this will hit the endpoint
+        await asyncio.wait_for(llm.ainvoke(VALIDATION_PROMPT), timeout=VALIDATION_TIMEOUT_SECONDS)
```
willkill07: This is nice, but we should be looping over all valid LLMFrameworkEnum types and:

wrapping `get_llm` within try-except
* if exception occurs, try the next enumeration
* if exception doesn't occur, then continue the test

src/nat/eval/llm_validator.py
```diff
@@ -0,0 +1,324 @@
...
+
+    # Keep first and last portions to preserve both error description and stack trace
+    separator = " ... (truncated) ... "
+    keep_length = (max_length - len(separator)) // 2
+    return f"{message[:keep_length]}{separator}{message[-keep_length:]}"
+
+
+async def _validate_single_llm(
+    builder: WorkflowBuilder, llm_name: str, llm_config: LLMBaseConfig
+) -> tuple[str, str | None, str | None]:
```
willkill07: since we pass in the LLM name, I don't think we need to return it

src/nat/eval/llm_validator.py
```diff
@@ -0,0 +1,324 @@
...
+    try:
+        logger.info("Validating LLM '%s' (type: %s)", llm_name, llm_config.type)
+        start_time = time.time()
+
+        # Add LLM to builder (handles all LLM types)
+        await asyncio.wait_for(builder.add_llm(llm_name, llm_config), timeout=VALIDATION_TIMEOUT_SECONDS)
+
+        # Get LangChain-wrapped LLM instance
+        llm = await asyncio.wait_for(
+            builder.get_llm(llm_name, LLMFrameworkEnum.LANGCHAIN), timeout=VALIDATION_TIMEOUT_SECONDS
```
willkill07: I don't think this will timeout for any reason

src/nat/eval/llm_validator.py
```diff
@@ -0,0 +1,324 @@
...
+            base_url, model_name = _get_llm_endpoint_info(llm_config)
+
+            error_msg = (
+                f"LLM '{llm_name}' validation failed: Model not found (404).\n"
+                f"\nThis typically means:\n"
+                f"  1. The model has not been deployed yet\n"
+                f"  2. The model name is incorrect\n"
+                f"  3. A training job was canceled and the model was never deployed\n"
+                f"\nLLM Configuration:\n"
+                f"  Type: {llm_config.type}\n"
```
willkill07:
```suggestion
                f"  Type: {str(llm_config.type)}\n"
```

src/nat/eval/llm_validator.py
```diff
@@ -0,0 +1,324 @@
...
+    # Validate config structure
+    if not hasattr(config, "llms"):
+        raise ValueError("Config object does not have 'llms' attribute")
+
+    if not config.llms:
+        logger.debug("No LLMs defined in config, skipping validation")
+        return
+
+    if not isinstance(config.llms, dict):
+        raise ValueError(f"config.llms must be a dict, got {type(config.llms)}")
```
willkill07: The Pydantic model must be adhered to in order to load. No need to check this precondition
```suggestion
```

src/nat/eval/llm_validator.py
```diff
@@ -0,0 +1,324 @@
...
+    validation_warnings = []  # List of (llm_name, warning_message) tuples for non-critical errors
+
+    # Use WorkflowBuilder to instantiate and test LLMs
+    async with WorkflowBuilder() as builder:
+        # Get list of LLMs to validate
+        llm_items = list(config.llms.items())
+
+        # Validate in batches to respect rate limits
+        for batch_start in range(0, len(llm_items), CONCURRENT_VALIDATION_BATCH_SIZE):
+            batch = llm_items[batch_start : batch_start + CONCURRENT_VALIDATION_BATCH_SIZE]
```
willkill07: note to self: once we drop Python 3.11, we can use `itertools.batched`

src/nat/eval/llm_validator.py
```diff
@@ -0,0 +1,321 @@
...
+            try:
+                llm = await builder.get_llm(llm_name, framework)
+                logger.debug("LLM '%s' successfully loaded with framework '%s'", llm_name, framework.value)
+                break  # Found a working framework
+            except Exception as e:
+                logger.debug("LLM '%s' failed with framework '%s': %s", llm_name, framework.value, e)
+                continue  # Try next framework
+
+        if llm is None:
+            return ("warning", "Could not instantiate LLM with any framework")
```
willkill07: If we can't instantiate, then shouldn't validation implicitly fail?

willkill07 (approved):
Subjective nit on failing if no initialization found. Though I suppose with user-defined frameworks, a warning is good enough if we are unable to instantiate.

mnajafian-nv:
All feedback addressed in latest commit.

**Fixed:**
- Config validation (lines 260, 330)
- Error message context (line 225)
- Logger.exception usage (line 215)
- Edge case guard (line 128)
- NAT terminology (line 29)
- Linting: unused var, strict=True (line 272)

**Deferred:**
- Race condition (line 226): Low risk in practice (unique llm_names per task). Adding locks increases complexity. Can revisit if usage patterns change.

Ready for re-review.

packages/nvidia_nat_nemo_customizer/src/nat/plugins/customizer/dpo/trainer_adapter.py
```diff
@@ -400,13 +400,51 @@ async def wait_until_complete(self, ref: TrainingJobRef, poll_interval: float |
...
+                    if self.adapter_config.deploy_on_completion:
+                        raise RuntimeError(f"Training job {ref.run_id} was canceled at {progress_str}: {error_msg}. "
+                                           f"Model was not trained and will NOT be deployed. "
+                                           f"Evaluation will fail because the model does not exist. "
+                                           f"\n\nACTION REQUIRED:"
+                                           f"\n1. Check if the job was manually canceled or timed out"
+                                           f"\n2. Review NeMo MS platform health and resource availability"
+                                           f"\n3. Consider increasing deployment_timeout_seconds in config"
+                                           f"\n4. Use a fresh namespace to avoid conflicts: namespace: nat-dpo-test-v2"
+                                           f"\n5. Retry training: nat finetune --config_file=...")
```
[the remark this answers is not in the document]
mnajafian-nv: Ack. Can use generic placeholder if needed, but the context makes it clear this is an example.

willkill07 (approved):

PR #1375, opened 2026-01-09
chore: bump urllib3+langchain; specify werkzeug as transitive dep

willkill07:
## Description

Closes

## Summary by CodeRabbit

* **Chores**
  * Bumped werkzeug from 3.1.4 to 3.1.5 to keep compatibility and security patches up to date.
  * Added/updated urllib3 entry to ensure networking stability.
  * Upgraded langchain from 1.2.2 to 1.2.3 for improved library fixes and reliability.

### DIFF
--- a/packages/nvidia_nat_semantic_kernel/pyproject.toml
+++ b/packages/nvidia_nat_semantic_kernel/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
   "ruamel-yaml-clibz==0.3.5", 
   "semantic-kernel~=1.36",
   # transitive dependencies
-  "werkzeug~=3.1.4",
+  "werkzeug~=3.1.5",
 ]
 requires-python = ">=3.11,<3.14"
 description = "Subpackage for Semantic-Kernel integration in NeMo Agent toolkit"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,8 @@ dependencies = [
   "tabulate~=0.9",
   "uvicorn[standard]~=0.38",
   "wikipedia~=1.4",
+  # transitive dependencies
+  "urllib3~=2.6.3",
 ]
 requires-python = ">=3.11,<3.14"
 description = "NVIDIA NeMo Agent toolkit"
@@ -332,6 +334,8 @@ dev = [
   "twine~=6.0",
   "vale~=3.12",
   "yapf==0.43.*",
+  # transitive dependencies
+  "werkzeug~=3.1.5",
 ]
 
 [project.entry-points.'nat.components']
--- a/packages/nvidia_nat_langchain/pyproject.toml
+++ b/packages/nvidia_nat_langchain/pyproject.toml
@@ -21,7 +21,7 @@ dependencies = [
   # version when adding a new package. If unsure, default to using `~=` instead of `==`. Does not apply to nvidia-nat packages.
   # Keep sorted!!!
   "nvidia-nat~=1.4",
-  "langchain~=1.2.2",
+  "langchain~=1.2.3",
   "langchain-aws~=1.1.0",
   "langchain-classic~=1.0.1",
   "langchain-core~=1.2.6",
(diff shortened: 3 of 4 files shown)


Salonijain27 (approved):

packages/nvidia_nat_semantic_kernel/pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
   "ruamel-yaml-clibz==0.3.5", 
   "semantic-kernel~=1.36",
   # transitive dependencies
-  "werkzeug~=3.1.4",
+  "werkzeug~=3.1.5",
```
mnajafian-nv, line 29: Q1. I think werkzeug is only used in test files (`nvidia_nat_opentelemetry/tests/`), not in semantic-kernel source. Why is this a runtime dependency? I searched the codebase, werkzeug is only imported in test files: `packages/nvidia_nat_opentelemetry/tests/observability/test_otlp_span_adapter_integration.py`
Shall we move this to test dependencies or removing if not needed?

Q2. Doesn't our repo standard require a 2-digit precision? "werkzeug~=3.1" instead of "werkzeug~=3.1.5"

pyproject.toml
mnajafian-nv, line 34: * Inconsistency: Version Precision Guidelines vs. Practice*

- Current Guideline says use 2 digits precision in the version specifier, e.g., use `~=1.2` instead of `~=1.2.3`". However the codebase has **inconsistent adherence** to this guideline

pyproject.toml
```diff
@@ -67,6 +67,8 @@ dependencies = [
   "tabulate~=0.9",
   "uvicorn[standard]~=0.38",
   "wikipedia~=1.4",
+  # transitive dependencies
+  "urllib3~=2.6.3",
```
mnajafian-nv, line 71: Q. Why are we Pinning Transitive Dependencies? (Not blocking, but please provide context on the strategy here for alignment)

I noticed you're adding `urllib3` and `werkzeug` as explicit dependencies, but:

**Dependency tree shows:**

` urllib3 ← requests ← (used by aioboto3, datasets, opensearch-py)
 werkzeug ← flask ← (used by langchain/galileo tooling) `

And neither package is directly imported in your source code. My concern is that explicitly pinning transitive deps creates maintenance burden. Every security patch requires manual PR instead of automatic resolution.
1. What problem does explicit pinning solve here?
2. Why not update the primary dependencies (requests, flask) instead?
3. Is this from a security scanner? 

Here are alternative approaches:
- Update primary dependencies that pull these in
- Use `[tool.uv.override-dependencies]` to set minimum versions
- Document specific compatibility issue if one exists

mnajafian-nv (changes requested):
Hey! Thanks for the PR. Everything looks great just need a few small changes:
1. **Version format:** Change `~=2.6.3` to `~=2.6` and `~=3.1.5` to `~=3.1` (per your guideline lines)
2. **Alphabetical order:** Move urllib3 before uvicorn, werkzeug before yapf  
3. **Regenerate:** `uv pip install . --sync`

Q. Why pin urllib3/werkzeug explicitly when they're transitive deps? Does semantic-kernel actually need werkzeug?

pyproject.toml
willkill07: 1. We eventually will have to do this, so I'm doing it now to ensure that we actually address security vulnerabilities.
2. I arrived at the locations I did for `# transitive dependencies` via `uv pip tree --invert --package <name>`. If a root-level package under "dev" group was found to be in this, then I put it there, since it's the only place where it _can_ go. If a root-level package was found under an `nvidia-nat-*` library, I added it under a `# transitive dependency` section there.
3. `override-dependencies` does NOT propagate to distributed wheels and will not work

packages/nvidia_nat_semantic_kernel/pyproject.toml
willkill07: Security vulnerability resolution must specify patch level for solving

pyproject.toml
willkill07: Security vulnerability resolution must specify patch level for solving

pyproject.toml
mnajafian-nv: Great context, makes sense!

mnajafian-nv (approved):

PR #1382, opened 2026-01-11
fix: langchain<>huggingface integration

willkill07:
## Description

The nightly CI ran with this error:
```
tests.nat.llm_providers.test_langchain_agents::test_huggingface_langchain_agent
ValueError: The model has been loaded with `accelerate` and therefore cannot be moved to a specific device. Please discard the `device` argument when creating your pipeline object.
```

This PR resolves that error in addition to the following changes:
- Renaming HuggingFace config field to `dtype` (from deprecated `torch_dtype`) and uses it when loading models.
- Updating LangChain HuggingFace client to avoid passing `device` when accelerate sharding is present, passes `dtype` to pipelines, and wraps `ChatHuggingFace` with an async `_agenerate` using `asyncio.to_thread`.
- Fixing test to now use TinyLlama (newer, small model) for the HuggingFace agent test, and assert response contains “3” instead of just non-empty content.

Closes

## Release Notes

* **Breaking Changes**
  * Updated HuggingFace LLM configuration parameter name; users may need to update their configuration.

* **Improvements**
  * Enhanced HuggingFace model compatibility with distributed architectures to prevent errors.
  * Added async execution support for HuggingFace LLMs.

* **Documentation**
  * Updated HuggingFace LLM configuration documentation.

### DIFF
--- a/docs/source/build-workflows/llms/index.md
+++ b/docs/source/build-workflows/llms/index.md
@@ -162,7 +162,7 @@ The HuggingFace LLM provider is defined by the {py:class}`~nat.llm.huggingface_l
 
 * `model_name` - The HuggingFace model name or path (for example, `Qwen/Qwen3Guard-Gen-0.6B`)
 * `device` - Device for model execution: `cpu`, `cuda`, `cuda:0`, or `auto` (default: `auto`)
-* `torch_dtype` - Torch data type: `float16`, `bfloat16`, `float32`, or `auto` (default: `auto`)
+* `dtype` - Torch data type: `float16`, `bfloat16`, `float32`, or `auto` (default: `auto`)
 * `max_new_tokens` - Maximum number of new tokens to generate (default: `128`)
 * `temperature` - Sampling temperature (default: `0.0`)
 * `trust_remote_code` - Whether to trust remote code when loading the model (default: `false`)
--- a/packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
+++ b/packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
@@ -16,6 +16,7 @@
 
 import logging
 from collections.abc import Sequence
+from typing import Any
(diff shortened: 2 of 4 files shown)


mnajafian-nv (approved):
LGTM overall.

A couple clarifications / follow-ups:

1) Transformers 4.57.x: pipeline(dtype=...) looks correct, but for AutoModelForCausalLM.from_pretrained(...) can you confirm dtype is a supported kwarg in our pinned version (4.57.3) / mapped internally, vs needing torch_dtype? If dtype is not honored, we should pass torch_dtype= when loading.

2) Backward compat: renaming config torch_dtype → dtype is breaking for existing YAML? Right? Can we accept both (deprecated alias) for at least one release, or document this as breaking + migration note?

3) Test: TinyLlama may be heavy/flaky for CI downloads. Shall we consider keeping an env override again or using a smaller test model while still asserting on “3”.

willkill07:
> Transformers 4.57.x: pipeline(dtype=...) looks correct, but for AutoModelForCausalLM.from_pretrained(...) can you confirm dtype is a supported kwarg in our pinned version (4.57.3) / mapped internally, vs needing torch_dtype? If dtype is not honored, we should pass torch_dtype= when loading.

Yes, I can confirm that `dtype` is honored.

> Backward compat: renaming config torch_dtype → dtype is breaking for existing YAML? Right? Can we accept both (deprecated alias) for at least one release, or document this as breaking + migration note?

There is no existing yaml using this, as it's a brand new feature which has yet to be released. There is no reason to keep backwards compatibility as its a new feature.

> Test: TinyLlama may be heavy/flaky for CI downloads. Shall we consider keeping an env override again or using a smaller test model while still asserting on “3”.

TinyLlama is the smallest model I could find which has a built-in prompt template defined. I do not believe there is a smaller (meaningful) model.

mnajafian-nv:
Thank you!

PR #1383, opened 2026-01-11
Fix/agno flaky test release 1.4

mnajafian-nv:
## Description
Fixes the flaky `test_get_event_loop_called` test that's failing on arm64 architectures.

The test was expecting `get_running_loop()` to be called exactly once, but on arm64 it gets called 3 times (vs 1 on amd64). This is due to pydantic-core's C extensions behaving differently across architectures during function introspection.

Changed the assertion to check for known values `[1, 3]` instead of `assert_called_once()`. Added inline comments explaining the architecture differences so future maintainers understand this is expected behavior.

**Fixes:** CI Pipeline / Test (arm64, 3.12) failure in #1349

## Summary by CodeRabbit

* **Tests**
  * Made test validations architecture- and Python-version-aware to reduce flaky failures across ARM64 and x86_64 hosts.
  * Replaced a rigid single-call assertion with configurable call-count checks per environment and a safer fallback for unknown combos.
  * Improved failure messages to clearly explain mismatches and guide updates when CI behavior changes.

### DIFF
--- a/packages/nvidia_nat_agno/tests/test_tool_wrapper.py
+++ b/packages/nvidia_nat_agno/tests/test_tool_wrapper.py
@@ -185,34 +185,49 @@ class TestToolWrapper:
         # Call the function under test
         agno_tool_wrapper("test_tool", mock_function, mock_builder)
 
-        # Verify that get_running_loop was called with architecture-specific expectations
-        # Known behavior: 1 call on x86_64/amd64, 3 calls on arm64/aarch64
-        # (due to pydantic-core C extension architecture differences in function introspection)
+        # Verify that get_running_loop was called
+        # The call count varies by architecture and Python version due to pydantic-core C extension
+        # differences in function introspection. This is an implementation detail we track to detect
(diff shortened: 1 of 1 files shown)


packages/nvidia_nat_agno/tests/test_tool_wrapper.py
```diff
@@ -185,7 +185,14 @@ def test_get_event_loop_called(self, mock_get_running_loop, mock_function, mock_
         agno_tool_wrapper("test_tool", mock_function, mock_builder)
 
         # Verify that get_running_loop was called
-        mock_get_running_loop.assert_called_once()
+        # Known behavior: 1 call on amd64, 3 calls on arm64
+        # (due to pydantic-core C extension architecture differences in function introspection)
+        call_count = mock_get_running_loop.call_count
+        assert call_count in [1, 3], (
```
willkill07: Could we be a little more strict here and rely on `platform.machine()` ?

willkill07 (approved):

PR #1409, opened 2026-01-14
Temporarily skip agent e2e tests

dagardner-nv:

## Summary by CodeRabbit

* **Tests**
  * Temporarily disabled long-running agent workflow tests to optimize test suite performance. Tests can be re-enabled after workflow updates.


mnajafian-nv (approved):
I think pytest-timeout is a better approach than skipping. Using @pytest.mark.timeout would maintain test coverage while preventing hangs, and give us clear failure signals instead of silent skips. Should we use timeout instead?

PR #1410, opened 2026-01-14
fix: correct ReWOO planner prompt JSON example format

mnajafian-nv:
The prompt example used malformed escaping (\'{{) that caused LLMs to output evidence as a string instead of an object, breaking Pydantic validation.

Changes:
- Remove \' escaped quotes from JSON example
- Change tool_input from tuples to arrays (valid JSON)
- Add comment explaining {{ template escaping
- Fix line length for ruff compliance

Fixes ReWOO agent failures with Nemotron and Llama 3.3 models.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed invalid JSON in AI agent prompt examples so sample outputs are valid, parseable JSON.
  * Normalized sample planning evidence to consistently use array-style inputs instead of mixed/escaped fragments.
  * Cleaned up formatting of example blocks to present clear, consistent planning-step JSON for end-user reference.

### DIFF
--- a/src/nat/agent/rewoo_agent/prompt.py
+++ b/src/nat/agent/rewoo_agent/prompt.py
@@ -54,6 +54,7 @@ Here is an example of how a valid JSON output should look:
 
 Note: {{ and }} are used for template escaping and will render as single braces { and } in the final prompt.
 
+```json
 [
   {{
     "plan": "Find Alex's schedule on Sep 25, 2025",
@@ -81,6 +82,7 @@ Alex's schedule: #E1; Bill's schedule: #E2?"
     }}
   }}
 ]
+```
 
 Begin!
 """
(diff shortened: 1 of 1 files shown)


src/nat/agent/rewoo_agent/prompt.py
```diff
@@ -52,32 +52,34 @@
 
 Here is an example of how a valid JSON output should look:
 
+Note: {{ and }} are used for template escaping and will render as single braces { and } in the final prompt.
```
willkill07: Because the template gets rendered before being passed to the LLM, I do not think this note is necessary.

```suggestion
```

src/nat/agent/rewoo_agent/prompt.py
```diff
@@ -52,32 +52,34 @@
...
+    "evidence": {{
       "placeholder": "#E3",
       "tool": "llm_chat",
       "tool_input": "Find a common 1-hour time slot for Alex and Bill given their schedules. \
 Alex's schedule: #E1; Bill's schedule: #E2?"
-    }}\'
-  }}\'
+    }}
+  }}
 ]
```
willkill07, line 83: Would it be more meaningful/helpful to surround the JSON output with:
` ```json ` and ` ``` `

willkill07 (approved):
Approving with comment

PR #1440, opened 2026-01-20
RAG Library Mode integration

ericevans-nv:
## Description

Adds the `nvidia_nat_rag` plugin package that integrates the [NVIDIA RAG Blueprint](https://github.com/NVIDIA-AI-Blueprints/rag) pipeline into NeMo Agent Toolkit, providing grounded retrieval and generation capabilities.

## What's Included

**New Package: `nvidia_nat_rag`**
- `search` tool: Retrieves grounded document excerpts from indexed sources
- `generate` tool: Synthesizes answers grounded in cited source material
- Declarative YAML configuration for RAG pipeline features:
  - Two-stage retrieval (vector search + semantic reranking)
  - Query rewriting via LLM
  - Confidence filtering
  - Structured citations
- Uses [`nvidia-rag>=2.4.0`](https://pypi.org/project/nvidia-rag/2.4.0/) from PyPI
- Adds undeclared `nvidia-rag` transitive dependencies (`langchain-openai`, `opentelemetry-api`, `opentelemetry-sdk`) to ensure the package installs cleanly
- Declared as a conflict with `autogen` and `most` extras due to protobuf version incompatibility

**Example Configuration**
- `examples/RAG/simple_rag/configs/rag_library_mode_config.yml` demonstrating the plugin
- Updated README with setup instructions, bootstrap data commands, and example output

**Ingestion Script Enhancement**
- Added `--drop_collection` and `--embedding_model` flags to `scripts/langchain_web_ingest.py` for easier re-ingestion with different embedding models


examples/RAG/simple_rag/configs/rag_library_mode_config.yml
```diff
@@ -0,0 +1,90 @@
...
+  generate:
+    _type: rag_lib_generate
+    rag_client: rag_client
+    topic: NVIDIA technologies
+    description: "Generate a high-quality, self-verified answer with citations. Uses reflection to reduce errors."
+    collection_names:
+      - cuda_docs
+      - mcp_docs
+    use_knowledge_base: true
+    enable_citations: true
```
willkill07: If `search` and `generate` had subconfigs under `nvidia_rag_lib` then you wouldn't need these standalone functions and could actually rely on a Function Group.

As written, the group has no functions, so it's kind of confusing to see the complete separation of group (which is effectively a client) and its functions.

examples/deploy/docker-compose.milvus.yml
```diff
@@ -21,8 +21,8 @@ services:
         MINIO_ACCESS_KEY: minioadmin
         MINIO_SECRET_KEY: minioadmin
       ports:
-        - "9001:9001"
-        - "9000:9000"
+        - "19001:9001"
+        - "19000:9000"
```
willkill07: Why are these ports being changed? These are non-standard.

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,234 @@
...
+
+from nvidia_rag.utils.configuration import EmbeddingConfig as NvidiaRAGEmbeddingConfig
+from nvidia_rag.utils.configuration import FilterExpressionGeneratorConfig as NvidiaRAGFilterGeneratorConfig
+from nvidia_rag.utils.configuration import LLMConfig as NvidiaRAGLLMConfig
+from nvidia_rag.utils.configuration import NvidiaRAGConfig
+from nvidia_rag.utils.configuration import QueryDecompositionConfig as NvidiaRAGQueryDecompositionConfig
+from nvidia_rag.utils.configuration import QueryRewriterConfig as NvidiaRAGQueryRewriterConfig
+from nvidia_rag.utils.configuration import ReflectionConfig as NvidiaRAGReflectionConfig
+from nvidia_rag.utils.configuration import VectorStoreConfig as NvidiaRAGVectorStoreConfig
+from nvidia_rag.utils.configuration import VLMConfig as NvidiaRAGVLMConfig
```
willkill07: Shouldn't we defer these as much as possible?

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/config.py
```diff
@@ -0,0 +1,74 @@
...
+from nat.data_models.component_ref import EmbedderRef
+from nat.data_models.component_ref import LLMRef
+from nat.data_models.component_ref import RetrieverRef
+from nat.embedder.nim_embedder import NIMEmbedderModelConfig
+from nat.llm.nim_llm import NIMModelConfig
+from nat.retriever.milvus.register import MilvusRetrieverConfig
+from nat.retriever.nemo_retriever.register import NemoRetrieverConfig
+
+# Type aliases for component configuration
+LLMConfigType = NIMModelConfig | NvidiaRAGLLMConfig | LLMRef | None
```
willkill07: Why are we letting LLMConfigType be anything other than `LLMRef` ?

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/config.py
```diff
@@ -0,0 +1,74 @@
...
+
+EmbedderConfigType = NIMEmbedderModelConfig | NvidiaRAGEmbeddingConfig | EmbedderRef | None
```
willkill07: Why are we letting EmbedderConfigType be anything other than `EmbedderRef`?

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/config.py
```diff
@@ -0,0 +1,74 @@
...
+
+RetrieverConfigType = MilvusRetrieverConfig | NemoRetrieverConfig | NvidiaRAGVectorStoreConfig | RetrieverRef | None
```
willkill07: Why are we letting RetrieverConfigType be anything other than `RetrieverRef` ?

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/tools/generate.py
```diff
@@ -0,0 +1,132 @@
...
+                                      ge=0.0,
+                                      le=2.0,
+                                      description="Sampling temperature. None uses client default.")
+    max_tokens: int | None = Field(default=None,
+                                   ge=1,
+                                   description="Maximum tokens to generate. None uses client default.")
+    top_p: float | None = Field(default=None,
+                                ge=0.0,
+                                le=1.0,
+                                description="Nucleus sampling probability. None uses client default.")
```
willkill07: I don't understand why a user would want to specify this as part of the configuration for the generate tool over the LLM configuration.

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/tools/generate.py
```diff
@@ -0,0 +1,132 @@
...
+                             description="Tool description. Use {topic} placeholder to include topic.")
+    collection_names: list[str] | None = Field(default=None,
+                                               description="Collections for context. None uses client defaults.")
+    use_knowledge_base: bool = Field(default=True, description="Whether to use RAG (True) or pure LLM (False).")
+    confidence_threshold: float | None = Field(
+        default=None, ge=0.0, le=1.0, description="Minimum relevance score for context. None uses client default.")
+    enable_citations: bool | None = Field(default=None,
+                                          description="Whether to include citations. None uses client default.")
+    enable_guardrails: bool | None = Field(default=None,
+                                           description="Whether to enable guardrails. None uses client default.")
```
willkill07: If you switch to a function group you wouldn't need these at all. (same with `collection_names`)

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/tools/search.py
```diff
@@ -0,0 +1,93 @@
...
+    confidence_threshold: float | None = Field(default=None,
+                                               ge=0.0,
+                                               le=1.0,
+                                               description="Minimum relevance score. None uses client default.")
+    enable_query_rewriting: bool | None = Field(default=None,
+                                                description="Whether to rewrite queries. None uses client default.")
+    enable_reranker: bool | None = Field(default=None,
+                                         description="Whether to use reranking. None uses client default.")
+    enable_filter_generator: bool | None = Field(
+        default=None, description="Whether to auto-generate filters. None uses client default.")
```
willkill07: Again, for things that are optional that fall back to client defaults, I don't think we want to surface another configuration location.

willkill07 (commented):
Just a round of comments to think about the configuration space.

Right now there are so many knobs and required interconnection for this to work. I think it can be simplified a lot without reducing any functionality.

Remember, this is RAG Library Mode for NeMo Agent Toolkit, not exposing every feature of RAG Library Mode within NeMo Agent Toolkit.

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/config.py
ericevans-nv: The union explicitly defines which retriever types are supported by the RAG library. If a user passes a RetrieverRef that resolves to an unsupported type, they’ll be able to see which retriever types are supported, since not all retrievers are compatible.

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/config.py
ericevans-nv: The union explicitly defines which embedder types are supported by the RAG library. If a user passes a EmbedderRef that resolves to an unsupported type, they’ll be able to see which embedders types are supported, since not all embedder are compatible.

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/config.py
ericevans-nv: The union explicitly defines which llm types are supported by the RAG library. If a user passes a LLMRef that resolves to an unsupported type, they’ll be able to see which llm types are supported, since not all llms are compatible.

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,308 @@
...
+    Exposes search and generate tools that share a single RAG client.
+    """
+
+    llm: LLMConfigType = Field(default=None, description="LLM configuration")
+    embedder: EmbedderConfigType = Field(default=None, description="Embedder configuration")
+    retriever: RetrieverConfigType = Field(default=None, description="Vector store configuration")
+    rag_pipeline: RAGPipelineConfig = Field(default_factory=RAGPipelineConfig)
+
+    topic: str | None = Field(default=None, description="Topic for tool descriptions.")
+    collection_names: list[str] | None = Field(default=None, description="Collections to query.")
```
willkill07: Correct me if I'm wrong here, but aren't all of these required?
```suggestion
    llm: LLMConfigType = Field(description="LLM configuration")
    embedder: EmbedderConfigType = Field(description="Embedder configuration")
    retriever: RetrieverConfigType = Field(description="Vector store configuration")
    rag_pipeline: RAGPipelineConfig = Field(default_factory=RAGPipelineConfig)

    topic: str = Field(default, description="Topic for tool descriptions.")
    collection_names: list[str] = Field(default_factory=list, description="Collections to query.")
```

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,308 @@
...
+    )
+
+    await _resolve_llm_config(config.llm, builder, rag_config)
+    await _resolve_embedder_config(config.embedder, builder, rag_config)
+    await _resolve_retriever_config(config.retriever, builder, rag_config)
+
+    return rag_config
+
+
+async def _resolve_llm_config(llm: LLMConfigType, builder: Builder, rag_config: "NvidiaRAGConfig") -> None:
```
willkill07: This function still has a lot of extra branching

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,308 @@
...
+        if "server_url" not in rag_config.filter_expression_generator.model_fields_set and llm.base_url:
+            rag_config.filter_expression_generator.server_url = llm.base_url
+        if "api_key" not in rag_config.filter_expression_generator.model_fields_set and llm.api_key:
+            rag_config.filter_expression_generator.api_key = llm.api_key
+        return
+
+    raise ValueError(f"Unsupported LLM config type: {type(llm)}")
+
+
+async def _resolve_embedder_config(embedder: EmbedderConfigType, builder: Builder,
```
willkill07: This function still has a lot of extra branching no longer required.

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,308 @@
...
+        if embedder.api_key:
+            rag_config.embeddings.api_key = embedder.api_key
+        if embedder.dimensions is not None:
+            rag_config.embeddings.dimensions = embedder.dimensions
+        return
+
+    raise ValueError(f"Unsupported embedder config type: {type(embedder)}")
+
+
+async def _resolve_retriever_config(retriever: RetrieverConfigType, builder: Builder,
```
willkill07: This function has a lot of extra branching due to the reduced configuration space

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/config.py
willkill07: I know that, but that doesn't mean the toolkit needs to expose it to be anything. We can mandate people define LLMs, Embedders, and Retrievers using toolkit constructs.

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/config.py
willkill07: I still disagree on this because you can add validation for component refs by grabbing the config and analyzing

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/config.py
willkill07: Same here. We are making it more cumbersome by enabling users to inject configurations

packages/nvidia_nat_rag_lib/pyproject.toml
```diff
@@ -0,0 +1,55 @@
...
+
+[project]
+name = "nvidia-nat-rag-lib"
+dynamic = ["version"]
+dependencies = [
+  # Keep package version constraints as open as possible to avoid conflicts with other packages. Always define a minimum
+  # version when adding a new package. If unsure, default to using `~=` instead of `==`. Does not apply to nvidia-nat packages.
+  # Keep sorted!!!
+  "nvidia-nat~=1.5",
+  "nvidia-rag>=2.4.0",  # TODO: Update version constraint when nvidia-rag is published to PyPI
```
mnajafian-nv: I suggest removing TODO comments

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,308 @@
...
+        try:
+            citations: Citations = await rag_client.search(
+                query=query,
+                collection_names=config.collection_names,
+                reranker_top_k=config.reranker_top_k,
+            )
+            return RAGSearchResult(citations=citations)
+        except Exception:
+            logger.exception("RAG search failed")
+            return RAGSearchResult(citations=Citations(total_results=0, results=[]))
```
mnajafian-nv:
```suggestion
    async def search(query: str) -> RAGSearchResult:
        """Search for relevant documents."""
        from nvidia_rag.rag_server.response_generator import Citations
        
        try:
            citations: Citations = await rag_client.search(
                query=query,
                collection_names=config.collection_names,
                reranker_top_k=config.reranker_top_k,
            )
            return RAGSearchResult(citations=citations)
        except (ConnectionError, TimeoutError) as e:
            logger.error("RAG search connection failed: %s", e)
            return RAGSearchResult(citations=Citations(total_results=0, results=[]))
        except ValueError as e:
            logger.error("RAG search validation error: %s", e)
            return RAGSearchResult(citations=Citations(total_results=0, results=[]))
        except Exception:
            logger.exception("RAG search failed with unexpected error")
            return RAGSearchResult(citations=Citations(total_results=0, results=[]))
```

mnajafian-nv: Catching `Exception` too broadly without contextual information, which makes production debugging difficult, my suggestion aims to provides specific error context for operations teams

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,308 @@
...
+                    parsed: ChainResponse = ChainResponse.model_validate_json(raw_chunk)
+                    if parsed.choices:
+                        choice = parsed.choices[0]
+                        if choice.delta and choice.delta.content:
+                            content = choice.delta.content
+                            if isinstance(content, str):
+                                chunks.append(content)
+                    if parsed.citations and parsed.citations.results:
+                        final_citations = parsed.citations
+                except Exception:
```
mnajafian-nv: Silently ignores parsing errors

mnajafian-nv:
```suggestion
                try:
                    parsed: ChainResponse = ChainResponse.model_validate_json(raw_chunk)
                    if parsed.choices:
                        choice = parsed.choices[0]
                        if choice.delta and choice.delta.content:
                            content = choice.delta.content
                            if isinstance(content, str):
                                chunks.append(content)
                    if parsed.citations and parsed.citations.results:
                        final_citations = parsed.citations
                except (json.JSONDecodeError, ValidationError) as e:
                    logger.debug("Skipping malformed chunk: %s", e)  # Expected during streaming
                    continue
                except Exception:
                    logger.warning("Unexpected error parsing chunk", exc_info=True)
                    continue
```

mnajafian-nv: Bare exception handler without specific error context

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,308 @@
...
+
+        if "model_name" not in rag_config.filter_expression_generator.model_fields_set:
+            rag_config.filter_expression_generator.model_name = llm.model_name
+        if "server_url" not in rag_config.filter_expression_generator.model_fields_set and llm.base_url:
+            rag_config.filter_expression_generator.server_url = llm.base_url
+        if "api_key" not in rag_config.filter_expression_generator.model_fields_set and llm.api_key:
+            rag_config.filter_expression_generator.api_key = llm.api_key
+        return
+
+    raise ValueError(f"Unsupported LLM config type: {type(llm)}")
```
mnajafian-nv:
```suggestion
    raise ValueError(
        f"Unsupported LLM config type: {type(llm).__name__}. "
        f"Expected one of: NIMModelConfig, NvidiaRAGLLMConfig, LLMRef, or None"
    )
```

mnajafian-nv: Better error messages with actual type and expected types

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,308 @@
...
+        rag_config.embeddings.model_name = embedder.model_name
+        if embedder.base_url:
+            rag_config.embeddings.server_url = embedder.base_url
+        if embedder.api_key:
+            rag_config.embeddings.api_key = embedder.api_key
+        if embedder.dimensions is not None:
+            rag_config.embeddings.dimensions = embedder.dimensions
+        return
+
+    raise ValueError(f"Unsupported embedder config type: {type(embedder)}")
```
mnajafian-nv:
```suggestion
    raise ValueError(
        f"Unsupported embedder config type: {type(embedder).__name__}. "
        f"Expected one of: NIMEmbedderModelConfig, NvidiaRAGEmbeddingConfig, EmbedderRef, or None"
    )
```

mnajafian-nv: Better error messages with actual type and expected types

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,308 @@
...
+
+    if isinstance(retriever, MilvusRetrieverConfig):
+        rag_config.vector_store.url = str(retriever.uri)
+        if retriever.collection_name:
+            rag_config.vector_store.default_collection_name = retriever.collection_name
+        if retriever.connection_args:
+            if "user" in retriever.connection_args:
+                rag_config.vector_store.username = retriever.connection_args["user"]
+            if "password" in retriever.connection_args:
+                rag_config.vector_store.password = SecretStr(retriever.connection_args["password"])
```
mnajafian-nv:
```suggestion
            if "password" in retriever.connection_args:
                pwd = retriever.connection_args["password"]
                if isinstance(pwd, SecretStr):
                    rag_config.vector_store.password = pwd
                elif isinstance(pwd, str):
                    logger.warning("Password should be SecretStr to avoid exposure in logs")
                    rag_config.vector_store.password = SecretStr(pwd)
                else:
                    raise ValueError(f"Password must be str or SecretStr, got {type(pwd).__name__}")
```

mnajafian-nv: There is a password exposure risk in connection_args
- If `connection_args["password"]` is already logged/printed before this point, secret is exposed
- SecretStr wrapping after-the-fact doesn't retroactively protect

packages/nvidia_nat_rag/src/nat/plugins/rag/client.py
```diff
@@ -0,0 +1,308 @@
...
+    async def search(query: str) -> RAGSearchResult:
+        """Search for relevant documents."""
+        from nvidia_rag.rag_server.response_generator import Citations
+
+        try:
+            citations: Citations = await rag_client.search(
+                query=query,
+                collection_names=config.collection_names,
+                reranker_top_k=config.reranker_top_k,
+            )
```
mnajafian-nv, line 161: No Validation for Empty collection_names: Unclear error messages if user forgets to set `collection_names`and fails deep in nvidia_rag client instead of at config validation. 

2 options
```
# In NvidiaRAGLibConfig validation
from pydantic import field_validator

@field_validator('collection_names')
@classmethod
def validate_collection_names(cls, v: list[str] | None) -> list[str] | None:
    if v is not None and len(v) == 0:
        raise ValueError("collection_names must not be empty if provided")
    return v

# Or add validation in nvidia_rag_lib function
async def nvidia_rag_lib(config: NvidiaRAGLibConfig, builder: Builder) -> AsyncGenerator[FunctionGroup, None]:
    if not config.collection_names:
        raise ValueError(
            "collection_names must be specified for NVIDIA RAG Library. "
            "Add 'collection_names: [\"your_collection\"]' to your config."
        )
    # ... rest of function
```

packages/nvidia_nat_rag_lib/tests/test_tools.py
```diff
@@ -0,0 +1,146 @@
...
+from nat.plugins.rag_lib.client import NvidiaRAGLibConfig
+from nat.plugins.rag_lib.client import nvidia_rag_lib
+from nat.plugins.rag_lib.models import RAGGenerateResult
+from nat.plugins.rag_lib.models import RAGSearchResult
+
+
+class TestNvidiaRAGLib:
+
+    @pytest.fixture
+    def mock_builder(self) -> MagicMock:
```
mnajafian-nv:
```suggestion    
      @pytest.fixture(name="mock_builder")
      def fixture_mock_builder(self) -> MagicMock:
```

mnajafian-nv: Pytest fixtures not following naming convention, missing Fixture name= Parameter

packages/nvidia_nat_rag_lib/tests/test_tools.py
```diff
@@ -0,0 +1,146 @@
...
+        builder = MagicMock(spec=Builder)
+        builder.get_llm_config = MagicMock(return_value=None)
+        builder.get_embedder_config = MagicMock(return_value=None)
+        builder.get_retriever_config = AsyncMock(return_value=None)
+        return builder
+
+    @pytest.fixture
+    def config(self) -> NvidiaRAGLibConfig:
```
mnajafian-nv:
```suggestion
    @pytest.fixture(name="config")
    def fixture_config(self) -> NvidiaRAGLibConfig:
```

mnajafian-nv: Pytest fixtures not following naming convention, missing Fixture name= Parameter

packages/nvidia_nat_rag_lib/tests/test_tools.py
```diff
@@ -0,0 +1,146 @@
...
+        return NvidiaRAGLibConfig(collection_names=["test_collection"], )
+
+    @pytest.fixture
+    def mock_rag_client(self) -> MagicMock:
```
mnajafian-nv:
```suggestion
      @pytest.fixture(name="mock_rag_client")
      def fixture_mock_rag_client(self) -> MagicMock:
```

mnajafian-nv: Pytest fixtures not following naming convention, missing Fixture name= Parameter

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,308 @@
...
+from nat.retriever.nemo_retriever.register import NemoRetrieverConfig
+
+if TYPE_CHECKING:
+    from nvidia_rag.rag_server.response_generator import RAGResponse
+    from nvidia_rag.utils.configuration import NvidiaRAGConfig
+
+logger: Logger = logging.getLogger(__name__)
+
+
+class NvidiaRAGLibConfig(FunctionGroupBaseConfig, name="nvidia_rag_lib"):
```
mnajafian-nv: Consider adding timeout parameters for search/generate operations in your In NvidiaRAGLibConfig

search_timeout_seconds: float = Field(default=30.0, description="Timeout for search operations")
generate_timeout_seconds: float = Field(default=120.0, description="Timeout for generate operations")

mnajafian-nv (commented):
Excellent work on this PR, I added some in line comments and suggestions as well as following two questions :

Q1) Can we add observability for production monitoring?
Right now you're logging errors, which is great! But in production, you'll want to track performance trends and alert on degradation. A quick win could be adding structured metrics to your search/generate functions. 

Why this matters:

- Track p50/p95/p99 latencies - "Search is suddenly taking 3 seconds instead of 500ms"
- Monitor error rates - "We're seeing 5% failures on the generate endpoint"
- Correlate quality metrics - "Low citation counts after the embedder upgrade"
- Set up alerts - "Page if search latency > 2s for 5 minutes"

What you can do with these metrics:

- Export to Prometheus/Datadog for dashboards
- Set up PagerDuty alerts on SLO violations
- Track before/after performance for changes
- Debug production incidents with structured data

Q2) Can we level up the config documentation? e.g.  Adding examples, constraints, and gotchas to your docstrings.

Your config fields have descriptions, but developers will have questions like:
- "What's a good value for confidence_threshold?"
- "What's the performance impact of enabling query rewriting?"
- "Help, I'm getting no results, what did I misconfigure?"

packages/nvidia_nat_rag_lib/pyproject.toml
ericevans-nv: Correct. Once the package is uploaded to PyPI, the TODO will be removed.

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
ericevans-nv: The rag library handles timeouts at the HTTP client level. Adding timeouts here would create double-timeout complexity.

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,216 @@
...
+    """Configuration for NVIDIA RAG Library.
+
+    Exposes search and generate tools that share a single RAG client.
+    """
+    llm: LLMRef = Field(description="LLM reference")
+    embedder: EmbedderRef = Field(description="Embedder reference")
+    retriever: RetrieverRef = Field(description="Retriever reference")
+    rag_pipeline: RAGPipelineConfig = Field(default_factory=RAGPipelineConfig)
```
willkill07: Can we add a description?

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,216 @@
...
+    )
+
+    # resolve LLM config
+    nim_llm_config = builder.get_llm_config(config.llm)
+    if not isinstance(nim_llm_config, NIMModelConfig):
+        raise ValueError(f"Unsupported LLM config type: {type(config.llm)}. Expected NIMModelConfig.")
+
+    base_dict = nim_llm_config.model_dump(include={"base_url", "model_name", "api_key"}, exclude_none=True)
+    if "base_url" not in base_dict:
+        raise ValueError("base_url is required for LLM config specified in in NVIDIA RAG Config.")
```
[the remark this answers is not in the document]
willkill07: This is a good suggestion :)

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/client.py
```diff
@@ -0,0 +1,216 @@
...
+    rag_config.filter_expression_generator = rag_config.filter_expression_generator.model_copy(update=base_dict)
+
+    # resolve embedder config
+    nim_embedder_config = builder.get_embedder_config(config.embedder)
+    if not isinstance(nim_embedder_config, NIMEmbedderModelConfig):
+        raise ValueError(f"Unsupported embedder config type: {type(config.embedder)}. Expected NIMEmbedderModelConfig.")
+    base_dict = nim_embedder_config.model_dump(include={"base_url", "model_name", "api_key", "dimensions"},
+                                               exclude_none=True)
+    if "base_url" not in base_dict:
+        raise ValueError("base_url is required for embedder config specified in in NVIDIA RAG Config.")
```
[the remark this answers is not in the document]
willkill07: Same

packages/nvidia_nat_rag/src/nat/plugins/rag/client.py
```diff
@@ -0,0 +1,216 @@
...
+                    if parsed.choices:
+                        choice = parsed.choices[0]
+                        if choice.delta and choice.delta.content:
+                            content = choice.delta.content
+                            if isinstance(content, str):
+                                chunks.append(content)
+                    if parsed.citations and parsed.citations.results:
+                        final_citations = parsed.citations
+                except (ValueError, TypeError, KeyError) as e:
+                    logger.debug("Failed to parse RAG response chunk: %s - %s", type(e).__name__, e)
```
willkill07, line 197: Shouldn't we fail here?

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/config.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # Multimodal (optional)
+    vlm: NvidiaRAGVLMConfig | None = Field(default=None)
+
+    # Pipeline flags
+    enable_citations: bool = Field(default=True)
+    enable_guardrails: bool = Field(default=False)
+    enable_vlm_inference: bool = Field(default=False)
+    vlm_to_llm_fallback: bool = Field(default=True)
+    default_confidence_threshold: float = Field(default=0.0)
```
willkill07: Could we have descriptions for all of these?

packages/nvidia_nat_rag_lib/src/nat/plugins/rag_lib/models.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def __str__(self) -> str:
+        return self.citations.model_dump_json()
+
+
+class RAGGenerateResult(RAGResultBase):
+    """RAG generation result."""
+
+    answer: str
+    citations: "Citations | None" = None
```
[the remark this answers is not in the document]
willkill07: This seems like a sensible error that needs addressed.

packages/nvidia_nat_rag_lib/tests/test_models.py
```diff
@@ -0,0 +1,87 @@
...
+
+import pytest
+from nvidia_rag.rag_server.response_generator import Citations
+
+from nat.plugins.rag_lib.models import RAGGenerateResult
+from nat.plugins.rag_lib.models import RAGSearchResult
+
+# Rebuild models to resolve forward references to Citations
+RAGSearchResult.model_rebuild()
+RAGGenerateResult.model_rebuild()
```
willkill07: We shouldn't need these

packages/nvidia_nat_rag/src/nat/plugins/rag/client.py
ericevans-nv: No. Individual chunk parse failures shouldn't abort the entire stream. The outer try/except handles actual operation failures, this inner handler gracefully skips malformed chunks while accumulating successful ones.

packages/nvidia_nat_rag_lib/tests/test_tools.py
```diff
@@ -0,0 +1,161 @@
...
+                assert any(name.endswith("generate") for name in function_names)
+                function_names = list(functions.keys())
+                assert any(name.endswith("search") for name in function_names)
+                assert any(name.endswith("generate") for name in function_names)
+                function_names = list(functions.keys())
+                assert any(name.endswith("search") for name in function_names)
+                assert any(name.endswith("generate") for name in function_names)
+                function_names = list(functions.keys())
+                assert any(name.endswith("search") for name in function_names)
+                assert any(name.endswith("generate") for name in function_names)
```
mnajafian-nv:
```suggestion
    async def test_group_exposes_both_tools(
    self,
    config: NvidiaRAGLibConfig,
    mock_builder: MagicMock,
    mock_rag_client: MagicMock
) -> None:
    with patch("nvidia_rag.NvidiaRAG", return_value=mock_rag_client):
        async with nvidia_rag_lib(config, mock_builder) as group:
            functions = await group.get_all_functions()
            function_names = list(functions.keys())
            assert any(name.endswith("search") for name in function_names)
            assert any(name.endswith("generate") for name in function_names)
```

mnajafian-nv: Copy-paste error: The test_group_exposes_both_tools function has 8x identical assertion blocks

packages/nvidia_nat_rag_lib/pyproject.toml
```diff
@@ -0,0 +1,56 @@
...
+
+
+[tool.uv]
+managed = true
+config-settings = { editable_mode = "compat" }
+
+
+[tool.uv.sources]
+nvidia-nat = { workspace = true }
+nvidia-rag = { path = "vendor/nvidia_rag-2.4.0.dev0-py3-none-any.whl" }  # TODO: Remove when nvidia-rag>=2.4.0 is on PyPI
```
mnajafian-nv: The pyproject.toml references a vendored wheel file that doesn't exist in this PR, which means the package cannot be installed. Since nvidia-rag releases on PyPI around Feb 10 but your release is Feb 2, you have an 8-day gap. You have two options: 
Option (A) Vendor the wheel: include the pre-release wheel file in a vendor directory (tracked via git-lfs), then migrate to PyPI in a patch release after Feb 10, 
or Option (B) Make it optional: mark as optional dependency and delay the feature until Feb 10. 

I recommend Option A because it ships on schedule, users can install immediately, it's standard practice for pre-release dependencies, and the 8-day gap is short enough that vendoring is perfectly reasonable. Just document why it's temporary and create a follow-up issue for the migration.

packages/nvidia_nat_rag/src/nat/plugins/rag/client.py
```diff
@@ -0,0 +1,217 @@
...
+            rag_config.vector_store.url = str(retriever_config.uri)
+            if retriever_config.collection_name:
+                rag_config.vector_store.default_collection_name = retriever_config.collection_name
+            if retriever_config.nvidia_api_key:
+                rag_config.vector_store.api_key = retriever_config.nvidia_api_key
+            if retriever_config.top_k:
+                rag_config.retriever.top_k = retriever_config.top_k
+        case _:
+            raise ValueError(f"Unsupported retriever config type: {type(retriever_config)}")
+
```
mnajafian-nv, line 148:
```suggestion
    from tenacity import retry, stop_after_attempt, wait_exponential

    @retry(stop=stop_after_attempt(3), wait=wait_exponential(multiplier=1, min=2, max=10))
    def _init_rag_client(config: NvidiaRAGConfig) -> NvidiaRAG:
        return NvidiaRAG(config=config)
```

mnajafian-nv: Suggestion: Consider Adding Retry Logic for RAG Client Initialization:  The NvidiaRAG client is initialized once. If the external service is temporarily unavailable, the entire workflow fails. Consider wrapping the initialization with retry logic using the tenacity library to improve production resilience.

packages/nvidia_nat_rag/tests/test_rag_function.py
```diff
@@ -0,0 +1,299 @@
...
+    @pytest.mark.parametrize("llm_ref", list(LLM_CONFIGS.keys()))
+    @pytest.mark.parametrize(
+        "embedder_ref",
+        [
+            "nim_embedder",
+            pytest.param(
+                "nim_embedder_e5",
+                marks=pytest.mark.xfail(reason="nvidia_rag passes dimensions param which nv-embedqa-e5-v5 rejects")),
+        ])
+    @pytest.mark.parametrize("retriever_ref", list(RETRIEVER_CONFIGS.keys()))
```
mnajafian-nv, line 197:
```suggestion
    @pytest.mark.parametrize("retriever_ref", list(RETRIEVER_CONFIGS.keys()))
    @pytest.mark.timeout(300)  # 5 minute timeout for live NIM calls

```

mnajafian-nv: I suggest adding Timeout Configuration to Integration Tests: It prevents tests from hanging indefinitely when NIMs are slow or rate-limited. Apply the same marker to` test_generate` and `test_health. `

examples/RAG/simple_rag/README.md
```diff
@@ -356,3 +368,171 @@ The final workflow result should look similar to the following:
...
+
+Install the NVIDIA RAG Library:
+```bash
+uv pip install -e packages/nvidia_nat_rag_lib
+```
+
+### Bootstrap Data
+
+The NVIDIA RAG Library example uses a different embedding model (`nvidia/llama-3.2-nv-embedqa-1b-v2`) than the basic quickstart. If you have an existing `cuda_docs` collection from the quickstart, drop and re-ingest with the correct embedding model:
+
```
mnajafian-nv, line 402:
```suggestion
**Important:** Different embedding models are incompatible. If you have an existing `cuda_docs` collection from the quickstart, you MUST drop and re-ingest with the correct embedding model:

```

mnajafian-nv: I suggest clarifying embedding model compatibility in documentation: It makes the requirement more prominent and prevents users from debugging incompatible embeddings.

packages/nvidia_nat_rag/src/nat/plugins/rag/client.py
```diff
@@ -0,0 +1,217 @@
...
+            citations: Citations = await rag_client.search(
+                query=query,
+                collection_names=config.collection_names,
+                reranker_top_k=config.reranker_top_k,
+            )
+            return RAGSearchResult(citations=citations)
+        except Exception:
+            logger.exception("RAG search failed")
+            raise
+
```
mnajafian-nv, line 166:
```suggestion
    # Server-Sent Events (SSE) protocol prefix for streaming responses
```

mnajafian-nv: I suggest adding comment for SSE protocol parsing to clarify the purpose of this magic string for future maintainers.

mnajafian-nv (approved):
Conditional approval after addressing the new inline change requests and comments and suggestions.

packages/nvidia_nat_rag_lib/pyproject.toml
ericevans-nv: The wheel is included in the PR, but it will not be part of the 1.4 release and is instead targeting 1.5. The intent is not to merge until the official package is published to PyPI and all tests and examples pass.

packages/nvidia_nat_rag/src/nat/plugins/rag/client.py
ericevans-nv: This is not needed, as no external service is required to import or instantiate the NvidiaRAG class. These checks are intended to ensure the package can be properly installed and imported, and that all listed components are compatible with the class.

packages/nvidia_nat_rag/tests/test_rag_function.py
ericevans-nv: This is not needed as these tests should complete within the globally configured pytest timeout, and any that exceed it will be caught automatically.

packages/nvidia_nat_rag_lib/pyproject.toml
```diff
@@ -0,0 +1,56 @@
...
+include = ["nat.*"]
+
+
+[tool.setuptools_scm]
+git_describe_command = "git describe --long --first-parent"
+root = "../.."
+
+
+[project]
+name = "nvidia-nat-rag-lib"
```
mdemoret-nv: I dislike this package name. Can we ask the RAG blueprint team if there is a better alternative. Like `nvidia-nat-foundation-rag`, `nvidia-nat-rag-bp`, or something more specific? Adding `_lib` in the name is redundant. Everything is a library.

ericevans-nv: Dropping the _lib. The team decided to simply go with nvidia-nat-rag.

packages/nvidia_nat_rag/src/nat/plugins/rag/client.py
```diff
@@ -0,0 +1,218 @@
...
+        from nvidia_rag.rag_server.response_generator import ChainResponse
+        from nvidia_rag.rag_server.response_generator import Citations
+        from nvidia_rag.utils.configuration import FilterExpressionGeneratorConfig
+        from nvidia_rag.utils.configuration import NvidiaRAGConfig
+        from nvidia_rag.utils.configuration import QueryDecompositionConfig
+        from nvidia_rag.utils.configuration import QueryRewriterConfig
+        from nvidia_rag.utils.configuration import ReflectionConfig
+        from nvidia_rag.utils.configuration import VLMConfig
+    except ImportError as e:
+        raise ImportError("nvidia-rag package is not installed.") from e
```
willkill07, line 77: This is impossible to fail

packages/nvidia_nat_rag/pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+[project.urls]
+documentation = "https://docs.nvidia.com/nemo/agent-toolkit/latest/"
+source = "https://github.com/NVIDIA/NeMo-Agent-Toolkit"
+
+[tool.setuptools_dynamic_dependencies]
+dependencies = [
+  # Keep package version constraints as open as possible to avoid conflicts with other packages. Always define a minimum
+  # version when adding a new package. If unsure, default to using `~=` instead of `==`. Does not apply to nvidia-nat packages.
+  # Keep sorted!!!
+  "langchain-openai>=1.1.6,<2.0.0",
```
willkill07: Where is this a dependency? I didn't see an import in the code

packages/nvidia_nat_rag/pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+  "nvidia-nat-core == {version}",
+  "nvidia-rag>=2.4.0",
+  "opentelemetry-api~=1.2",
+  "opentelemetry-sdk~=1.3",
```
willkill07: Why are these dependencies?

willkill07 (approved):
Approving with feedback needing to be addressed before merge.

Also, please update the .pre-commit-config.yml file for the new package's uv.lock (which also needs to be committed)

`uv lock --project packages/nvidia-nat-rag`

packages/nvidia_nat_rag/pyproject.toml
ericevans-nv: I'm not a fan of this either, but these are undeclared dependencies imported at the module level. The package was just uploaded to PyPI early this morning.

packages/nvidia_nat_rag/pyproject.toml
ericevans-nv: Another hidden dependency.

packages/nvidia_nat_rag/pyproject.toml
willkill07: Add a comment to the `pyproject.toml` that says this.

packages/nvidia_nat_rag/pyproject.toml
willkill07: Add a comment to the `pyproject.toml` that says this.

mnajafian-nv:
Looks good overall. Most earlier feedback is already in. A few things before merge.

**Before merge:** I suggest we run ./ci/scripts/license_diff.py and add a short note in the PR (or description) listing new licenses and added packages. That was the remaining conditional from the approval.

**Suggested fixes**: For client.py I suggest we add a short module docstring after the SPDX block (what the module does and how it fits in). For README in the example snippet, I suggest we change “Agent's thoughts” and “Tool's input/response” to “Agent thoughts” and “Tool input/response” so the doc style is consistent (no possessive for inanimate nouns).

**Follow-up:** Simplifying the config surface and adding observability and richer config docstrings can be a separate pass. No need to block this PR on them.

Salonijain27 (approved):

willkill07:
>  suggest we change “Agent's thoughts” and “Tool's input/response” to “Agent thoughts” and “Tool input/response” so the doc style is consistent (no possessive for inanimate nouns).

That's the explicit console output and should not be changed.

willkill07:
/ok to test [3545e7b](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1440/commits/3545e7b5cd5c2e3478cfdb8bbec9cc16647a4f74)

PR #1503, opened 2026-01-28
Enable per-user workflow support in nat eval

### LINKED ISSUE #1311: re-initialize stateful tools during `nat eval ...`

Improvement

High

Some agent tools may be "stateful" per chat session, and should be cleared for a new session. For example, a Todo-List tool to maintain the agent's plans and ideas, or a tool to create and edit a text document's content. In this cases, a new chat session should begin with a clean slate, meaning reset these tools to its initialized state.

The problem I experienced is when running `nat eval`. When eval runs the workflow on multiple input prompts, the tools are initialized only once. This is not a problem for stateless tools such as a web search tool. In fact, it's more efficient this way when running multiple prompts with concurrency. But stateful tools always need a reset per chat session. Otherwise, states for a previous eval run is visible to a new eval run. It shouldn't be the AI agent's burden to restart its tool state. 

Add the ability to specify if a tool should be re-initialized for a new chat session. 

In the config file, change functions or function groups to something like:
```yaml
function_groups:
  file_edit_group:
    _type: file_edit_group
    stateful: true
```

Then, recognize functions with "stateful=true" flag and re-initialized them per new chat conversation. 

### Additional context

_No response_

### Code of Conduct

ericevans-nv:
## Description

Closes #1311

Fixes issue where stateful tools retain state across eval items, causing state contamination during `nat eval` runs.

### Changes
- **New `per_input_user_id` config option**: `eval.general.per_input_user_id` (default: `True`) generates a unique user_id for each eval item (`{user_id}_{item_id}_{uuid}`). For per-user workflows, this creates a fresh workflow instance per eval item, resetting all stateful tools to their initial state. Set to `False` to disable this behavior.
- **Eval data model improvements**: Updated eval config classes (`EvalCustomScriptConfig`, `JobManagementConfig`, `EvalOutputConfig`, `EvalGeneralConfig`) to use Pydantic `Field` with descriptions instead of inline comments.

### Usage
To reset stateful tools during evaluation, use a per-user workflow (decorated with `@register_per_user_function`). When `per_input_user_id` is enabled (default), each eval item will get its own workflow instance with fresh state.

## Summary by CodeRabbit

* **New Features**
  - Added workflow alias for display in the evaluation UI
  - Added per-item user ID generation to isolate state for each evaluation input

* **Improvements**
  - Validates LLM endpoints before runs to catch deployment issues early
  - Evaluation configuration fields now include descriptive metadata and safer defaults
  - Renamed isolate_workflow_state to per_input_user_id and adjusted default behavior


src/nat/builder/eval_per_user_builder.py
```diff
@@ -0,0 +1,190 @@
...
+from nat.data_models.config import Config
+from nat.data_models.function import FunctionBaseConfig
+from nat.data_models.function import FunctionGroupBaseConfig
+
+logger = logging.getLogger(__name__)
+
+
+class EvalPerUserWorkflowBuilder(PerUserWorkflowBuilder):
+    """
+    Builder for eval mode that rebuilds all functions per user.
```
AnuradhaKaruppiah: I  am not sure there is a need for a separate builder. We can use just the existing builder and allow evaluation of 
1. shared workflows with single workflow instance
2. per-user workflows with workflow instance per-input by simply setting the user-id to a different value for each input.

ericevans-nv: The existing PerUserWorkflowBuilder does not rebuild functions that are not decorated as per-user, so simply setting a different user_id is insufficient for functions that need their state reset across workflow runs. Consider an agent with an internal scratchpad tool used to record plans and intermediate thoughts. This state is internal to the agent and does not require per-user decoration for isolation. However, during evaluation, each eval item should start with a clean scratchpad; otherwise, later items inherit previous state from earlier ones, contaminating the results. The EvalPerUserWorkflowBuilder extends the existing builder to allow any function to be rebuilt during evaluation, providing isolation between eval items without requiring per-user decoration.

src/nat/eval/evaluate.py
```diff
@@ -181,8 +180,12 @@ async def run_one(item: EvalInputItem):
             if stop_event.is_set():
                 return "", []
 
-            async with session_manager.session(user_id=self.config.user_id) as session:
-                async with session.run(item.input_obj, runtime_type=RuntimeTypeEnum.EVALUATE) as runner:
+            user_id = self.config.user_id
+            if self.eval_config.general.per_input_user_id:
+                user_id = f"{self.config.user_id}_{item.id}_{uuid4()}"
```
willkill07: This is more than sufficient:
```suggestion
                user_id += f"-{uuid4()}"
```

willkill07 (approved):
Thanks for adding Pydantic model information :)

Minor nit on generating a unique ID. Otherwise, let's rename this PR appropriately?

AnuradhaKaruppiah (commented):
LGTM. thx for the changes @ericevans-nv

PR #1506, opened 2026-01-28
Refactor `_to_dict` for OTLP-compatible JSON serialization

dnandakumar-nv:
## Summary

- Fix OTLP span attribute encoding failures when exporting telemetry (e.g. to LangSmith)
- `_to_dict` in `SpanExporter` previously returned raw Python objects (LangChain messages, dicts with nested `None` values) as `input.value_obj`/`output.value_obj` span attributes, which OTLP's protobuf encoder cannot serialize
- Changed `_to_dict` to always return a JSON string via `json.dumps(result, default=str)`, guaranteeing OTLP compatibility for all input types
- Added 15 new tests covering Pydantic models, lists of messages, nested `None` values, arbitrary objects, and the always-returns-string invariant

## Test plan

- [x] All 39 tests pass in `tests/nat/observability/exporter/test_span_exporter.py` (24 existing + 15 new)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved span export serialization reliability with consistent JSON format handling and enhanced fallback mechanisms.

* **Tests**
  * Added comprehensive test coverage for span serialization across various data types and edge cases.

### DIFF
--- a/src/nat/observability/exporter/span_exporter.py
+++ b/src/nat/observability/exporter/span_exporter.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import logging
 import os
 import re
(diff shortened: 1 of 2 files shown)


src/nat/observability/exporter/span_exporter.py
```diff
@@ -322,20 +323,31 @@ def _process_end_event(self, event: IntermediateStep):
...
+
+        try:
+            if hasattr(data, 'model_dump'):
+                result = data.model_dump(exclude_none=True)
+            elif isinstance(data, dict):
+                result = {k: v for k, v in data.items() if v is not None}
+            elif isinstance(data, list):
+                result = [item.model_dump(exclude_none=True) if hasattr(item, 'model_dump') else item for item in data]
+            else:
+                return str(data)
```
bbednarski9, line 343: I'm worries that the current implementation could be unreliable in the following ways:

1) silent failure str(data) could return non serializable json

2) inconsistent normalization for different inputs like lists and nested dicts.

Something like this could harden it a bit:

'''
def _to_json_string(self,  typing.Any) -> str:

    def _normalize(obj: typing.Any) -> typing.Any:
        # Pydantic models
        if hasattr(obj, "model_dump"):
            dumped = obj.model_dump(exclude_none=True)
            return _normalize(dumped)

        # Dicts: drop None values, normalize values
        if isinstance(obj, dict):
            return {
                k: _normalize(v)
                for k, v in obj.items()
                if v is not None
            }

        # Lists / tuples: normalize each element
        if isinstance(obj, (list, tuple)):
            return [_normalize(item) for item in obj]

        # Anything else: leave as-is; json.dumps(default=str) will sanitize it
        return obj

    try:
        normalized = _normalize(data)
        return json.dumps(normalized, default=str)
    except Exception:
        # Last-resort: always valid JSON string, but clearly a fallback
        return str(data)
'''

tests/nat/observability/exporter/test_span_exporter.py
```diff
@@ -573,3 +575,177 @@ async def test_usage_info_without_token_usage(self, span_exporter):
...
+            "top_none": None,
+        }
+        result = exporter._to_dict(data)
+        assert isinstance(result, str)
+        # Should be valid JSON regardless of nested Nones
+        parsed = json.loads(result)
+        assert "top_none" not in parsed
+        assert parsed["level1"]["level2"][0]["key"] == "value"
+
+    def test_arbitrary_object_falls_back_to_str(self, exporter):
```
bbednarski9, line 695: Would a fallback to a json.dumps be more reliable?

bbednarski9 (commented):
an improvement but left some questions above to see if we can harden the serialization to be more robust

PR #1535, opened 2026-01-31
fix(testing): guard huggingface integration test with importorskip

willkill07:
## Description

Extended nightly run incorrectly tested for huggingface support.

Fix this by directly attempting to import torch

Closes

## Summary by CodeRabbit

* **Tests**
  * Simplified dependency handling by removing module-level import guards and using runtime checks that skip tests when optional ML libraries are unavailable.
  * Deferred importing of optional config/components until after runtime checks to prevent import errors and maintain consistent test behavior when dependencies are present.

### DIFF
--- a/tests/nat/llm_providers/test_langchain_agents.py
+++ b/tests/nat/llm_providers/test_langchain_agents.py
@@ -26,6 +26,7 @@ from nat.llm.azure_openai_llm import AzureOpenAIModelConfig
 from nat.llm.nim_llm import NIMModelConfig
 from nat.llm.openai_llm import OpenAIModelConfig
 
+
 @pytest.mark.integration
 @pytest.mark.usefixtures("nvidia_api_key")
 async def test_nim_langchain_agent():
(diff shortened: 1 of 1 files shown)


tests/nat/llm_providers/test_langchain_agents.py
```diff
@@ -148,14 +142,14 @@ async def test_azure_openai_react_e2e(test_data_dir: str):
...
 
 @pytest.mark.integration
-@pytest.mark.skipif(not HAS_HUGGINGFACE, reason="HuggingFace dependencies (transformers, torch) not installed")
 async def test_huggingface_langchain_agent():
     """
     Test HuggingFace LLM with LangChain/LangGraph agent.
     Requires transformers and torch to be installed (optional dependencies).
     """
-    from nat.llm.huggingface_llm import HuggingFaceConfig
+    pytest.importorskip("torch", reason="HuggingFace dependencies (transformers, torch) not installed")
```
mnajafian-nv, line 152:
```suggestion
    pytest.importorskip("torch", reason="HuggingFace dependencies (transformers, torch) not installed")
    pytest.importorskip("transformers", reason="HuggingFace dependencies (transformers, torch) not installed")

```

mnajafian-nv: users might have `torch` but not `transformers`, which would cause the test to fail with `ValueError` instead of being cleanly skipped.

mnajafian-nv (approved):
Nice refactor! One suggestion: Add `transformers` check too.

PR #1541, opened 2026-02-02
Preserve workflow state across WebSocket reconnections

### LINKED ISSUE #1513: Support WebSocket reconnection for HITL interactions

Improvement

High

When a user refreshes the chat UI during a HITL interaction, the WebSocket connection resets and creates a new handler instance with _user_interaction_response = None. The original workflow remains blocked waiting for a response, but there's no mechanism to route the user's response from the new connection to the original pending interaction. This prevents workflow resumption after refresh, forcing users to abandon in-progress runs and restart from scratch.

The solution establishes a mechanism to persist pending HITL interaction state across connection lifecycles, using conversation_id to link reconnected clients with their original workflow execution. When a user refreshes or reconnects, the system uses the conversation_id to identify which pending HITL interaction the new connection corresponds to and routes the response to resume the original workflow. The system should gracefully handle cases where no pending interaction exists without crashing and introduce minimal overhead during normal operation.

### Additional context

_No response_

### Code of Conduct

ericevans-nv:
## Description

Closes https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues/1513

### Summary

- Persist `WebSocketMessageHandler` instances in a module-level dictionary keyed by `conversation_id`
- On WebSocket reconnect, swap the socket reference on the disconnected handler and copy its state to the new handler
- Re-send pending HITL prompts when a client reconnects so the UI displays them again

### Problem

When a user refreshes the chat UI during a Human-In-The-Loop (HITL) interaction, the WebSocket connection resets and creates a new handler instance. The original workflow remains blocked waiting for a response, but there was no mechanism to route the user's response from the new connection to the original pending interaction. This forced users to abandon in-progress runs and restart from scratch.

### Solution

Introduced a `_conversation_handlers` dictionary that persists handler instances by `conversation_id`. When a new WebSocket connection is established with an existing `conversation_id`:

1. The disconnected handler's `_socket` is updated to the new connection (so the running workflow can send messages through it)
2. The new handler copies the disconnected handler's state (`_user_interaction`, `_running_workflow_task`, etc.)
3. Any pending HITL prompt is re-sent to trigger the UI to display it again

### Test Plan

- Start a workflow with HITL (e.g., `simple_calculator_hitl`)
- Trigger the HITL prompt and wait for it to display
- Refresh the browser page
- Verify the HITL prompt reappears after reconnection
- Submit a response and verify the workflow completes successfully

## Summary by CodeRabbit

* **New Features**
  * Per-conversation WebSocket reconnection support to resume in-progress workflows after disconnects.

* **Bug Fixes**
  * Re-delivery of Human-In-The-Loop prompts on reconnect to prevent lost prompts.

* **Improvements**
  * More reliable handling of pending human responses and conversation state to ensure prompts and responses persist across reconnects.


packages/nvidia_nat_core/src/nat/front_ends/fastapi/message_handler.py
```diff
@@ -57,6 +59,18 @@
 
 logger = logging.getLogger(__name__)
 
+# Module-level handler store keyed by conversation_id for reconnection support
+_conversation_handlers: dict[str, "WebSocketMessageHandler"] = {}
```
willkill07: module-level data is a code smell to me.

What actually owns this? (because it should) It would also reduce the scope to be only where it was needed.

Other components like sessions have a TTL before automatically being cleaned up. The same should be true for the conversation handlers.

For instance, consider a workflow with a HITL request. If the user disconnects and never reconnects, this will never clean up.

packages/nvidia_nat_core/src/nat/front_ends/fastapi/message_handler.py
```diff
@@ -85,14 +99,46 @@ def __init__(self, socket: WebSocket, session_manager: SessionManager, step_adap
...
+    async def _restore_execution_state(self) -> None:
+        """Restore execution state on reconnection by swapping handler state."""
+        conversation_id = self._socket.query_params.get("conversation_id")
+        if not conversation_id or conversation_id not in _conversation_handlers:
+            return
+
+        # Swap socket on disconnected handler so its running workflow can send through new connection
+        _conversation_handlers[conversation_id]._socket = self._socket
+
+        # Copy disconnected handler's state so this handler can receive and process messages
```
willkill07, line 133: Should `_pending_observability_trace` also be copied?

willkill07 (changes requested):

packages/nvidia_nat_core/src/nat/front_ends/fastapi/message_handler.py
ericevans-nv: Good suggestion. The FastApiFrontEndPluginWorker owns the WebSocketMessageHandler, so the module-level dictionary can be moved into that class. For cleanup, the _done_callback on the workflow task already handles teardown when the workflow completes, since its lifetime is tied to the workflow rather than the WebSocket connection. That said, what would you consider a reasonable TTL for handling abandoned workflows?

willkill07: Without creating another configuration variable, I think we can reuse the same value used by the session

packages/nvidia_nat_core/src/nat/front_ends/fastapi/message_handler.py
ericevans-nv: No, the state remains intact and doesn’t need to be restored.

packages/nvidia_nat_core/src/nat/front_ends/fastapi/message_handler.py
ericevans-nv: This isn’t needed, since the handler is automatically destroyed when the workflow completes.

willkill07 (approved):

PR #1555, opened 2026-02-04
fix(milvus): Fix vector_field config mapping and document_id type

rmalani-nv:
## Summary

Two bug fixes for the Milvus retriever:

- **Fix vector_field config mapping**: The config uses `vector_field` but the retriever's `_search` method expects `vector_field_name` as the parameter name. Without this mapping, the vector field from config was being ignored and the default "vector" was always used.

- **Fix document_id type casting**: Milvus returns integer IDs but the `Document` model expects `document_id: str | None` (see `models.py` line 33). This caused Pydantic validation errors when using collections with integer primary keys.

## Test plan

- [ ] Existing unit tests in `tests/nat/retriever/test_retrievers.py` continue to pass
- [ ] Manual testing with a Milvus collection that has:
  - A custom vector field name (not "vector")
  - Integer primary keys

🤖 Generated with [Claude Code](https://claude.ai/code)

## Bug Fixes
- Fixed Milvus retriever configuration parameter mapping to ensure proper binding of optional settings during initialization
- Standardized document identifiers in retrieval results to consistently return strings across all result formats and types, improving data consistency

### DIFF
--- a/packages/nvidia_nat_core/src/nat/retriever/milvus/register.py
+++ b/packages/nvidia_nat_core/src/nat/retriever/milvus/register.py
@@ -84,6 +84,10 @@ async def milvus_retriever_client(config: MilvusRetrieverConfig, builder: Builde
     model_dict = config.model_dump()
     optional_args = {field: model_dict[field] for field in optional_fields if model_dict[field] is not None}
 
+    # Map vector_field config to vector_field_name parameter expected by retriever
+    if "vector_field" in optional_args:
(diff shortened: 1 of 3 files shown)


willkill07:
/ok to test [03db7ca](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1555/commits/03db7ca4e29efafb0541c188a4e5f6e27292c659)

rmalani-nv:
Thanks @willkill07 ! Any feedback on the changes?

willkill07 (approved):

PR #1432, opened 2026-01-20
Adds support for running Open Agent Spec agents as a new workflow type

YashaPushak:
## Description
Adds support for running Open Agent Spec agents as a new workflow type in NeMo Agent Toolkit. 
Closes #1431

## By Submitting this PR I confirm:
- I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/docs/source/resources/contributing/index.md).
- We require that all contributors "sign-off" on their commits. This certifies that the contribution is your original work, or you have rights to submit it under the same license, or a compatible license.
  - Any contribution which contains commits that are not Signed-Off will not be accepted.
- When the PR is ready for review, new or existing tests cover these changes.
- When the PR is ready for review, the documentation is up to date with these changes.

## Summary by CodeRabbit

* **New Features**
  * Added Agent Spec Workflow integration supporting YAML and JSON configuration formats
  * Optional dependency group for Agent Spec components

* **Documentation**
  * New documentation for Agent Spec Workflow integration with configuration examples and usage requirements

* **Tests**
  * Added unit and integration test coverage for Agent Spec configuration validation and workflow functionality

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1566, opened 2026-02-04
Feature/oracle agent spec integration

afourniernv:
## Overview

This PR represents a major refactoring and enhancement of the Agent-Spec integration work originally proposed in PR #1432 ([YashaPushak:feat/agentspec](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1432)). This branch addresses review feedback from @willkill07 and implements a more robust, production-ready Agent-Spec plugin that follows NeMo Agent Toolkit's standard plugin architecture patterns.

## What This Branch Does

This branch implements a complete Oracle Agent-Spec integration plugin for NeMo Agent Toolkit, allowing users to run Agent-Spec YAML/JSON configurations as native NAT workflows with full observability, profiling, and evaluation capabilities.

### Key Features

1. **Agent-Spec Workflow Wrapper**: Converts Agent-Spec YAML/JSON configurations to LangGraph `CompiledStateGraph` components using `pyagentspec`'s LangGraph adapter, then wraps them as NAT Functions following the same pattern as `LanggraphWrapperFunction`.

2. **Flexible Configuration Support**:
   - File-based: `spec_file` for referencing Agent-Spec YAML/JSON files
   - Inline YAML/JSON: `spec_yaml` and `spec_json` for programmatic/dynamic configurations
   - Validator ensures exactly one source is provided

3. **NAT Tool Integration**: 
   - `tool_names` field supports referencing NAT tools via `FunctionRef` or `FunctionGroupRef`
   - Automatic tool resolution through `builder.get_tools()`
   - Merges NAT tools with Agent-Spec-defined tools (NAT tools take precedence)

4. **Component Reuse Support**:
   - `components_registry` for manually mapping component IDs to NAT-managed components (e.g., LLMs)
   - Documented limitations and best practices for component reuse with embedded Agent-Spec configs

5. **Conversation History Management**:
   - `max_history` field for trimming message history before execution
   - Uses `langchain_core.messages.trim_messages` to limit context size

6. **Auto-detection Features**:
   - Automatically detects `ClientTool` usage and configures `MemorySaver` checkpointer when needed

### Architecture

This implementation follows NeMo Agent Toolkit's standard plugin architecture:

- **Plugin Registration**: Uses `@register_function` decorator with proper framework wrappers (`LLMFrameworkEnum.LANGCHAIN`)
- **Function Wrapper Pattern**: `AgentSpecWrapperFunction` follows the same pattern as `LanggraphWrapperFunction`, handling input/output conversion and delegating execution to the underlying LangGraph
- **Separate Package**: Implemented as `nvidia-nat-agent-spec` package with proper dependencies (`nvidia-nat-core`, `pyagentspec[langgraph,langgraph_mcp]`)
- **Type Safety**: Full Pydantic models for configuration (`AgentSpecWrapperConfig`) and input/output (`AgentSpecWrapperInput`, `AgentSpecWrapperOutput`)
- **Error Handling**: Comprehensive error handling with clear error messages

### Changes from Original PR #1432

This branch addresses several architectural concerns raised in the review:

1. **Enhanced E2E Test Validation**: Added comprehensive validation in smoke tests (graph type, message types, content checks)
2. **Type Annotations**: Added explicit return type annotation (`-> Self`) to `_validate_sources` validator
3. **Documentation**: Documented component reuse limitations and best practices for embedded Agent-Spec configurations
4. **Package Structure**: Maintains separate package structure (not an optional dependency group)
5. **Code Organization**: Implementation details properly separated from configuration models

### Testing

- **55 unit tests** covering:
  - Configuration validation (file, YAML, JSON sources)
  - Tool integration (`tool_names`, `tool_registry`)
  - Component registry functionality
  - Message trimming (`max_history`)
  - Error handling and edge cases
  
- **2 end-to-end integration tests** (smoke tests):
  - File-based Agent-Spec YAML execution
  - Inline YAML execution
  - Tests skip gracefully without `NVIDIA_API_KEY`

### Usage Example

functions:
  agent_spec_workflow:
    _type: agent_spec_wrapper
    spec_file: path/to/agent_spec.yaml
    tool_names:
      - FunctionRef("wiki_search")
      - FunctionGroupRef("search_tools")
    max_history: 20
    description: "Agent-Spec workflow with NAT tool integration"
    
    
    
     ## Important: Dependency Conflicts
   
   The `agent-spec` extra cannot be installed alongside `langchain`, `most`, or `vanna` extras due to incompatible `langchain-core`/`langgraph` version requirements:
   - `pyagentspec` requires: `langchain-core<1.0.0, langgraph<1.0.0`
   - `nvidia-nat-langchain` requires: `langchain-core>=1.2.6, langgraph>=1.0.5`
   
   Users must choose one or the other, or use separate environments.

   ## Lock File Updates
   
   The `uv.lock` file has been updated to include dependencies for the new `nvidia-nat-agent-spec` package. This is expected and necessary when adding a new package.

### DIFF
--- a/packages/nvidia_nat_agent_spec/pyproject.toml
+++ b/packages/nvidia_nat_agent_spec/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 build-backend = "setuptools.build_meta"
-requires = ["setuptools >= 64", "setuptools-scm>=8"]
+requires = ["setuptools>=64", "setuptools-scm>=8", "setuptools_dynamic_dependencies>=1.0.0"]
 
 
 [tool.setuptools.packages.find]
@@ -15,14 +15,7 @@ root = "../.."
 
 [project]
 name = "nvidia-nat-agent-spec"
-dynamic = ["version"]
-dependencies = [
-  # Keep package version constraints as open as possible to avoid conflicts with other packages. Always define a minimum
-  # version when adding a new package. If unsure, default to using `~=` instead of `==`. Does not apply to nvidia-nat packages.
-  # Keep sorted!!!
-  "nvidia-nat~=1.5",
-  "pyagentspec[langgraph,langgraph_mcp]>=26.1.0",
-]
+dynamic = ["version", "dependencies"]
 requires-python = ">=3.11,<3.14"
 description = "Subpackage for Oracle Agent-Spec integration in NeMo Agent toolkit"
 readme = "src/nat/meta/pypi.md"
@@ -41,14 +34,18 @@ classifiers = [
 documentation = "https://docs.nvidia.com/nemo/agent-toolkit/latest/"
 source = "https://github.com/NVIDIA/NeMo-Agent-Toolkit"
 
+[tool.setuptools_dynamic_dependencies]
+dependencies = [
+  # Keep package version constraints as open as possible to avoid conflicts with other packages. Always define a minimum
+  # version when adding a new package. If unsure, default to using `~=` instead of `==`. Does not apply to nvidia-nat packages.
+  # Keep sorted!!!
+  "nvidia-nat-core == {version}",
+  "pyagentspec[langgraph,langgraph_mcp]>=26.1.0",
+]
 
 [tool.uv]
 managed = true
 config-settings = { editable_mode = "compat" }
-
-
-[tool.uv.sources]
-nvidia-nat = { workspace = true }
 
 
 [project.entry-points.'nat.components']
--- a/packages/nvidia_nat_agent_spec/pyproject.toml
+++ b/packages/nvidia_nat_agent_spec/pyproject.toml
@@ -1,0 +1,55 @@
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = ["setuptools >= 64", "setuptools-scm>=8"]
+
+
+[tool.setuptools.packages.find]
+where = ["src"]
+include = ["nat.*"]
+
+
+[tool.setuptools_scm]
+git_describe_command = "git describe --long --first-parent"
+root = "../.."
+
+
+[project]
+name = "nvidia-nat-agent-spec"
+dynamic = ["version"]
+dependencies = [
+  # Keep package version constraints as open as possible to avoid conflicts with other packages. Always define a minimum
+  # version when adding a new package. If unsure, default to using `~=` instead of `==`. Does not apply to nvidia-nat packages.
+  # Keep sorted!!!
+  "nvidia-nat~=1.5",
+  "pyagentspec[langgraph,langgraph_mcp]>=26.1.0",
+]
+requires-python = ">=3.11,<3.14"
+description = "Subpackage for Oracle Agent-Spec integration in NeMo Agent toolkit"
+readme = "src/nat/meta/pypi.md"
+keywords = ["ai", "rag", "agents"]
+license = { text = "Apache-2.0" }
+authors = [{ name = "NVIDIA Corporation" }]
+maintainers = [{ name = "NVIDIA Corporation" }]
+classifiers = [
+  "Programming Language :: Python",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+]
+
+[project.urls]
+documentation = "https://docs.nvidia.com/nemo/agent-toolkit/latest/"
+source = "https://github.com/NVIDIA/NeMo-Agent-Toolkit"
+
+
+[tool.uv]
+managed = true
+config-settings = { editable_mode = "compat" }
+
+
+[tool.uv.sources]
+nvidia-nat = { workspace = true }
+
+
+[project.entry-points.'nat.components']
+nat_agent_spec = "nat.plugins.agent_spec.register"
--- a/packages/nvidia_nat_agent_spec/src/nat/plugins/agent_spec/agent_spec_workflow.py
+++ b/packages/nvidia_nat_agent_spec/src/nat/plugins/agent_spec/agent_spec_workflow.py
@@ -27,6 +27,7 @@ Flow:
 6. NAT runtime executes it with evaluation, profiling, observability, middleware, etc.
 """
 
+import json
 import logging
 import yaml
 from collections.abc import AsyncGenerator
@@ -41,6 +42,7 @@ from pydantic import BaseModel
 from pydantic import ConfigDict
 from pydantic import Field
 from pydantic import FilePath
+from pydantic import model_validator
 
 from nat.builder.builder import Builder
 from nat.builder.framework_enum import LLMFrameworkEnum
@@ -74,23 +76,55 @@ class AgentSpecWrapperConfig(FunctionBaseConfig, name="agent_spec_wrapper"):
     converted to a LangGraph CompiledStateGraph and executed as a NAT Function.
 
     Example usage in NAT workflow config:
+        # Option 1: From file
         workflow:
           _type: agent_spec_wrapper
           spec_file: "my_agent_spec.yaml"
           description: "Oracle agent-spec agent"
 
+        # Option 2: Inline YAML
+        workflow:
+          _type: agent_spec_wrapper
+          spec_yaml: |
+            component_type: Agent
+            name: "My Agent"
+            ...
+          description: "Oracle agent-spec agent"
+
+        # Option 3: Inline JSON
+        workflow:
+          _type: agent_spec_wrapper
+          spec_json: '{"component_type": "Agent", "name": "My Agent", ...}'
+          description: "Oracle agent-spec agent"
+
     Attributes:
-        spec_file: Path to the Agent-Spec YAML configuration file (required).
+        spec_file: Path to the Agent-Spec YAML/JSON configuration file (optional if spec_yaml or spec_json provided).
+        spec_yaml: Inline Agent-Spec YAML content (optional if spec_file or spec_json provided).
+        spec_json: Inline Agent-Spec JSON content (optional if spec_file or spec_yaml provided).
         description: Optional description of the workflow.
         tool_registry: Optional dictionary mapping tool names to LangGraph tools
             or callables. If provided, these tools will be available to the
             Agent-Spec workflow.
+
+    Note:
+        Exactly one of spec_file, spec_yaml, or spec_json must be provided.
     """
 
     model_config = ConfigDict(extra="forbid")
 
     description: str = Field(default="", description="Description of the Agent-Spec workflow")
-    spec_file: FilePath = Field(description="Path to the Agent-Spec YAML configuration file")
+    spec_file: FilePath | None = Field(
+        default=None,
+        description="Path to the Agent-Spec YAML/JSON configuration file"
+    )
+    spec_yaml: str | None = Field(
+        default=None,
+        description="Inline Agent-Spec YAML content"
+    )
+    spec_json: str | None = Field(
+        default=None,
+        description="Inline Agent-Spec JSON content"
+    )
     tool_registry: dict[str, Any] | None = Field(
         default=None,
         description="Optional dictionary mapping tool names to LangGraph tools or callables",
@@ -107,6 +141,21 @@ class AgentSpecWrapperConfig(FunctionBaseConfig, name="agent_spec_wrapper"):
         "ClientTool in Agent-Spec YAML. If not provided and ClientTool is detected, "
         "MemorySaver will be used automatically.",
     )
+
+    @model_validator(mode="after")
+    def _validate_sources(self):
+        """Ensure exactly one of spec_file, spec_yaml, or spec_json is provided."""
+        provided = [self.spec_file, self.spec_yaml, self.spec_json]
+        cnt = sum(1 for v in provided if v is not None)
+        if cnt != 1:
+            raise ValueError(
+                "Exactly one of spec_file, spec_yaml, or spec_json must be provided. "
+                f"Found {cnt} provided: "
+                f"spec_file={'provided' if self.spec_file else 'None'}, "
+                f"spec_yaml={'provided' if self.spec_yaml else 'None'}, "
+                f"spec_json={'provided' if self.spec_json else 'None'}"
+            )
+        return self
 
 
 class AgentSpecWrapperFunction(Function[AgentSpecWrapperInput, NoneType, AgentSpecWrapperOutput]):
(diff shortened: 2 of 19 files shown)


packages/nvidia_nat_agent_spec/src/nat/plugins/agent_spec/agent_spec_workflow.py
```diff
@@ -0,0 +1,371 @@
...
+    if config.tool_names:
+        try:
+            tools = await builder.get_tools(tool_names=config.tool_names, wrapper_type=LLMFrameworkEnum.LANGCHAIN)
+            if tools:
+                tool_registry = {tool.name: tool for tool in tools}
+        except Exception as e:
+            logger.warning(f"Failed to resolve tools from tool_names: {e}")
+
+    if config.tool_registry:
+        tool_registry.update(config.tool_registry)
```
willkill07, line 345: We should raise an error on any intersection of tool names rather than blindly take precedence

packages/nvidia_nat_agent_spec/src/nat/plugins/agent_spec/agent_spec_workflow.py
```diff
@@ -0,0 +1,371 @@
...
+        assert config.spec_yaml is not None  # Type narrowing: validator ensures this is set
+        spec_content = config.spec_yaml
+        spec_format = "yaml"
+        source_description = "inline YAML"
+    else:
+        # Use inline JSON (config.spec_json)
+        assert config.spec_json is not None  # Type narrowing: validator ensures this is set
+        spec_content = config.spec_json
+        spec_format = "json"
+        source_description = "inline JSON"
```
willkill07, line 309: Hoist as a `_load_spec` function?

packages/nvidia_nat_agent_spec/src/nat/plugins/agent_spec/agent_spec_workflow.py
```diff
@@ -0,0 +1,371 @@
...
+    """
+    try:
+        from pyagentspec.adapters.langgraph import AgentSpecLoader
+    except ImportError as e:
+        raise ImportError(
+            "Failed to import pyagentspec.adapters.langgraph. "
+            "Install pyagentspec with langgraph extras:\n"
+            "  uv pip install \"pyagentspec[langgraph,langgraph_mcp]>=26.1.0\"\n\n"
+            "Note: The root pyproject.toml includes override-dependencies to resolve "
+            "langchain-core version conflicts between pyagentspec and nvidia-nat-langchain."
```
willkill07, line 281: This should be a message to install `nvidia-nat[agent_spec]`

packages/nvidia_nat_agent_spec/pyproject.toml
```diff
@@ -0,0 +1,52 @@
...
+documentation = "https://docs.nvidia.com/nemo/agent-toolkit/latest/"
+source = "https://github.com/NVIDIA/NeMo-Agent-Toolkit"
+
+[tool.setuptools_dynamic_dependencies]
+dependencies = [
+  # Keep package version constraints as open as possible to avoid conflicts with other packages. Always define a minimum
+  # version when adding a new package. If unsure, default to using `~=` instead of `==`. Does not apply to nvidia-nat packages.
+  # Keep sorted!!!
+  "nvidia-nat-core == {version}",
+  "pyagentspec[langgraph,langgraph_mcp]>=26.1.0",
```
willkill07, line 43: Is unbounded upper-bound really safe? I'd rather see: `~=26.1` (>=26.1,<27.0)

pyproject.toml
```diff
@@ -199,12 +152,20 @@ conflicts = [
     [{ extra = "adk" }, { extra = "ragaai" }],
     [{ extra = "strands" }, { extra = "ragaai" }],
     [{ extra = "most" }, { extra = "ragaai" }],
+    # agent-spec<>langchain have incompatible langchain-core/langgraph versions
+    # pyagentspec (used by agent-spec) requires langchain-core<1.0.0, langgraph<1.0.0
+    # nvidia-nat-langchain requires langchain-core>=1.2.6, langgraph>=1.0.5
+    # These version ranges are incompatible and cannot coexist in the same workspace
+    # Users must choose one or the other, or install agent-spec in a separate environment
```
willkill07, line 159: remove unnecessary wording
```suggestion
```

pyproject.toml
willkill07, line 1: You should also generate a `uv.lock` in this directory

packages/nvidia_nat_agent_spec/tests/conftest.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # If NVIDIA_API_KEY is set but OPENAI_API_KEY is not, set it
+    if nvidia_key and not openai_key:
+        os.environ["OPENAI_API_KEY"] = nvidia_key
+        yield
+        # Clean up: only remove if we set it
+        if "OPENAI_API_KEY" in os.environ:
+            del os.environ["OPENAI_API_KEY"]
+    else:
+        yield
```
willkill07, line 50: This has the potential to be extremely confusing and dangerous. There has to be a better way.

At minimum:
- do not autouse
- give a name such as `map_nvidia_api_key_to_openai_api_key`

packages/nvidia_nat_agent_spec/tests/test_nim_integration.py
```diff
@@ -0,0 +1,122 @@
...
+
+from nat.builder.workflow_builder import WorkflowBuilder
+from nat.plugins.agent_spec.agent_spec_workflow import AgentSpecWrapperConfig, register
+
+
+@pytest.mark.integration
+@pytest.mark.skipif(
+    not os.getenv("NVIDIA_API_KEY"),
+    reason="Requires NVIDIA_API_KEY environment variable"
+)
```
willkill07, line 31: We already have an `nvidia_api_key` fixture that can be used -- it is part of `nvidia_nat_test`

packages/nvidia_nat_agent_spec/tests/test_e2e_nat_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+from pathlib import Path
+
+import pytest
+
+
+@pytest.mark.integration
+@pytest.mark.skipif(
+    not os.getenv("NVIDIA_API_KEY"),
+    reason="Requires NVIDIA_API_KEY environment variable"
+)
```
willkill07, line 40: Prefer `nvidia_api_key` fixture as parameter

packages/nvidia_nat_agent_spec/tests/test_e2e_nat_workflow.py
```diff
@@ -0,0 +1,204 @@
...
+                pytest.skip(f"Invocation failed (may need network/API): {e}")
+
+
+
+
+@pytest.mark.integration
+@pytest.mark.skipif(
+    not os.getenv("NVIDIA_API_KEY"),
+    reason="Requires NVIDIA_API_KEY environment variable"
+)
```
willkill07, line 122: Prefer `nvidia_api_key` fixture as parameter

packages/nvidia_nat_agent_spec/tests/test_nim_integration.py
```diff
@@ -0,0 +1,122 @@
...
+                print(f"✓ Successfully loaded Agent-Spec YAML: {test_yaml}")
+                print(f"✓ Graph type: {type(wrapper_function._graph)}")
+                print(f"✓ Graph compiled successfully")
+
+
+@pytest.mark.integration
+@pytest.mark.skipif(
+    not os.getenv("NVIDIA_API_KEY"),
+    reason="Requires NVIDIA_API_KEY environment variable"
+)
```
willkill07, line 77: Prefer `nvidia_api_key` fixture as parameter

packages/nvidia_nat_agent_spec/tests/fixtures/weather_agent_with_tool.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+component_type: Agent
+description: 'Weather agent with a client tool'
+name: weather_agent
+inputs: []
+llm_config:
+  component_type: OpenAiCompatibleConfig
+  name: "NIM LLM"
+  model_id: "meta/llama-3.1-8b-instruct"
+  # NIM endpoint - will use NVIDIA_API_KEY from environment
```
willkill07, line 12: Is this true? because it seems like you have to set `OPENAI_API_KEY`

packages/nvidia_nat_agent_spec/tests/fixtures/weather_agent_with_tool.yaml
```diff
@@ -0,0 +1,26 @@
...
+  url: "https://integrate.api.nvidia.com/v1"
```
willkill07, line 13: Just a note on how messed up their implementation is...

https://github.com/oracle/agent-spec/blob/main/pyagentspec/src/pyagentspec/llms/openaicompatibleconfig.py#L29

`OpenAiCompatibleConfig` has an optional `api_key`.

But ironically, they don't even map it!

https://github.com/oracle/agent-spec/blob/main/pyagentspec/src/pyagentspec/adapters/langgraph/_langgraphconverter.py#L1233-L1242

packages/nvidia_nat_agent_spec/src/nat/plugins/agent_spec/agent_spec_workflow.py
```diff
@@ -0,0 +1,371 @@
...
+
+    model_config = ConfigDict(extra="forbid")
+
+    description: str = Field(default="", description="Description of the Agent-Spec workflow")
+    spec_file: FilePath | None = Field(
+        default=None,
+        description="Path to the Agent-Spec YAML/JSON configuration file. "
+        "Recommended for NAT YAML workflow configurations."
+    )
+    spec_yaml: str | None = Field(
```
jschweiz, line 102: here, the workflow config object contains only a stringified version of an agentspec config. This means that when using the optimizer, the optimizer cannot navigate in this config to discover parameters that can be optimized nor the values/ranges they should be optimized on.

I see two ways to make this work.

# Using a dynamic basemodel

This is what I'm currently experimenting with. I basically added 2 attributes to this workflow config, one "dynamic_optimizeable_parameters" that's empty initially, and a "optimizeable_parameters_dict" dict.
It looks like this in the config:
```
workflow:
  _type: agent_spec_wrapper
  spec_file: stage_2.yaml

  # this is required for adding dynamically the optimizable parameters in the config object
  optimizable_params:
    - dynamic_optimizable_parameters

  # dict for optimize-able parameters with values
  optimizable_parameters_dict:
    temperature_ref:
      range:
        low: 0.0
        high: 1.0
        step: 0.1
    max_tokens_ref:
      range:
        low: 2048
        high: 8096
        step: 2048
```
Then, as a post-validation of the `AgentSpecWrapperConfig` object, I create a dynamic base-model:
```
@model_validator(mode="after")
    def _build_prompts_opt(self) -> Self:
        fields: dict[str, tuple[type, Any]] = {}
        for var_name, var in self.optimizable_parameters_dict.items():
            if var.values is not None:
                fields[var_name] = (
                    dict,
                    OptimizableField(
                        default=var.values[0],
                        space=SearchSpace(values=var.values),
                    ),
                )
            elif var.range is not None:
                fields[var_name] = (
                    dict,
                    OptimizableField(
                        default=var.range.low,
                        space=SearchSpace(low=var.range.low, high=var.range.high, step=var.range.step),
                    ),
                )
            else:
                raise ValueError('Not supported')

        DynamicPrompts = create_model("DynamicPrompts", **fields)
        self.dynamic_optimizable = DynamicPrompts()
        return self
```

This allows the optimizer to still see all the parameters inside the config, even if these attributes are not decorated with OptimizableField.


# Cleaner approach: patch our config objects with the "optimizable" annotation

We also represent config objects as base models, so one way would be to extend the loading of the workflow config with our own loading logic, and patch our base models with the optimizable field annotator, for example on `temperature` like in https://github.com/oracle/agent-spec/blob/main/pyagentspec/src/pyagentspec/llms/llmgenerationconfig.py#L23.

This would allow a more seemless and native use of the optimizer with agentspec configs.


Do you see another potential solution, and what do you think about these?

packages/nvidia_nat_agent_spec/src/nat/plugins/agent_spec/agent_spec_workflow.py
```diff
@@ -0,0 +1,371 @@
...
+        "To use this, ensure your Agent-Spec YAML includes an 'id' field in the LLM config, "
+        "then map that ID to your NAT LLM: components_registry={'llm_id': nat_llm}.",
+    )
+    checkpointer: Any | None = Field(
+        default=None,
+        description="Optional LangGraph checkpointer (e.g., MemorySaver). Required when using "
+        "ClientTool in Agent-Spec YAML. If not provided and ClientTool is detected, "
+        "MemorySaver will be used automatically.",
+    )
+    max_history: int = Field(
```
jschweiz, line 144: we support this concept natively in agentspec, I don't think we need this here (we can remove the part about pruning the history before the invoke/stream calls too)

packages/nvidia_nat_agent_spec/src/nat/plugins/agent_spec/agent_spec_workflow.py
```diff
@@ -0,0 +1,371 @@
...
+from nat.builder.function import Function
+from nat.cli.register_workflow import register_function
+from nat.data_models.component_ref import FunctionGroupRef
+from nat.data_models.component_ref import FunctionRef
+from nat.data_models.function import FunctionBaseConfig
+
+logger = logging.getLogger(__name__)
+
+
+class AgentSpecWrapperInput(BaseModel):
```
jschweiz, line 58: we would need to extend to be also a dict of inputs, since soem agentspec config takes inputs and not a list of messages.

packages/nvidia_nat_agent_spec/src/nat/plugins/agent_spec/agent_spec_workflow.py
```diff
@@ -0,0 +1,371 @@
...
+    """Input model for the Agent-Spec wrapper."""
+
+    model_config = ConfigDict(extra="allow")
+
+    messages: list[BaseMessage]
+
+
+class AgentSpecWrapperOutput(BaseModel):
```
jschweiz, line 66: similarly, we would need to support outputting a dict of outputs. Do you already support handling such input/output dictionnaries, and how does it plain with the evaluator (the mappings `question` -> `WrapperInput` and `WrapperOutput` -> `output` are defined below in `AgentSpecWrapperFunction`, is that what we always want?

PR #1567, opened 2026-02-05
Allow for prompts to be stored in/loaded from files

pastorsj:
# What does this PR do?

## Description
Allow prompts to be loaded from external files using file:// prefix in YAML configs. This enables better organization and version control of prompts separately from configs.

### Features:
- Load prompt content when field name ends with "prompt" and value starts with "file://"
- Support relative paths (resolved from config file location) and absolute paths
- Security: restrict to safe file extensions (.txt, .md, .j2, .jinja2, .jinja, .prompt, .tpl, .template)
- Works with config inheritance

Example usage in config.yml:
  workflow: system_prompt: file://../prompts/system_prompt.txt

## Summary by CodeRabbit

* **New Features**
  * Support loading prompts from external files via file:// in configurations for easier editing, sharing, and versioning; content is loaded as plain text with allowed extensions.

* **Documentation**
  * Added detailed guidance and usage rules for file-based prompt loading; guidance duplicated in workflow docs for discoverability.

* **Examples**
  * Added a complete example project showing config, prompt files, and run instructions.

* **Tests**
  * Added extensive tests for file loading, path resolution, inheritance, validation, and error cases.

* **Chores**
  * Updated packaging/config to include the new example.


docs/source/build-workflows/workflow-configuration.md
```diff
@@ -176,3 +176,35 @@ When you run a workflow using `config-variant.yml`, the configurations are combi
...
+workflow:
+  _type: react_agent
+  llm_name: nim_llm
+  tool_names: [calculator]
+  system_prompt: file://../prompts/system_prompt.txt
+```
+
+The file loading follows these rules:
+
+- **Field name requirement**: The configuration key must end with `prompt` (case-insensitive). For example: `system_prompt`, `SystemPrompt`, `human_message_prompt`
```
ericevans-nv: Thanks for this feature! One architectural suggestion:
Instead of checking for field names ending in prompt at the YAML parsing layer, consider using Pydantic field validators on the specific fields that should support file loading:
```python
class ReActAgentWorkflowConfig(AgentBaseConfig):
    system_prompt: str | None = Field(default=None, ...)

    @field_validator('system_prompt', mode='before')
    @classmethod
    def load_from_file(cls, v):
        if isinstance(v, str) and v.startswith("file://"):
            path = Path(v[len("file://"):])
            _validate_prompt_file_extension(path)
            return _load_file_content(path)
        return v
```
**Benefits:**
- No naming convention required - explicit about which fields support file loading
- Self-documenting - the behavior is visible in the config class
- More flexible - could easily add to `additional_instructions` without renaming

The YAML parser shouldn't need to know what a "prompt field" is - that's domain logic that belongs on the field definition itself.

examples/prompt_from_file/configs/config.yml
```diff
@@ -0,0 +1,26 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
```
willkill07: We need the full license at the top rather than the abbreviated two-line SPDX

docs/source/build-workflows/workflow-configuration.md
willkill07: I would in general agree with Eric's suggestion, but the problem is we already handle:
- variable interpolation
- configuration inheritance
at the yaml parser level.

I'm not sure how we could do any sort of relative-to-configuration substitution. If it was just substitution based on CWD that would be another story.

willkill07: I would also remove the restriction that the field for substitution has to end in `prompt`. This could be valuable for descriptions and other components of the configuration file.

The feature makes more sense for expanding `file://` references -> `str` contents.

packages/nvidia_nat_core/src/nat/utils/io/yaml_tools.py
```diff
@@ -130,6 +223,9 @@ def yaml_load(config_path: StrPath, _visited: set[Path] | None = None) -> dict:
         config = deep_merge(base_config, config)
         config.pop("base", None)
 
+    # Resolve file:// references in prompt fields
+    config = _resolve_prompt_files(config, config_path_obj.parent)
+
```
willkill07: I'd rather see this as a separate chain call under `yaml_loads` due to this not technically being a variable interpolation

docs/source/build-workflows/workflow-configuration.md
pastorsj: Sounds good, I can do that.

packages/nvidia_nat_core/src/nat/utils/io/yaml_tools.py
pastorsj: Fair, let me do that.

docs/source/build-workflows/workflow-configuration.md
pastorsj: Fixed

examples/prompt_from_file/configs/config.yml
pastorsj: Fixed

packages/nvidia_nat_core/src/nat/utils/io/yaml_tools.py
pastorsj: This should be updatd @willkill07. Can you double check?

docs/source/build-workflows/workflow-configuration.md
ericevans-nv: What if we just require absolute paths or CWD-relative paths? It simplifies the implementation and users can use environment variables for portability.

pastorsj: @ericevans-nv @willkill07 I guess I was thinking, architecturally, loading it from the yaml file would broadly be more generic and would save us from writing duplicate code every time we wanted to load a field from a file. That would be applicable here since I changed from loading just prompts from files to any property can be loaded from a file.

willkill07 (approved):

willkill07:
/ok to test [54a8f44](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1567/commits/54a8f44e7ba17d2b1292f965b6a6222667d28d85)

willkill07:
/ok to test [4486b6d](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1567/commits/4486b6de12e152d0e8f3b40d8cd379524ee5cd78)

willkill07:
/ok to test [69d4669](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1567/commits/69d466973168f049c3a31511bcf65e00e139aa6f)

PR #1573, opened 2026-02-06
Responses API fixes to comply with Microsoft Foundry Hosted Agents 

antonslutskyms:
## Description

Closes

The following items are addressed in this PR:

1. Making "model" input field Optional
2. Adding support for "type" : "input_text" e.g.: {"type": "message", "role": "user", "content": [{"type": "input_text", "text": "hello"}]}
3. Updating message ID and response ID to comply with length requirement of 50 characters.

## Summary by CodeRabbit

* **New Features**
  * Added OpenAI Responses API support with automatic format detection and an explicit openai_api_v1_format override; supports streaming (SSE) and non-streaming endpoints and OpenAI-compatible response shaping.
  * Improved input compatibility with a new text-input attachment type and enhanced tool-handling for Responses-style requests.

* **Documentation**
  * Added a README and example config demonstrating setup, endpoints, streaming/non-streaming usage, instructions, and test curl commands.

* **Tests**
  * Expanded test coverage for Responses models, conversions, routing/path detection, streaming behavior, and tool handling.

### DIFF
--- a/examples/front_ends/responses_api_endpoint/README.md
+++ b/examples/front_ends/responses_api_endpoint/README.md
@@ -1,0 +1,208 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# OpenAI Responses API Endpoint
+
+**Complexity:** 🟢 Beginner
+
+This example demonstrates how to configure a NeMo Agent toolkit FastAPI frontend to accept requests in the [OpenAI Responses API format](https://platform.openai.com/docs/api-reference/responses).
+
+## Overview
+
+The OpenAI Responses API uses a different request format than the Chat Completions API:
+
+| Feature | Chat Completions API | Responses API |
+|---------|---------------------|---------------|
+| Input field | `messages` (array) | `input` (string or array) |
+| System prompt | In messages array | `instructions` field |
+| Response object | `chat.completion` | `response` |
+| Streaming events | `chat.completion.chunk` | `response.created`, `response.output_text.delta`, etc. |
+
+This example configures the `/v1/responses` endpoint to accept the Responses API format while the standard `/generate` and `/chat` endpoints continue using Chat Completions format.
+
+> **⚠️ Important**: The Responses API format is provided for pass-through compatibility with managed services that support stateful backends (such as OpenAI and Azure OpenAI). NeMo Agent toolkit workflows do not inherently support stateful backends. Features like `previous_response_id` will be accepted but ignored.
+
+## Prerequisites
+
+1. **Install LangChain integration** (required for `tool_calling_agent` workflow):
+
+```bash
+uv pip install -e '.[langchain]'
+```
+
+2. **Set up the NVIDIA API key**:
+
+```bash
+export NVIDIA_API_KEY=<YOUR_API_KEY>
+```
+
+## Start the Server
+
+```bash
+nat serve --config_file examples/front_ends/responses_api_endpoint/configs/config.yml
+```
+
+The server will start on port 8088 with the following endpoints:
+
+| Endpoint | Format | Description |
+|----------|--------|-------------|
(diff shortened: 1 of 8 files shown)


bbednarski9:
This PR does not yet address the issue of stateful agents. This API proxy would lead to some "false advertising". We do not plan to merge this until the above has been addressed

Yash-xoxo:
Good direction overall — the `input_text` handling and optional `model` are the right fixes for Foundry compatibility. A few things worth resolving before this is ready:

**Stateful agents (the core blocker)**
`previous_response_id` is silently ignored right now. Clients that rely on it will get stateless responses with no indication anything went wrong. Before removing `DO NOT MERGE`, this should either be properly handled or the API should return a clear error (e.g. `501 Not Implemented`) when `previous_response_id` is present — so callers aren't silently misled.

**`model` fallback never fires**
Since `model` is declared as `str | None = None`, the attribute always exists — so `getattr(payload, 'model', 'unknown-model')` always returns `None` when the field is omitted, never the fallback string. Fix:
```python
model_name = payload.model or 'unknown-model'
```

**Operator precedence in usage conversion**
```python
chat_response.usage.prompt_tokens or 0 if chat_response.usage else 0
```
This parses as `prompt_tokens or (0 if usage else 0)` — the guard never protects the attribute access. Add parentheses:
```python
(chat_response.usage.prompt_tokens or 0) if chat_response.usage else 0
```

**ID prefix and length inconsistency**
The PR objective says 50 characters, but the current code truncates to 64. Also `rsp_` should be `resp_` to match the OpenAI Responses API convention (e.g. `resp_67ccd2bed1ec8190...`). Worth settling on one length and documenting why.

**Missing tests for the PR's own objectives**
None of the three stated goals — optional `model`, `input_text` content type, ID length compliance — have direct test coverage. These are the easiest wins to add and will catch regressions if the conversion logic changes.

**Minor cleanup**
- `ResponsesInputItem` is defined but never referenced — can be removed
- `step_adaptor` parameter in `generate_responses_api_streaming` is unused and has a mutable default (Ruff B008/ARG001)
- README uses `NAT` acronym in a few places — guidelines require spelling out NeMo Agent Toolkit

chuenchen309:
Picking this up because it has been quiet since June and the `DO NOT MERGE` reason is still the only
thing standing between it and a review. Three findings, all checked against `develop` at `a35d30c`
rather than read off the diff.

**The feature has not been overtaken.** `git grep -n 'v1/responses\|previous_response_id\|responses_api'
origin/develop -- packages src` is empty, so nothing has landed in the meantime that makes this
redundant. Whatever happens to the branch, the gap it targets is still open.

**But the branch no longer applies to a tree that exists.** Every source and test path in it moved
during the packages restructure:

| in this PR | on `develop` today |
|---|---|
| `src/nat/data_models/api_server.py` | `packages/nvidia_nat_core/src/nat/data_models/api_server.py` |
| `src/nat/front_ends/fastapi/*` (4 files) | `packages/nvidia_nat_core/src/nat/front_ends/fastapi/*` |
| `tests/nat/front_ends/fastapi/test_openai_compatibility.py` | `packages/nvidia_nat_core/tests/nat/front_ends/fastapi/…` |

`git ls-tree origin/develop --name-only src/` returns nothing — the root `src/` is gone. So the
`DIRTY` state isn't a handful of conflicting hunks; +1461 lines have to be re-applied by hand
wherever they end up. That cost only grows, which seems worth weighing against how long the stateful
question has been open.

**Item 3 of the description is not in the code.** "Updating message ID and response ID to comply with
length requirement of 50 characters" — the IDs are minted as

```python
id: str = Field(default_factory=lambda: f"msg_{secrets.token_hex(30)}"[:64])
```

`secrets.token_hex(30)` is 60 hex characters, so with the 4-character prefix the string is exactly 64
and the `[:64]` slice never truncates anything. Same for `rsp_`, and for the two ids built in
`response_helpers.py`. The number 50 does not appear anywhere in the diff. If the Foundry limit really
is 50, `secrets.token_hex(23)` gives exactly 50 including the prefix; if the limit turned out to be 64
after all, the slice is dead code and the description is what needs correcting.

**On splitting**: items 1 and 2 are not in the same position. `model: str | None` only exists on the
new `ResponsesRequest`, so item 1 lives inside the part the February objection is about and can't be
separated from it. Item 2 can: `INPUT_TEXT`, `InputTextContent`, its slot in the `UserContent` union,
the `Message.content` serializer that maps it back to `type: "text"`, and one `isinstance` branch in
`message_handler.py` — none of that touches `/v1/responses`. I checked it would also be complete on
its own: `message_handler.py:263` is the only place on `develop` that walks `user_message.content` and
dispatches on the member type, so the single added branch covers every consumer, and nothing else
matches on `ChatContentType`. Foundry sends `input_text` on the chat-completions path too, so that
piece has value even if the Responses endpoint stays parked.

Not trying to relitigate the stateful-agents call — that one has been made twice and I have nothing to
add to it. This is just the state of the branch as of today, in case it helps decide between reviving
and closing.

*AI-assisted with agentic coding tools; the path table, the grep results and the id arithmetic are
output from running those commands against `develop` at `a35d30c`, not recalled.*

PR #1577, opened 2026-02-09
chore(llm-providers): Add env `OPENAI_BASE_URL` for `openai` ; unify llm provider configs

willkill07:
## Description
Add this so we can override `OPENAI_BASE_URL` and `OPENAI_API_KEY` when testing in nightly CI.

Also update the logic for passthrough of `OPENAI_API_KEY`

Closes

## Summary by CodeRabbit

* **New Features**
  * LLM plugins now accept API key and base URL from environment variables as fallbacks across providers.
  * Semantic Kernel integration now uses an asynchronous OpenAI client for improved async handling.

* **Refactor**
  * Unified config construction and cleanup across multiple LLM integrations for more consistent instantiation.

* **Tests**
  * Test suite updated to ensure API-key-related test setup is applied.


mnajafian-nv (approved):
Great work!

You mentioned you're open to suggestions for Semantic Kernel. Here's how I'd extend the same pattern there.

Based on my understanding the main difference is that unlike LangChain/LlamaIndex which accept base_url directly, Semantic Kernel's OpenAIChatCompletion needs a custom AsyncOpenAI client to handle non-default endpoints.

**Suggestions:** 

packages/nvidia_nat_semantic_kernel/src/nat/plugins/semantic_kernel/llm.py

Add these imports at the top:

```
import os

from typing import TypeVar

from openai import AsyncOpenAI

```

Update the openai_semantic_kernel function:

```
@register_llm_client(config_type=OpenAIModelConfig, wrapper_type=LLMFrameworkEnum.SEMANTIC_KERNEL)
async def openai_semantic_kernel(llm_config: OpenAIModelConfig, _builder: Builder):

    from semantic_kernel.connectors.ai.open_ai import OpenAIChatCompletion

    validate_no_responses_api(llm_config, LLMFrameworkEnum.SEMANTIC_KERNEL)

    # Determine base_url (config takes precedence, env var as fallback)
    base_url = llm_config.base_url
    if not base_url:
        base_url = os.getenv("OPENAI_BASE_URL")

    # Create custom AsyncOpenAI client if base_url is specified
    # This follows the official Semantic Kernel pattern for custom endpoints
    async_client = None
    if base_url:
        api_key = get_secret_value(llm_config.api_key) or os.getenv("OPENAI_API_KEY")
        async_client = AsyncOpenAI(
            api_key=api_key,
            base_url=base_url,
        )

    llm = OpenAIChatCompletion(
        ai_model_id=llm_config.model_name,
        async_client=async_client,
    )

    yield _patch_llm_based_on_config(llm, llm_config)
```

This approach:
• Matches Semantic Kernel's official pattern (https://github.com/microsoft/semantic-kernel/blob/main/python/samples/concepts/setup/chat_completion_services.py) used for DeepSeek/OpenRouter
• Uses the same fallback logic as your Strands implementation, but cleaner with a single os.getenv() call
• Non-breaking when base_url isn't set, it passes None and uses default behavior

Happy to open a PR with this if that's easier, or feel free to just copy it. Either way works!

PR #1584, opened 2026-02-10
Resolve user ID from JWT or nat-session cookie

AnuradhaKaruppiah:
## Description

User ID for per-user workflows can now come from either the nat-session cookie or the Authorization Bearer JWT. The nat-session cookie is preferred so existing behavior is unchanged; JWT is used only when the cookie is not set.
- nat-session cookie – if present, use it (unchanged for current users).
- JWT in Authorization: Bearer – if no cookie, decode the JWT payload (no verification) and use the first non-empty claim among: name, email, preferred_username, sub.

## Summary by CodeRabbit

* **New Features**
  * Per-user identification now supports JWT via the Authorization header as an alternative to cookie-based sessions.
  * MCP client CLI gained an --auth flag for connecting to protected servers.

* **Documentation**
  * Guides updated to prefer nat-session cookie with JWT as a fallback and to generalize per-user workflow behavior.
  * API reference updated to document JWT support for per-user monitoring.

* **Tests**
  * Added example scripts to verify both cookie-based and JWT-based authentication flows.


packages/nvidia_nat_core/src/nat/runtime/session.py
```diff
@@ -384,6 +387,79 @@ def _get_user_id_from_context(self) -> str | None:
...
+        # 2. Fallback: unpack JWT from Authorization header and read user identity claims
+        if auth_header_value:
+            parts = (auth_header_value or "").strip().split()
+            if len(parts) >= 2 and parts[0].lower() == "bearer":
+                payload = self._decode_jwt_payload_unverified(parts[1])
+                if payload:
+                    for claim in ("name", "email", "preferred_username", "sub"):
+                        val = payload.get(claim)
+                        if val and isinstance(val, str) and val.strip():
+                            return val.strip()
```
willkill07: Since auth headers should only have a single space, we can rely on maxsplit

Other logic can be slightly simplified
```suggestion
            parts = auth_header_value.strip().split(maxsplit=1)
            if len(parts) == 2 and parts[0].lower() == "bearer":
                if (payload := self._decode_jwt_payload_unverified(parts[1])):
                    if (val := payload.get(claim)) and isinstance(val, str) and val.strip():
                        return val.strip()
```

packages/nvidia_nat_mcp/scripts/check_ws_mcp_auth_jwt.py
```diff
@@ -0,0 +1,152 @@
...
+import json
+import sys
+import webbrowser
+
+import websockets
+
+try:
+    from authlib.jose import jwt
+except ImportError:
+    jwt = None  # type: ignore[assignment]
```
willkill07: I would raise an error here since it's fundamentally required

packages/nvidia_nat_mcp/scripts/check_ws_mcp_auth_jwt.py
```diff
@@ -0,0 +1,152 @@
...
+                if isinstance(content, dict):
+                    chunk = content.get("text") or content.get("output")
+                    if isinstance(chunk, str) and msg.get("status") == "in_progress":
+                        response_chunks.append(chunk)
+
+                if msg.get("status") == "complete":
+                    final_answer = "".join(response_chunks).strip()
+                    if final_answer:
+                        print(final_answer)
+                    return
```
willkill07: Feel free to dismiss outright, but for REPL parsing like this, Python's `match` is quite nice
```suggestion
            match msg.get("type"):
                case "system_interaction_message":
                   content = msg.get("content", {})
                   if content.get("input_type") == "oauth_consent":
                       if (url := content.get("text")):
                           webbrowser.open(url)
                   continue

                case "error_message":
                    match (content := msg.get("content", {}):
                        case dict():
                            print(f"Error: {content.get('message')}", file=sys.stderr)
                        case _:
                            print(f"Error: {content}", file=sys.stderr)
                    return

                case "system_response_message":
                    match (content := msg.get("content", {})):
                        case dict():
                            chunk = content.get("text") or content.get("output")
                            if isinstance(chunk, str) and msg.get("status") == "in_progress":
                                response_chunks.append(chunk)
                        case _:
                            if msg.get("status") == "complete":
                                if (final_answer := "".join(response_chunks).strip()):
                                    print(final_answer)
                        return
```

docs/source/extend/custom-components/custom-functions/per-user-functions.md
```diff
@@ -143,16 +143,19 @@ async def my_per_user_group(config: MyPerUserGroupConfig, builder: Builder):
 
 ### User Identification
 
-When using the FastAPI front end with `nat serve`, users are identified by the `nat-session` cookie. Each unique session ID represents a different user.
+When using the FastAPI front end with `nat serve`, the user ID for per-user workflows is derived as follows:
+
+1. **nat-session cookie** (preferred): If the request includes a `nat-session` cookie, its value is used as the user ID. Each unique session ID represents a different user.
+2. **JWT in Authorization header** (fallback): If no cookie is set, the server decodes the JWT from `Authorization: Bearer <token>` (without verification) and uses the first non-empty claim among `name`, `email`, `preferred_username`, and `sub` as the user ID.
```
willkill07, line 149: Could we have a warning admonition that states this is done without verification?

willkill07 (approved):
Approving with minor (dismissible) feedback

packages/nvidia_nat_mcp/scripts/check_ws_mcp_auth_jwt.py
AnuradhaKaruppiah: agree, addressed

docs/source/extend/custom-components/custom-functions/per-user-functions.md
AnuradhaKaruppiah: agree,  I have added the warning

packages/nvidia_nat_core/src/nat/runtime/session.py
AnuradhaKaruppiah: addressed

packages/nvidia_nat_mcp/scripts/check_ws_mcp_auth_jwt.py
AnuradhaKaruppiah: nice, definitely more crisp. thx for the feedback @willkill07

packages/nvidia_nat_core/src/nat/runtime/session.py
```diff
@@ -23,7 +23,9 @@
 from contextlib import asynccontextmanager
 from contextlib import nullcontext
 from datetime import datetime
+from http.cookies import SimpleCookie
 
+import jwt
```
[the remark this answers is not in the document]
AnuradhaKaruppiah: pyjwt is a transitive dependency, does it need to be explicit?

packages/nvidia_nat_core/src/nat/runtime/session.py
```diff
@@ -384,6 +386,74 @@ def _get_user_id_from_context(self) -> str | None:
             logger.debug(f"Could not extract user_id from context: {e}")
             return None
 
+    @staticmethod
+    def _get_auth_and_cookies_from_connection(connection: HTTPConnection) -> tuple[str | None, dict[str, str]]:
```
ericevans-nv: I can see the auth/cookie extraction logic potentially growing over time. Would it make sense to move these static methods into a separate helper module or utility class?

AnuradhaKaruppiah: agree, moved it to a separate module, For now i have kept it the module in runtime as utils seems to be generic helpers (like string parsers)

Salonijain27 (approved):

PR #1595, opened 2026-02-12
feat(agent): add token-by-token streaming to tool_calling_agent

MylesShannon:
## Description

This PR enables token-by-token streaming for the Tool Calling Agent when `stream=true` is set on chat completion requests. Previously, streaming requests returned the full response in a single Server-Sent Event; responses are now streamed incrementally as LLM tokens are produced.

**Changes:**

- **`register.py`**: Added a streaming entry point `_stream_fn` that uses LangGraph `astream(stream_mode="messages")` to consume message chunks from the graph and yield only `AIMessageChunk` content from the agent node (excluding tool-call chunks). The Tool Calling Agent is now registered with `FunctionInfo.create(single_fn=_response_fn, stream_fn=_stream_fn, ...)` so both single-call and streaming are supported.
- **`agent.py`**: The agent node now accepts `RunnableConfig`, forwards it to the underlying LLM, and uses `self.agent.astream()` instead of `ainvoke()` so the LLM streams. Chunks are accumulated into the final `AIMessage` for non-streaming paths, and config is merged so streaming callbacks propagate correctly.
- **Tests**: `test_tool_calling.py` updated so `agent_node` is called with a mock `config` argument to match the new signature.

No new user-facing documentation was added; existing docstrings and inline comments were updated to describe the streaming behavior.

## Summary by CodeRabbit

* **New Features**
  * Real-time streaming for the tool-calling workflow, delivering incremental token content via a streaming entry.

* **Refactor**
  * Agent interfaces now accept an explicit runtime/config parameter to control execution and propagate runtime settings.

* **Bug Fixes**
  * Streamed chunks are concatenated into a single response, appended to message state, and an error is raised if no output is produced.

* **Tests**
  * Tests updated to cover config-aware calls and streaming behavior.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/data_models/api_server.py
+++ b/packages/nvidia_nat_core/src/nat/data_models/api_server.py
@@ -436,7 +436,7 @@ class ChatResponseChunk(ResponseBaseModelOutput):
                                                              delta=ChoiceDelta(
                                                                  content=data,
                                                                  role=UserMessageContentRoleType.ASSISTANT),
-                                                             finish_reason="stop")
+                                                             finish_reason=None)
                                  ],
                                  created=created,
                                  model=model,
--- a/packages/nvidia_nat_core/src/nat/data_models/api_server.py
+++ b/packages/nvidia_nat_core/src/nat/data_models/api_server.py
@@ -419,7 +419,8 @@ class ChatResponseChunk(ResponseBaseModelOutput):
                     id_: str | None = None,
                     created: datetime.datetime | None = None,
                     model: str | None = None,
-                    object_: str | None = None) -> "ChatResponseChunk":
+                    object_: str | None = None,
+                    finish_reason: str | None = None) -> "ChatResponseChunk":
 
         if id_ is None:
             id_ = str(uuid.uuid4())
@@ -430,13 +431,15 @@ class ChatResponseChunk(ResponseBaseModelOutput):
         if object_ is None:
             object_ = "chat.completion.chunk"
 
+        final_finish_reason = finish_reason if finish_reason in FINISH_REASONS else None
+
         return ChatResponseChunk(id=id_,
                                  choices=[
                                      ChatResponseChunkChoice(index=0,
                                                              delta=ChoiceDelta(
                                                                  content=data,
                                                                  role=UserMessageContentRoleType.ASSISTANT),
-                                                             finish_reason=None)
+                                                             finish_reason=final_finish_reason)
                                  ],
                                  created=created,
                                  model=model,
--- a/packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/tool_calling_agent/agent.py
+++ b/packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/tool_calling_agent/agent.py
@@ -98,10 +98,13 @@ class ToolCallAgentGraph(DualNodeAgent):
             logger.debug("%s Starting the Tool Calling Agent Node", AGENT_LOG_PREFIX)
             if len(state.messages) == 0:
                 raise RuntimeError('No input received in state: "messages"')
-            response = await self.agent.ainvoke(
-                {"messages": state.messages},
-                config=self._runnable_config,
-            )
+            # Use astream so LangGraph's stream_mode="messages" can observe individual LLM tokens.
+            # Config is inherited from LangGraph's context, preserving streaming callbacks.
+            response = None
+            async for chunk in self.agent.astream({"messages": state.messages}):
+                response = chunk if response is None else response + chunk
+            if response is None:
+                raise RuntimeError('No response received from agent')
             if self.detailed_logs:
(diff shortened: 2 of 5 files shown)


packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/tool_calling_agent/agent.py
```diff
@@ -95,15 +95,20 @@ def __init__(
         self.return_direct = [tool.name for tool in return_direct] if return_direct else []
         logger.debug("%s Initialized Tool Calling Agent Graph", AGENT_LOG_PREFIX)
 
-    async def agent_node(self, state: ToolCallAgentGraphState):
+    async def agent_node(self, state: ToolCallAgentGraphState, config: RunnableConfig):
```
willkill07: None of the other nodes specify a config, so why add it to the interface here?

defaulting to the {`__pregel_runtime: DEFAULT_RUNTIME`} is fine. Otherwise your assignment on line `103` is going to change the passed configuration (which is dangerous mutating behavior)

willkill07 (changes requested):

packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/tool_calling_agent/agent.py
MylesShannon: We only need config on agent_node because this is the only node that calls the LLM (self.agent.astream). Passing the run’s RunnableConfig (including callbacks) into that call is what allows streaming events to propagate; without it we don’t get token-level streaming. The other nodes (tool_node, conditional_edge) don’t invoke the LLM, so they don’t need config.

Agreed on not mutating the passed config. I’ve updated it to build a new dict for the LLM call instead of assigning to config["configurable"]

willkill07:
/ok to test [154b826](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1595/commits/154b826974cf933a3930390933d07ba59e34ddb9)

packages/nvidia_nat_core/src/nat/data_models/api_server.py
```diff
@@ -436,7 +436,7 @@ def from_string(data: str,
                                                              delta=ChoiceDelta(
                                                                  content=data,
                                                                  role=UserMessageContentRoleType.ASSISTANT),
-                                                             finish_reason="stop")
+                                                             finish_reason=None)
```
willkill07: Could we extend the parameters for `from_string` to include an optional `finish_reason` and use this same logic:

```python
final_finish_reason = finish_reason if finish_reason in FINISH_REASONS else None
```

packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/tool_calling_agent/register.py
```diff
@@ -121,9 +122,50 @@ async def _response_fn(chat_request_or_message: ChatRequestOrMessage) -> str:
...
+        Uses graph.astream with stream_mode="messages" to yield token-level content chunks from the LLM,
+        enabling real-time SSE streaming over the OpenAI-compatible /v1/chat/completions endpoint.
+
+        Args:
+            chat_request_or_message (ChatRequestOrMessage): The input message to process
+
+        Yields:
+            str: Individual content chunks from the agent's response
+        """
+        from langchain_core.messages import AIMessageChunk
```
willkill07: Move this import to the top of `tool_calling_agent_workflow`

packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/tool_calling_agent/register.py
```diff
@@ -121,9 +122,50 @@ async def _response_fn(chat_request_or_message: ChatRequestOrMessage) -> str:
...
+            raise
+
     try:
-        yield FunctionInfo.from_fn(_response_fn, description=config.description)
+        yield FunctionInfo.create(single_fn=_response_fn, stream_fn=_stream_fn, description=config.description)
     except GeneratorExit:
         logger.exception("%s Workflow exited early!", AGENT_LOG_PREFIX)
     finally:
-        logger.debug("%s Cleaning up react_agent workflow.", AGENT_LOG_PREFIX)
+        logger.debug("%s Cleaning up tool_calling_agent workflow.", AGENT_LOG_PREFIX)
```
willkill07: You can replace all of this with:

```python
yield FunctionInfo.create(single_fn=_response_fn, stream_fn=_stream_fn, description=config.description)
```

willkill07 (changes requested):
Minor changes, but otherwise LGTM!

willkill07:
/ok to test [8869d68](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1595/commits/8869d68af60b41bc97691263c0083b816a514135)

willkill07 (approved):

PR #1601, opened 2026-02-13
Refactor latency sensitivity to use integers instead of enums.

dnandakumar-nv:
Replaces the `LatencySensitivity` enum (`LOW`/`MEDIUM`/`HIGH`) with a plain `int`, giving users full control over latency sensitivity values instead of being locked to three predefined levels. The default value is `2` (preserving current MEDIUM behavior).

### Motivation

The three-level enum was unnecessarily restrictive. Users working with custom Dynamo routing setups need to express finer-grained priority — for example, distinguishing between "slightly above normal" (3) and "critical real-time" (10). An integer value maps naturally to the downstream transport:

- **HTTP headers**: sent as `str(int)` (e.g. `"2"`, `"5"`)
- **nvext.agent_hints**: sent as `float(int)` (e.g. `2.0`, `5.0`)

No backwards compatibility shim is needed — the enum was internal and not yet part of any public/stable API.

### Changes (6 files, net -51 lines)

| File | Change |
|------|--------|
| `nat/profiler/decorators/latency.py` | Deleted `LatencySensitivity` enum and `StrEnum` import. Decorator accepts `int`, validates with `isinstance`, raises `TypeError` on misuse. |
| `nat/builder/context.py` | Stack type `list[int]`, default `[2]`. Property returns `int` via plain `max(stack)`. |
| `nat/llm/dynamo_llm.py` | Header: `str(latency_sensitivity)`. Body: `float(latency_sensitivity)`. Fallback: `2`. |
| `tests/.../test_latency.py` | Rewrote enum tests as integer validation tests (accepts int, rejects str/float/None, accepts 0/negative/large/arbitrary). All 42 tests use plain ints. |
| `tests/.../test_dynamo_llm.py` | Two assertion updates: header `"MEDIUM"` → `"2"`, agent_hints `"MEDIUM"` → `2.0`. |
| `examples/.../graph_nodes.py` | Removed enum import, replaced 7 `LatencySensitivity.HIGH` → `3`. |

### Usage

```python
from nat.profiler.decorators.latency import latency_sensitive
from nat.builder.context import Context

@latency_sensitive(3)
async def critical_llm_call():
    return await llm.generate()

@latency_sensitive(1)
def background_task():
    pass

# Read current effective sensitivity (max across the call stack)
sensitivity = Context.get().latency_sensitivity  # returns int
```

### Prior art: APIs that use integer priority

| System | Field | Type | Range | Notes |
|--------|-------|------|-------|-------|
| HTTP/2 | Stream weight | `uint8` | 1–256 | RFC 7540 §5.3 — higher weight = more resources |
| HTTP/3 | Urgency | `int` | 0–7 | RFC 9218 Extensible Priorities — lower = more urgent |
| POSIX / Linux | `nice` value | `int` | -20 to 19 | Lower = higher scheduling priority |
| Kubernetes | Pod `priority` | `int32` | unbounded | Higher = more important; preempts lower-priority pods |
| RabbitMQ | Message priority | `uint8` | 0–255 | Higher = delivered first from priority queues |
| NATS JetStream | Consumer priority | `int` | 0+ | Lower = preferred consumer for message delivery |
| Python `asyncio` | `Task` (3.12+) | — | — | PEP discussion for integer-based task priority |
| gRPC | Retry priority | `int` | impl-defined | Controls retry/hedging order |
| CUDA Streams | Stream priority | `int` | device-range | Lower numerical value = higher priority |

The common pattern is: an unbounded or wide-range integer is more flexible than a fixed enum, and the interpretation (higher-is-more-urgent vs lower-is-more-urgent) is defined by convention, not type.

### Test plan

- [x] `pytest tests/nat/profiler/decorators/test_latency.py -v` — 42/42 passed
- [x] `pytest tests/nat/llm/test_dynamo_llm.py -v -k latency` — 2/2 passed
- [x] `grep -r "LatencySensitivity" packages/nvidia_nat_core/src/` — no matches

## Summary by CodeRabbit

* **Refactor**
  * Updated latency sensitivity handling to use numeric values instead of enum-based representation. Applications specifying latency sensitivity now use integer values rather than predefined constants, maintaining consistent functionality across prediction headers and context management.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/profiler/decorators/latency.py
+++ b/packages/nvidia_nat_core/src/nat/profiler/decorators/latency.py
@@ -16,8 +16,8 @@
 Latency sensitivity decorator for marking functions with latency requirements.
 
 This module provides the @latency_sensitive decorator that allows marking functions
-with latency sensitivity levels (LOW, MEDIUM, HIGH). The sensitivity propagates through
-the call stack with priority-based merging, where higher sensitivity takes precedence.
+with integer latency sensitivity levels. The sensitivity propagates through
+the call stack with max-based merging, where higher values take precedence.
 
 Use cases:
 - LLM routing: Direct high-sensitivity requests to low-latency backends
@@ -25,17 +25,17 @@ Use cases:
 - Observability: Track which parts of workflows have strict latency requirements
 
 Example:
-    Basic usage with enum::
+    Basic usage with integers::
 
-        from nat.profiler.decorators.latency import LatencySensitivity, latency_sensitive
+        from nat.profiler.decorators.latency import latency_sensitive
 
-        @latency_sensitive(LatencySensitivity.HIGH)
+        @latency_sensitive(3)
         async def critical_llm_call():
             return await llm.generate()
 
-    Using string form::
+    Using integer values::
 
-        @latency_sensitive("low")
+        @latency_sensitive(1)
         def background_task():
             pass
 
@@ -45,7 +45,7 @@ Example:
 
         def my_function():
             sensitivity = Context.get().latency_sensitivity
-            if sensitivity == LatencySensitivity.HIGH:
+            if sensitivity >= 3:
                 # Use fast path
                 pass
 """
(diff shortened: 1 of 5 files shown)


packages/nvidia_nat_core/src/nat/profiler/decorators/latency.py
```diff
@@ -45,122 +45,53 @@ def background_task():
 
         def my_function():
             sensitivity = Context.get().latency_sensitivity
-            if sensitivity == LatencySensitivity.HIGH:
+            if sensitivity >= 3:
                 # Use fast path
                 pass
```
willkill07, line 50: This is somewhat duplicated now that we don't have enum + string

willkill07 (approved):

packages/nvidia_nat_core/src/nat/profiler/decorators/latency.py
AnuradhaKaruppiah: @dnandakumar-nv I am going to merge this PR as I need it for the eval-refactor. Can you pls reflect-on and address this comment in a separate PR as needed.

PR #1608, opened 2026-02-17
Refactor call index tracking for prefix predictions

dnandakumar-nv:
## Description
Replaced external call tracker logic with an internal per-prefix call counter for managing call indices. This simplifies dependency management by removing reliance on intermediate_step_manager and ensures increment logic remains self-contained within the class.

## Summary by CodeRabbit

* **Chores**
  * Improved internal per-prefix call tracking and concurrency handling to provide more consistent call indexing and safer concurrent behavior; no changes to public APIs or visible behaviors.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/llm/dynamo_llm.py
+++ b/packages/nvidia_nat_core/src/nat/llm/dynamo_llm.py
@@ -471,6 +471,9 @@ class _DynamoTransport(httpx.AsyncBaseTransport):
         self._prediction_lookup = prediction_lookup
         self._use_raw_values = use_raw_values
         self._disable_headers = disable_headers
+        # Per-prefix call counter so call_index advances across requests
+        # for the same prefix_id (keyed by prefix_id string).
+        self._call_counts: dict[str, int] = {}
 
     async def handle_async_request(self, request: "httpx.Request") -> "httpx.Response":
         # Get prefix ID from context (supports depth-awareness and overrides)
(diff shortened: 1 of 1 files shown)


packages/nvidia_nat_core/src/nat/llm/dynamo_llm.py
```diff
@@ -471,6 +471,9 @@ def __init__(
         self._prediction_lookup = prediction_lookup
         self._use_raw_values = use_raw_values
         self._disable_headers = disable_headers
+        # Per-prefix call counter so call_index advances across requests
+        # for the same prefix_id (keyed by prefix_id string).
+        self._call_counts: dict[str, int] = {}
```
willkill07, line 477: If we have simultaneous (concurrent + parallel) async requests, we need to also have some sort of lock when updating the value of the dictionary. Let's call that `_call_counts_lock` here. (`threading.Lock` should work)

packages/nvidia_nat_core/src/nat/llm/dynamo_llm.py
```diff
@@ -491,17 +494,13 @@ async def handle_async_request(self, request: "httpx.Request") -> "httpx.Respons
...
-                # Get call index for current parent function
-                call_index = 1
-                active_fn = ctx.active_function
-                if active_fn and active_fn.function_id != "root":
-                    tracker = get_call_tracker()
-                    call_index = tracker.counts.get(active_fn.function_id, 1)
+                # Increment per-prefix call counter to advance through trie predictions.
+                # This is self-contained — no dependency on intermediate_step_manager.
+                self._call_counts[prefix_id] = self._call_counts.get(prefix_id, 0) + 1
+                call_index = self._call_counts[prefix_id]
```
willkill07: Update logic can be simplified... let's only do a single read + write.
```suggestion
                # Increment per-prefix call counter to advance through trie predictions.
                # This is self-contained — no dependency on intermediate_step_manager.
                with self._call_counts_lock:
                    call_index = self._call_counts.get(prefix_id, 0) + 1
                    self._call_counts[prefix_id] = call_index
```

willkill07 (changes requested):
I like removing the cross-cutting dependency :)

willkill07 (approved):

PR #1609, opened 2026-02-17
Add cache pinning strategy for KV cache with TTL control

dnandakumar-nv:
## Description
Introduced `CachePinType` to manage Dynamo KV cache entry retention strategies, defaulting to ephemeral with computed TTL. Updated relevant classes, methods, and tests to support this feature, including TTL formatting and optional disabling of cache control hints.

## Summary by CodeRabbit

* **New Features**
  * Added cache pinning configuration for Dynamo LLM with a new CachePinType option (default: EPHEMERAL).
  * Automatic TTL computation and injection of cache-control hints (formatted in seconds or minutes) based on request patterns; can be disabled by setting cache pinning to None.
* **Tests**
  * Expanded coverage validating cache_pin_type handling, TTL formatting, injection behavior, and client wiring.


packages/nvidia_nat_core/src/nat/llm/dynamo_llm.py
```diff
@@ -578,6 +605,24 @@ async def handle_async_request(self, request: "httpx.Request") -> "httpx.Respons
...
 
+                    # Inject cache_control for KV cache lifetime management.
+                    # TTL = total_requests * iat_raw (ms): the estimated total
+                    # conversation duration, i.e. how long the cache entry
+                    # should be retained before auto-expiring due to inactivity.
+                    # Formatted as "<N>m" (minutes) or "<N>s" (seconds),
+                    # rounded up to the nearest whole second.
+                    if self._cache_pin_type is not None:
+                        ttl_ms = total_requests * iat_raw
+                        ttl_seconds = max(1, -(-ttl_ms // 1000))  # ceil division
```
bbednarski9, line 616: Note: could return and add a buffer (ex: +10%) if final requests in some conversations see early evictions

bbednarski9 (approved):

packages/nvidia_nat_core/src/nat/llm/dynamo_llm.py
dnandakumar-nv: Agree; I think we should punt that for after more benchmarking to tune it up or down

PR #1612, opened 2026-02-18
Add max_sensitivity for latency-based prioritization

dnandakumar-nv:
## Description
Introduced `max_sensitivity` to compute request priorities based on latency sensitivity, with clamping to prevent negative values. Updated relevant methods, tests, and documentation to ensure alignment with this enhanced prioritization logic.

Dynamo treats latency sensitivity and priority independently, so we need to supply both.

## Summary by CodeRabbit

* **New Features**
  * Added `max_sensitivity` configuration parameter for Dynamo LLM models to control request priority computation.
  * Implemented priority-based request handling: priority is calculated as `max_sensitivity - latency_sensitivity`, with automatic clamping to 0 on overflow.

* **Tests**
  * Added tests validating priority clamping behavior when latency sensitivity exceeds configured maximum.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/llm/dynamo_llm.py
+++ b/packages/nvidia_nat_core/src/nat/llm/dynamo_llm.py
@@ -114,6 +114,10 @@ _OSL_CATEGORY_TO_INT: dict[str, int] = {"LOW": 128, "MEDIUM": 512, "HIGH": 2048}
 # HIGH: 750ms (midpoint of 500-1000ms range)
 _IAT_CATEGORY_TO_INT: dict[str, int] = {"LOW": 50, "MEDIUM": 250, "HIGH": 750}
 
+# Fallback when Context is unavailable (e.g. outside a workflow run).
+# Mid-range default on the [0, max_sensitivity] scale.
+_DEFAULT_LATENCY_SENSITIVITY: int = 2
+
 
 class CachePinType(StrEnum):
     """Cache pinning strategy for KV cache entries.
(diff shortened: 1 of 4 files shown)


packages/nvidia_nat_core/src/nat/llm/dynamo_llm.py
```diff
@@ -584,12 +595,24 @@ async def handle_async_request(self, request: "httpx.Request") -> "httpx.Respons
                 body = json.loads(content.decode("utf-8", errors="replace"))
                 if isinstance(body, dict):
                     # Build agent_hints dict (int or str depending on raw mode)
+                    # Priority is the integer complement of latency_sensitivity:
+                    # lower priority value = higher priority request.
+                    priority = self._max_sensitivity - latency_sensitivity
```
bbednarski9, line 615: The code handles the underflow case (priority < 0) with a warning and clamp, which is good. However, the constraint latency_sensitivity ∈ [0, max_sensitivity] is only enforced at runtime in the transport, not at configuration time.

```
# Current — late detection of misconfiguration
if priority < 0:
    logger.warning(
        "latency_sensitivity (%s) exceeds max_sensitivity (%s); "
        "clamping priority to 0", ...
    )
    priority = 0
```

Recommendation: Add a Pydantic model_validator that checks this relationship at instantiation when latency_sensitivity is promoted to a field. If it must stay context-driven, at least add an assert/raise rather than silently clamping.

bbednarski9:
latency_sensitivity is silently read from Context with a hardcoded fallback of 2, but it is never declared as a field on DynamoModelConfig. This means:

Users cannot configure it via YAML/config files

It's invisible in documentation generated from the model schema

The default value 2 is undocumented and arbitrary

```
# Current — fragile silent fallback
try:
    ctx = Context.get()
    latency_sensitivity = ctx.latency_sensitivity
except Exception:
    latency_sensitivity = 2
```

Recommendation: Either add it as an explicit field (even if read-only/computed) or document via a constant with a clear rationale:

```
_DEFAULT_LATENCY_SENSITIVITY: int = 2  # Mid-range default; range [0, max_sensitivity]

latency_sensitivity: int = Field(
    default=_DEFAULT_LATENCY_SENSITIVITY,
    ge=0,
    description="Latency sensitivity for request priority. "
                "Priority = max_sensitivity - latency_sensitivity. "
                "Override via Context.latency_sensitivity at runtime.",
)
```

bbednarski9:
The two validators are asymmetric. _coerce_prefix_osl has full type checking with explicit raise TypeError for non-int/non-str inputs. _coerce_prefix_iat silently falls through for unexpected types, relying on Pydantic's own coercion downstream.

```
# prefix_osl: raises TypeError for bad types ✓
raise TypeError(f"prefix_osl must be int or str, got {type(v)}")

# prefix_iat: falls through silently for unknown types ✗
def _coerce_prefix_iat(cls, v: object) -> object:
    if isinstance(v, str) and v.upper() in _IAT_CATEGORY_TO_INT:
        return _IAT_CATEGORY_TO_INT[v.upper()]
    return v  # floats, None, etc. pass through here
```

Recommendation: Make both validators consistent — either both raise on bad types, or both delegate to Pydantic. Given latency_sensitivity also uses integer arithmetic, this style inconsistency is relevant to any future latency_sensitivity field validator.

bbednarski9:
Double check the ctx.latency_sensitivity typing in this module. You already acknbowledge ambiguity with the ""latency_sensitivity": float(latency_sensitivity)," but this could fail silently at runtime

bbednarski9 (changes requested):
See comments above.

bbednarski9 (approved):

PR #1614, opened 2026-02-19
Feature/optimizer module and registry

afourniernv:
WIP DRAFT PR

## Description

Extracts the optimizer into a standalone `nvidia-nat-optimizer` package, refactors the GA as an evolutionary-base implementation, and introduces a pluggable registry for optimizer strategies—enabling third-party optimizers without modifying core code.

### Package extraction

- New `packages/nvidia_nat_optimizer` package with its own `pyproject.toml` and minimal deps. Installable as `nvidia-nat-optimizer` (standalone) or included by default via `nvidia-nat`.
- Parameter and prompt optimizer code moved out of `nat_core` profiler into `nat.optimizer`. Core `nat optimize` CLI imports optimizer resiliently and suggests installation when missing.
- Examples, LangChain prompt-optimizer hooks, and docs updated for the new package layout.

### Evolutionary base and GA refactor

- Introduced `evolutionary_base.py` with a shared evaluation loop and template method; subclasses implement fitness computation and persistence.
- Replaced monolithic `prompt_optimizer` with `ga_prompt_optimizer` as the first evolutionary implementation. Oracle feedback moves out of the shared config into `model_extra` and GA-specific config.

### Pluggable optimizer registry

- New `OptimizerStrategyBaseConfig` base and registry in `type_registry` (`register_optimizer`, `get_optimizer`). `NumericOptimizationConfig` and `PromptGAOptimizationConfig` extend it with `name="numeric"` and `name="ga"`.
- `register_optimizer` decorator in `register_workflow.py` (mirrors `register_evaluator`). Optimizer runtime dispatches via `get_optimizer(type(config))` instead of direct imports.
- `nat.optimizer.register` registers built-in strategies when the package loads. Existing YAML configs remain valid; strategy selection uses the inferred config type.

### Note on diff size

~5k additions are mostly: **uv.lock** for the new package (~3.6k lines) and **one new file** `ga_prompt_optimizer.py` (~526 lines). Moved files (e.g. `oracle_feedback.py`, `update_helpers.py`) show as "renamed without changes" so blame is preserved.


packages/nvidia_nat_core/src/nat/data_models/optimizer.py
```diff
@@ -49,10 +60,15 @@ class NumericOptimizationConfig(BaseModel):
     )
 
 
-class PromptGAOptimizationConfig(BaseModel):
+class PromptGAOptimizationConfig(OptimizerStrategyBaseConfig, name="ga"):
```
dnandakumar-nv, line 63: I feel like this should just be `PromptOptimizationConfig` with the genetic algorithm being a registered variant.

packages/nvidia_nat_core/src/nat/data_models/optimizer.py
```diff
@@ -49,10 +60,15 @@ class NumericOptimizationConfig(BaseModel):
...
     """
     Configuration for prompt optimization using a Genetic Algorithm.
+    Oracle feedback and other implementation-specific options are not part of this
```
dnandakumar-nv, line 66: The GA-specific registered variant of the `PromptOptimization` config can also have oracle feedback. Passing it in as model extra limits the amount of type checking and validation we can do, as well as the discoverability of available fields.

packages/nvidia_nat_core/tests/nat/data_models/test_optimizer_oracle_feedback.py
dnandakumar-nv, line 1: Theses tests will need to change once we pull oracle feedback and associated config items back out of model extra

packages/nvidia_nat_config_optimizer/src/nat/optimizer/ga_config.py
```diff
@@ -0,0 +1,79 @@
...
+
+from typing import Literal
+
+from pydantic import Field
+
+from nat.data_models.optimizer import BaseOptimizerConfig
+from nat.data_models.optimizer import PromptGAOptimizationConfig
+
+
+class GAOptimizerConfig(BaseOptimizerConfig):
```
dnandakumar-nv, line 33: Why do we have `PromptGAOptimizationConfig` and also this?

packages/nvidia_nat_config_optimizer/src/nat/optimizer/evolutionary_base.py
```diff
@@ -0,0 +1,124 @@
...
+from nat.data_models.optimizable import SearchSpace
+from nat.data_models.optimizer import OptimizerConfig
+from nat.data_models.optimizer import OptimizerRunConfig
+from nat.eval.evaluate import EvaluationRun
+from nat.eval.evaluate import EvaluationRunConfig
+from nat.optimizer.ga_individual import Individual
+from nat.optimizer.update_helpers import apply_suggestions
+
+
+class BaseEvolutionaryPromptOptimizer(ABC):
```
dnandakumar-nv, line 34: Seems like we need one more level of interface above this. Something like `BasePromptOptimizer` which defines methods that all prompt optimizers must implement. And then the `GAPromptOptimizer` is a subclass of that. `BaseEvolutionaryPromptOptimizer` seems unecessary unless we expect to have many evolutionary algorithms AND other algorithms we want to support.

packages/nvidia_nat_config_optimizer/src/nat/optimizer/evolutionary_base.py
dnandakumar-nv, line 1: Another suggestion/note on the directory structure. Perhaps the dilenation would be clearer if we have two directories under `packages/nvidia_nat_optimizer/src/nat/optimizer`. One for parameters and one for prompts

packages/nvidia_nat_optimizer/src/nat/optimizer/parameter_optimizer.py
dnandakumar-nv, line 38: I feel like we also need an ABC for parameter optimization so that developers can also register new ways to do parameter optimization but still enforce a standard interface.

dnandakumar-nv (changes requested):

afourniernv:
Superseded by:

Dhruv comments have been addressed in the latest PR. 

https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1637 
#1637

PR #1620, opened 2026-02-20
Resolve user ID from configurable HTTP header

ericevans-nv:
## Description

Adds `auth_token_header` config field to `MCPClientConfig` that specifies an HTTP header containing a JWT for user identification. When set, the user ID is extracted from JWT claims and used as the per-user session key, as an alternative to the `nat-session` cookie. Authentication with the MCP server still proceeds through the normal OAuth flow.

This supports deployments where an upstream service sends the user's identity as a JWT in an HTTP header (e.g., `Authorization`) rather than a session cookie, such as HTTP streaming scenarios.

## Summary by CodeRabbit

* **New Features**
  * Support for HTTP header–based authentication using JWTs to improve user identification and session routing
  * New configuration option to specify which HTTP header contains authentication tokens for routing decisions

* **Tests**
  * Added test coverage for JWT-based authentication header resolution

### DIFF
--- a/packages/nvidia_nat_mcp/src/nat/plugins/mcp/client/client_config.py
+++ b/packages/nvidia_nat_mcp/src/nat/plugins/mcp/client/client_config.py
@@ -142,6 +142,8 @@ class MCPClientConfig(MCPClientBaseConfig, name="mcp_client"):
     """
     session_aware_tools: bool = Field(default=True,
                                       description="Session-aware tools are created if True. Defaults to True.")
+    auth_token_header: str | None = Field(
+        default=None, description="HTTP header name containing a token used for user identification.")
     max_sessions: int = Field(default=100,
                               ge=1,
                               description="Maximum number of concurrent session clients. Defaults to 100.")
--- a/packages/nvidia_nat_mcp/src/nat/plugins/mcp/client/client_impl.py
+++ b/packages/nvidia_nat_mcp/src/nat/plugins/mcp/client/client_impl.py
@@ -253,6 +253,32 @@ class MCPFunctionGroup(FunctionGroup):
         except Exception:
             return None
 
+    def _resolve_user_id_from_auth_header(self) -> str | None:
+        """Extract a JWT from the configured HTTP header and resolve the user ID from its claims."""
+        if not self._client_config or not self._client_config.auth_token_header:
+            return None
+
+        from nat.builder.context import Context as _Ctx
+        from nat.runtime.connection_auth import resolve_user_id
+
+        headers = getattr(_Ctx.get().metadata, "headers", None)
+        if headers is None:
+            return None
+
+        user_auth_header: str | None = headers.get(self._client_config.auth_token_header)
+        if not user_auth_header:
+            return None
+
+        user_id: str | None = resolve_user_id(user_auth_header, {})
+        if not user_id:
+            logger.warning("Could not resolve user ID from header '%s'", self._client_config.auth_token_header)
+            return None
+
+        logger.debug("Resolved user_id '%s' from header '%s'",
+                     truncate_session_id(user_id),
+                     self._client_config.auth_token_header)
+        return user_id
+
     async def cleanup_sessions(self, max_age: timedelta | None = None) -> int:
         """
         Manually trigger cleanup of inactive sessions.
@@ -516,7 +542,10 @@ def mcp_session_tool_function(tool, function_group: MCPFunctionGroup):
         """Response function for the session-aware tool."""
         try:
             # Route to the appropriate session client
-            session_id = function_group._get_session_id_from_context()
+            if (function_group._client_config and function_group._client_config.auth_token_header):
+                session_id = function_group._resolve_user_id_from_auth_header()
+            else:
+                session_id = function_group._get_session_id_from_context()
 
             # If no session is available and default-user fallback is disabled, deny the call
             if function_group._shared_auth_provider and session_id is None:
--- a/packages/nvidia_nat_mcp/tests/client/test_mcp_client_impl.py
+++ b/packages/nvidia_nat_mcp/tests/client/test_mcp_client_impl.py
@@ -18,6 +18,7 @@ from typing import Any
 from unittest.mock import MagicMock
 from unittest.mock import patch
 
+import jwt
 import pytest
 from pydantic import BaseModel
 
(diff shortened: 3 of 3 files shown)


willkill07 (changes requested):
Documentation should be added/updated with a clear security warning. This is an attack vector that has to be well documented.

packages/nvidia_nat_mcp/src/nat/plugins/mcp/client/client_impl.py
```diff
@@ -253,6 +253,32 @@ def _get_session_id_from_context(self) -> str | None:
...
+
+        user_id: str | None = resolve_user_id(user_auth_header, {})
+        if not user_id:
+            logger.warning("Could not resolve user ID from header '%s'", self._client_config.auth_token_header)
+            return None
+
+        logger.debug("Resolved user_id '%s' from header '%s'",
+                     truncate_session_id(user_id),
+                     self._client_config.auth_token_header)
+        return user_id
```
mnajafian-nv, line 280:
```suggestion
    def _resolve_user_id_from_auth_header(self) -> str | None:
        """Extract a JWT from the configured HTTP header and resolve the user ID from its claims."""
        if not self._client_config or not self._client_config.auth_token_header:
            return None

        try:
            from nat.builder.context import Context as _Ctx
            from nat.runtime.connection_auth import resolve_user_id

            headers = getattr(_Ctx.get().metadata, "headers", None)
            if headers is None:
                return None

            user_auth_header: str | None = headers.get(self._client_config.auth_token_header)
            if not user_auth_header:
                return None

            user_id: str | None = resolve_user_id(user_auth_header, {})
            if not user_id:
                logger.warning("Could not resolve user ID from header '%s'", self._client_config.auth_token_header)
                return None

            logger.debug("Resolved user_id '%s' from header '%s'",
                         truncate_session_id(user_id),
                         self._client_config.auth_token_header)
            return user_id
        except Exception:
            logger.debug("Failed to resolve user ID from auth header", exc_info=True)
            return None
```

mnajafian-nv: I suggest, wrapping this block in try/except Exception: return None so we match _get_session_id_from_context and don’t leak exceptions (and internal errors) to the caller via _response_fn. 
Optional: log at debug with exc_info=True in the except.

packages/nvidia_nat_mcp/tests/client/test_mcp_client_impl.py
```diff
@@ -174,6 +175,22 @@ def _make_group(server_cfg=None, client_cfg=None):
...
+    mock_context_cls.get.return_value = ctx
+
+    server_cfg = MCPServerConfig(transport="streamable-http", url="http://localhost:9901/mcp")
+    client_cfg = MCPClientConfig(server=server_cfg, auth_token_header="Authorization")
+    group = MCPFunctionGroup(config=client_cfg)
+    group._client_config = client_cfg
+
+    assert group._resolve_user_id_from_auth_header() == "Alice"
+
+
```
mnajafian-nv, line 193:
```suggestion
@patch("nat.builder.context.Context")
def test_resolve_user_id_from_auth_header_missing_header_returns_none(mock_context_cls):
    """When auth_token_header is set but the header is absent, returns None."""
    ctx = MagicMock()
    ctx.metadata.headers = {}
    mock_context_cls.get.return_value = ctx

    server_cfg = MCPServerConfig(transport="streamable-http", url="http://localhost:9901/mcp")
    client_cfg = MCPClientConfig(server=server_cfg, auth_token_header="Authorization")
    group = MCPFunctionGroup(config=client_cfg)
    group._client_config = client_cfg

    assert group._resolve_user_id_from_auth_header() is None


@patch("nat.builder.context.Context")
def test_resolve_user_id_from_auth_header_context_raises_returns_none(mock_context_cls):
    """When Context.get() raises, returns None instead of propagating."""
    mock_context_cls.get.side_effect = RuntimeError("No context")

    server_cfg = MCPServerConfig(transport="streamable-http", url="http://localhost:9901/mcp")
    client_cfg = MCPClientConfig(server=server_cfg, auth_token_header="Authorization")
    group = MCPFunctionGroup(config=client_cfg)
    group._client_config = client_cfg

    assert group._resolve_user_id_from_auth_header() is None


@patch("nat.runtime.connection_auth.resolve_user_id")
@patch("nat.builder.context.Context")
def test_resolve_user_id_from_auth_header_malformed_jwt_returns_none(mock_context_cls, mock_resolve_user_id):
    """When the header value is invalid, returns None (e.g. decode error or no user)."""
    mock_resolve_user_id.return_value = None

    ctx = MagicMock()
    ctx.metadata.headers = {"Authorization": "Bearer not-a-valid-jwt"}
    mock_context_cls.get.return_value = ctx

    server_cfg = MCPServerConfig(transport="streamable-http", url="http://localhost:9901/mcp")
    client_cfg = MCPClientConfig(server=server_cfg, auth_token_header="Authorization")
    group = MCPFunctionGroup(config=client_cfg)
    group._client_config = client_cfg

    assert group._resolve_user_id_from_auth_header() is None
```

mnajafian-nv: I suggest we add tests for: (1) header configured but missing → None, (2) Context.get() raises → None, (3) invalid token / resolve_user_id returns None or raises → None. Patch Context at nat.builder.context.Context; for (3) patch resolve_user_id at nat.runtime.connection_auth.resolve_user_id so the test is self-contained.

packages/nvidia_nat_mcp/src/nat/plugins/mcp/client/client_config.py
```diff
@@ -142,6 +142,8 @@ class MCPClientConfig(MCPClientBaseConfig, name="mcp_client"):
     """
     session_aware_tools: bool = Field(default=True,
                                       description="Session-aware tools are created if True. Defaults to True.")
+    auth_token_header: str | None = Field(
+        default=None, description="HTTP header name containing a token used for user identification.")
```
mnajafian-nv, line 146:
```suggestion
    auth_token_header: str | None = Field(
        default=None,
        description="HTTP header name containing a token used for user identification. "
        "When set, the header value is used to resolve the session user ID (e.g. from JWT claims). "
        "Only set this when the header is set by a trusted upstream (e.g. gateway); no verification is performed here.")
```

mnajafian-nv (approved):
Conditional approval upon your review of Will's comments and my inline feedbacks.

packages/nvidia_nat_mcp/src/nat/plugins/mcp/client/client_config.py
```diff
@@ -142,6 +142,8 @@ class MCPClientConfig(MCPClientBaseConfig, name="mcp_client"):
     """
     session_aware_tools: bool = Field(default=True,
                                       description="Session-aware tools are created if True. Defaults to True.")
+    auth_token_header: str | None = Field(
```
AnuradhaKaruppiah, line 145: @ericevans-nv Thanks for this change. I think we can simplify/unify user-id resolution by first using Context.get().user_id (already populated during request/session handling) and only falling back to header parsing when it’s missing.

AnuradhaKaruppiah: I missed that this PR touches only shared MCPFunctionGroup. per_user_mcp_client_function_group already uses Context.get().user_id, which is the right source of truth.
I’d suggest aligning shared MCP routing to the same pattern and removing auth_token_header from MCP config to avoid duplicate JWT parsing logic in the plugin.

ericevans-nv: The auth_token_header config serves a different purpose: it allows deployers to specify a custom header name containing the JWT, since the UI does not send nat-session cookies over HTTP and the header name is only known to the deployer. Context.user_id is populated exclusively from the Authorization header or the nat-session cookie, so it won't have the correct value when a non-standard header is used. That said, I agree we should align the pattern, I'll update the PR so the custom header resolution works for both the shared MCPFunctionGroup and per_user_mcp_client_function_group.

ericevans-nv:
Closing this PR. After discussion with the team, HTTP headers will always send JWTs in the standard Authorization header, which the session layer already handles. The custom header support isn't needed at this time. If a non-standard header requirement comes up in the future, we can revisit.

PR #1622, opened 2026-02-20
Allow running `pytest` from project root

dagardner-nv:
## Description
* Rename `pytest.ini` to `.pytest.ini`, per [pytest docs](https://docs.pytest.org/en/stable/reference/customize.html#pytest-ini) these files are treated the same.
* Specify `adopts= --ignore=examples/ --ignore=packages/nvidia_nat_openpipe_art` in `.pytest.ini` so that when someone runs `pytest` from the root it will run as many tests as it is able to.
* Rename some tests to avoid name collisions.

This doesn't break `run_tests.py` since the `--ignore` flag is ignored if the path is explicitly specified such that running:
```bash
pytest --ignore=packages/nvidia_nat_core/ packages/nvidia_nat_core
``` 
runs the tests for `nvidia_nat_core`

## Summary by CodeRabbit

* **Chores**
  * Adjusted test configuration to exclude example and two package directories from normal test runs; added inline guidance on running tests in those locations; test discovery and markers unchanged.
  * Updated project metadata to add and reorder an optional dependency entry in the "most" set, causing that package to appear there as well.


.pytest.ini
```diff
@@ -1,4 +1,5 @@
 [pytest]
+addopts = --ignore=examples/ --ignore=packages/nvidia_nat_openpipe_art
```
willkill07: This list should also match opinionated exclusions from `--extra most`

willkill07 (approved):
`./ci/scripts/run_tests.py` also does isolated testing for each package independent of any installed environment

willkill07 (approved):

PR #1626, opened 2026-02-21
fix: Pass request_timeout through to OpenAI/Azure LLM clients

### LINKED ISSUE #1617: LLM HTTP timeout is effectively disabled for OpenAI/Azure providers
### Version

1.5.0a20260216

Source

### Describe the bug.

# LLM HTTP timeout is effectively disabled for OpenAI/Azure providers

## Summary

The `create_metadata_injection_client()` in `nat/llm/utils/hooks.py` accepts a `timeout` parameter, but the timeout is never applied to actual LLM requests. This is because LangChain's `ChatOpenAI` passes `timeout=None` to the OpenAI SDK, which overrides the httpx client's timeout setting.

## Affected Providers

- `OpenAIModelConfig` (`_type: openai`)
- `AzureOpenAIModelConfig` (`_type: azure_openai`)

## Root Cause

The issue spans three layers:

### 1. NAT creates an httpx client with a timeout ([[`nat/plugins/langchain/llm.py:210`](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/c4b19b52e9ffbbce11c0077bbc60ffadf980bbc4/packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py#L210))

```python
http_async_client: httpx.AsyncClient = create_metadata_injection_client()
# httpx client has timeout=600s (default)

client = ChatOpenAI(
                http_async_client=http_async_client,  # type: ignore[call-arg]
                stream_usage=True,
                use_responses_api=True,  # type: ignore[call-arg]
                use_previous_response_id=True,  # type: ignore[call-arg]
                **config_dict)
```

### 2. LangChain `ChatOpenAI` passes `timeout=None` to the OpenAI SDK (`langchain_openai/chat_models/base.py:953-956`)

```python
client_params: dict = {
    ...
    "timeout": self.request_timeout,  # self.request_timeout defaults to None
    ...
}
```

NAT does not set `request_timeout` or `timeout` on the `ChatOpenAI` instance, so it remains `None`.

### 3. The OpenAI SDK treats `None` as an explicitly provided value (`openai/_base_client.py:1456-1467`)

```python
def __init__(self, ..., timeout: float | Timeout | None | NotGiven = not_given, http_client=None):
    if not is_given(timeout):
        # Only enters this branch if timeout is NotGiven
        if http_client and http_client.timeout != HTTPX_DEFAULT_TIMEOUT:
            timeout = http_client.timeout  # Would pick up httpx client's timeout
        else:
            timeout = DEFAULT_TIMEOUT
```

`is_given()` checks `not isinstance(obj, NotGiven)`. Since `None` is not `NotGiven`, the SDK considers `timeout=None` as explicitly provided and skips the httpx client timeout detection. The SDK then uses `timeout=None` for every request via `build_request(timeout=self.timeout)`, which means **no timeout at all**.

### Flow diagram

```
NAT creates httpx.AsyncClient(timeout=600s)
  → ChatOpenAI(http_async_client=..., timeout=None)  # None from unset request_timeout
    → AsyncOpenAI(http_client=..., timeout=None)      # None is "given", skips httpx detection
      → self.timeout = None                           # No timeout on any request
        → build_request(timeout=None)                 # Every request has no timeout
```

## Impact

- LLM API calls have **no timeout**, regardless of what is configured on the httpx client.
- The `timeout` parameter on `create_metadata_injection_client()` has no effect for `openai` and `azure_openai` providers.
- Long-running or hanging LLM requests will block indefinitely.

## Suggested Fix

In [nat/plugins/langchain/llm.py](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/c4b19b52e9ffbbce11c0077bbc60ffadf980bbc4/packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py), add a `request_timeout` field to `OpenAIModelConfig` and pass it through:

```python

# In openai_langchain():
http_async_client = create_metadata_injection_client(timeout=llm_config.request_timeout)

client = ChatOpenAI(
    http_async_client=http_async_client,
    timeout=llm_config.request_timeout,
    stream_usage=True,
    **config_dict,
)
```

The same fix should be applied to `azure_openai_langchain()`.

## Environment

- `nvidia_nat==1.5.0a20260216`
- `langchain-openai` (latest as of 2026-02)
- `openai` Python SDK

## Reproducer

```python
from langchain_openai import ChatOpenAI
import httpx

# Default: no httpx client passed
c = ChatOpenAI(api_key='test', model='test')
print(f'request_timeout: {c.request_timeout}')
print(f'root_async_client.timeout: {c.root_async_client.timeout}')

# With explicit httpx timeout=30s — still ignored
http_async = httpx.AsyncClient(timeout=httpx.Timeout(30.0))
c2 = ChatOpenAI(api_key='test', model='test', http_async_client=http_async)
print(f'request_timeout: {c2.request_timeout}')
print(f'root_async_client.timeout: {c2.root_async_client.timeout}')
```

Output:

```
request_timeout: None
root_async_client.timeout: None

request_timeout: None
root_async_client.timeout: None
```

### Relevant log output

</pre></details>

### Other/Misc.

_No response_

### Code of Conduct

bledden:
## Summary

Fixes #1617

I noticed that the `create_metadata_injection_client()` httpx timeout was getting silently overridden by LangChain/OpenAI SDK behavior. When `ChatOpenAI` passes `timeout=None` (its default for `request_timeout`), the OpenAI SDK treats that as an explicit "no timeout" rather than falling back to the httpx client's timeout. This means LLM requests can hang indefinitely regardless of what timeout is configured on the httpx client.

The Dynamo provider already handles this correctly with its own `request_timeout` field, so I followed the same pattern for OpenAI and Azure OpenAI:

- Added `request_timeout` field to `OpenAIModelConfig` and `AzureOpenAIModelConfig` (defaults to `None` to preserve current behavior)
- When set, the timeout is passed to both `create_metadata_injection_client()` and the `ChatOpenAI`/`AzureChatOpenAI` constructor via LangChain's `request_timeout` parameter
- Updated the LLM docs to document the new field

I validated the fix locally by confirming:
1. Without `request_timeout` set, behavior is unchanged (SDK defaults apply)
2. With `request_timeout` set, the timeout is properly enforced at both the httpx and SDK layers

## Test plan

- [ ] Verify `request_timeout` is properly passed through to the OpenAI SDK for both providers
- [ ] Verify default behavior is preserved when `request_timeout` is not set
- [ ] Existing tests pass

## Summary by CodeRabbit

* **New Features**
  * Added a configurable request_timeout for OpenAI and Azure OpenAI providers; supported integrations now propagate and honor this HTTP request timeout when set, improving control over LLM call latency and failures.

* **Documentation**
  * Updated provider documentation to describe the new timeout setting and how to configure and use it.

### DIFF
--- a/packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
+++ b/packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
@@ -153,13 +153,11 @@ async def azure_openai_langchain(llm_config: AzureOpenAIModelConfig, _builder: B
     http_async_client: httpx.AsyncClient = create_metadata_injection_client(**client_kwargs)
 
     config_dict = llm_config.model_dump(
-        exclude={"type", "thinking", "api_type", "api_version", "request_timeout"},
+        exclude={"type", "thinking", "api_type", "api_version"},
         by_alias=True,
         exclude_none=True,
         exclude_unset=True,
     )
-    if llm_config.request_timeout is not None:
-        config_dict["request_timeout"] = llm_config.request_timeout
 
     try:
         client = AzureChatOpenAI(
@@ -207,13 +205,11 @@ async def openai_langchain(llm_config: OpenAIModelConfig, _builder: Builder):
     http_async_client: httpx.AsyncClient = create_metadata_injection_client(**client_kwargs)
 
     config_dict = llm_config.model_dump(
-        exclude={"type", "thinking", "api_type", "api_key", "base_url", "request_timeout"},
+        exclude={"type", "thinking", "api_type", "api_key", "base_url"},
         by_alias=True,
         exclude_none=True,
         exclude_unset=True,
     )
-    if llm_config.request_timeout is not None:
-        config_dict["request_timeout"] = llm_config.request_timeout
     if (api_key := get_secret_value(llm_config.api_key) or os.getenv("OPENAI_API_KEY")):
         config_dict["api_key"] = api_key
     if (base_url := llm_config.base_url or os.getenv("OPENAI_BASE_URL")):
--- a/packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
+++ b/packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
@@ -152,18 +152,16 @@ async def azure_openai_langchain(llm_config: AzureOpenAIModelConfig, _builder: B
         client_kwargs["timeout"] = llm_config.request_timeout
     http_async_client: httpx.AsyncClient = create_metadata_injection_client(**client_kwargs)
 
-    config_dict = llm_config.model_dump(
-        exclude={"type", "thinking", "api_type", "api_version", "request_timeout"},
-        by_alias=True,
-        exclude_none=True,
-        exclude_unset=True,
-    )
-
     try:
         client = AzureChatOpenAI(
             http_async_client=http_async_client,  # type: ignore[call-arg]
             api_version=llm_config.api_version,  # type: ignore[call-arg]
-            **config_dict,
+            **llm_config.model_dump(
+                exclude={"type", "thinking", "api_type", "api_version"},
+                by_alias=True,
+                exclude_none=True,
+                exclude_unset=True,
+            ),
         )
         if "http_async_client" in client.model_kwargs:
             del client.model_kwargs["http_async_client"]
@@ -205,7 +203,7 @@ async def openai_langchain(llm_config: OpenAIModelConfig, _builder: Builder):
     http_async_client: httpx.AsyncClient = create_metadata_injection_client(**client_kwargs)
 
     config_dict = llm_config.model_dump(
-        exclude={"type", "thinking", "api_type", "api_key", "base_url", "request_timeout"},
+        exclude={"type", "thinking", "api_type", "api_key", "base_url"},
         by_alias=True,
         exclude_none=True,
         exclude_unset=True,
--- a/packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
+++ b/packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
@@ -153,7 +153,7 @@ async def azure_openai_langchain(llm_config: AzureOpenAIModelConfig, _builder: B
     http_async_client: httpx.AsyncClient = create_metadata_injection_client(**client_kwargs)
 
     config_dict = llm_config.model_dump(
-        exclude={"type", "thinking", "api_type", "api_version"},
+        exclude={"type", "thinking", "api_type", "api_version", "request_timeout"},
         by_alias=True,
         exclude_none=True,
         exclude_unset=True,
@@ -205,7 +205,7 @@ async def openai_langchain(llm_config: OpenAIModelConfig, _builder: Builder):
     http_async_client: httpx.AsyncClient = create_metadata_injection_client(**client_kwargs)
 
     config_dict = llm_config.model_dump(
-        exclude={"type", "thinking", "api_type", "api_key", "base_url"},
+        exclude={"type", "thinking", "api_type", "api_key", "base_url", "request_timeout"},
         by_alias=True,
         exclude_none=True,
         exclude_unset=True,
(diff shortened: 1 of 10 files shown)


bledden:
## Validation

Wrote a test script to reproduce and confirm the fix:

**Reproduction (confirmed the bug):**
```python
http_async = httpx.AsyncClient(timeout=httpx.Timeout(30.0))
client = ChatOpenAI(api_key='test', model='test', http_async_client=http_async)
# Bug: client.root_async_client.timeout is None despite httpx having 30s timeout
```

The OpenAI SDK treats `timeout=None` (LangChain's default for `request_timeout`) as an explicit override, so the httpx timeout gets silently ignored.

**Fix validation:**
```python
http_async = httpx.AsyncClient(timeout=httpx.Timeout(30.0))
client = ChatOpenAI(api_key='test', model='test', http_async_client=http_async, request_timeout=30.0)
# Fixed: timeout is properly set at both httpx and SDK layers
```

Same pattern verified for `AzureChatOpenAI`. Default behavior (no `request_timeout` set) is unchanged.

packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
```diff
@@ -147,16 +147,25 @@ async def azure_openai_langchain(llm_config: AzureOpenAIModelConfig, _builder: B
...
+    http_async_client: httpx.AsyncClient = create_metadata_injection_client(**client_kwargs)
+
+    config_dict = llm_config.model_dump(
+        exclude={"type", "thinking", "api_type", "api_version", "request_timeout"},
+        by_alias=True,
+        exclude_none=True,
+        exclude_unset=True,
+    )
+    if llm_config.request_timeout is not None:
+        config_dict["request_timeout"] = llm_config.request_timeout
```
willkill07: Why exclude `request_timeout` here just to conditionally include immediately below? `exclude=None` is specified in the `model_dump()`

packages/nvidia_nat_langchain/src/nat/plugins/langchain/llm.py
```diff
@@ -192,14 +201,19 @@ async def openai_langchain(llm_config: OpenAIModelConfig, _builder: Builder):
...
 
     config_dict = llm_config.model_dump(
-        exclude={"type", "thinking", "api_type", "api_key", "base_url"},
+        exclude={"type", "thinking", "api_type", "api_key", "base_url", "request_timeout"},
         by_alias=True,
         exclude_none=True,
         exclude_unset=True,
     )
+    if llm_config.request_timeout is not None:
+        config_dict["request_timeout"] = llm_config.request_timeout
```
willkill07: Same issue down here.

willkill07 (changes requested):

willkill07 (changes requested):
Sorry for the extra churn here.

We support more than langchain with the OpenAI and Azure OpenAI LLM types -- adding it to the config requires ensuring no breakage with other frameworks or, more preferably, add timeout support for other frameworks using OpenAI LLM configs.

bledden:
Good call — I went through every framework integration that uses `OpenAIModelConfig` / `AzureOpenAIModelConfig` and made two changes for each:

1. **Excluded `request_timeout` from `model_dump()`** so it never leaks into client constructors that don't expect it
2. **Mapped it to each framework's native timeout parameter** where supported

| Framework | Exclusion | Timeout support | Parameter mapped to |
|---|---|---|---|
| langchain | `request_timeout` added to exclude | Already handled via httpx client | `httpx.AsyncClient(timeout=...)` |
| llama_index | `request_timeout` added to exclude | Added | `timeout` |
| crewai | `request_timeout` added to exclude | Added | `timeout` (LiteLLM) |
| strands | `request_timeout` added to exclude | Added | `client_args["timeout"]` (AsyncOpenAI) |
| adk | `request_timeout` added to exclude | Added | `timeout` (LiteLLM) |
| agno | `request_timeout` added to exclude | Added | `timeout` |
| autogen | `request_timeout` added to exclude | Not added | N/A (strict constructor validation) |
| semantic_kernel | Not needed (manual construction, no `model_dump`) | Not added | N/A |

The key invariant is that `request_timeout` is excluded from every `model_dump()` call involving these configs, so even if a user sets it, it won't break any framework. Where we can map it to a native timeout, we do.

bledden:
@willkill07 Good catch on the langchain file — fixed in 147580b. Removed `request_timeout` from the exclude set so it passes through to LangChain natively (since `BaseChatOpenAI` accepts it directly). The httpx client timeout is still set separately for the transport layer.

willkill07 (approved):

PR #1627, opened 2026-02-21
fix: Skip output directory cleanup when --skip_workflow is set

### LINKED ISSUE #1587: nat eval --skip_workflow can delete existing trajectories
### Version

1.4

PyPi

### Describe the bug.

When running `nat eval --config_file config.yml --skip_workflow` without 
`--dataset workflow_output.json`, the already existing `dataset workflow_output.json` will be overwritten and all existing trajectories will be deleted.

- IMO one could improve this behavior by some more restrictive rules, preventing accidental overwrite of existing data, e.g. adding an `--overwrite_existing` flag. This feature might be useful in general, as nat seems to be very generous with deleting existing experiment data.
- `workflow_output.json` is already defined in `config.yml`, so I'd expect `--skip_workflow` to check if  `workflow_output.json` already contains trajectories.

```shell
See commands above, should work with any nat agent that has some evaluator.
```

### Relevant log output

_No response_

### Other/Misc.

_No response_

### Code of Conduct

bledden:
## Summary

Fixes #1587

I ran into this while testing eval workflows: running `nat eval --skip_workflow` without `--dataset` was silently deleting the `workflow_output.json` from a previous run. The cleanup step in `run_and_evaluate()` runs before the dataset is loaded, so by the time the code tries to load the previous output, it's already been wiped by `shutil.rmtree()`.

Since `--skip_workflow` exists specifically to reuse existing workflow output, it doesn't make sense to clean up the output directory when that flag is set. This change skips cleanup when `--skip_workflow` is active and logs an info message explaining why.

Normal eval behavior (without `--skip_workflow`) is unchanged.

## Test plan

- [ ] Run `nat eval` normally, confirm output directory cleanup still works
- [ ] Run `nat eval --skip_workflow` after a previous run, confirm `workflow_output.json` is preserved
- [ ] Existing eval tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected output directory cleanup so it is skipped when the workflow is intentionally bypassed, preserving generated files.
  * Added user-facing logs that indicate when cleanup operations are being skipped to improve transparency.

### DIFF
--- a/packages/nvidia_nat_eval/src/nat/plugins/eval/runtime/evaluate.py
+++ b/packages/nvidia_nat_eval/src/nat/plugins/eval/runtime/evaluate.py
@@ -585,9 +585,11 @@ class EvaluationRun:
         workflow_alias = self._get_workflow_alias(config.workflow.type)
         logger.debug("Loaded %s evaluation configuration: %s", workflow_alias, self.eval_config)
 
-        # Cleanup the output directory
-        if self.eval_config.general.output:
+        # Cleanup the output directory (skip when reusing existing workflow output)
+        if self.eval_config.general.output and not self.config.skip_workflow:
             self.cleanup_output_directory()
+        elif self.config.skip_workflow:
+            logger.info("Skipping output directory cleanup because --skip_workflow is set")
 
         # Generate a job_id if append_job_id_to_output_dir is enabled and no job_id provided
         if (self.eval_config.general.output
--- a/packages/nvidia_nat_eval/src/nat/plugins/eval/runtime/evaluate.py
+++ b/packages/nvidia_nat_eval/src/nat/plugins/eval/runtime/evaluate.py
@@ -586,10 +586,11 @@ class EvaluationRun:
         logger.debug("Loaded %s evaluation configuration: %s", workflow_alias, self.eval_config)
 
         # Cleanup the output directory (skip when reusing existing workflow output)
-        if self.eval_config.general.output and not self.config.skip_workflow:
-            self.cleanup_output_directory()
-        elif self.config.skip_workflow:
-            logger.info("Skipping output directory cleanup because --skip_workflow is set")
+        if self.eval_config.general.output:
+            if self.config.skip_workflow:
+                logger.info("Skipping output directory cleanup because --skip_workflow is set")
+            else:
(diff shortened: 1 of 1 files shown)


bledden:
## Validation

I wrote a targeted test to both reproduce the bug and validate the fix:

**Bug reproduction (reverted fix):**
- With the original code, `cleanup_output_directory()` gets called even when `skip_workflow=True`
- This deletes `workflow_output.json` before the dataset handler can read it

**Fix validation (with the change):**
- `skip_workflow=True`: `cleanup_output_directory` is correctly NOT called, `workflow_output.json` survives
- `skip_workflow=False`: `cleanup_output_directory` IS called as expected (normal behavior preserved)

Also ran the full eval test suite (`test_evaluate.py`) - all 23 tests pass including both parametrized `test_run_and_evaluate[True]` and `test_run_and_evaluate[False]`.

packages/nvidia_nat_eval/src/nat/plugins/eval/runtime/evaluate.py
```diff
@@ -585,9 +585,11 @@ async def run_and_evaluate(self,
         workflow_alias = self._get_workflow_alias(config.workflow.type)
         logger.debug("Loaded %s evaluation configuration: %s", workflow_alias, self.eval_config)
 
-        # Cleanup the output directory
-        if self.eval_config.general.output:
+        # Cleanup the output directory (skip when reusing existing workflow output)
+        if self.eval_config.general.output and not self.config.skip_workflow:
             self.cleanup_output_directory()
+        elif self.config.skip_workflow:
+            logger.info("Skipping output directory cleanup because --skip_workflow is set")
```
willkill07: Your logic was unnecessarily complicated.

A nested conditional not only has to perform fewer tests, but is easier to read.

```suggestion
        # Cleanup the output directory (skip when reusing existing workflow output)
        if self.eval_config.general.output:
            if self.config.skip_workflow:
                logger.info("Skipping output directory cleanup because --skip_workflow is set")
            else:
                self.cleanup_output_directory()
```

willkill07 (changes requested):

willkill07 (approved):

PR #1660, opened 2026-02-25
Add monkey patching for ADK's litellm import

saglave:
## Description

The `ADKProfilerHandler` monkey patch for `litellm.acompletion` works usually intercepts LLM calls made through `litellm`. But depending on how NAT and ADK are initiated, this would at times fail.

Root Cause
Google ADK's `lite_llm.py` uses a direct import of `acompletion`

This creates a local reference in the ADK module's namespace. When the profiler patched `litellm.acompletion`, it only modified the reference in the `litellm` module, not the local copy that ADK already had.

Since `LiteLLMClient.acompletion()` calls the local `acompletion`, it bypassed the monkey patch entirely.

This fix adds monkey-patching for `google.adk.models.lite_llm.acompletion` alongside the existing `litellm.acompletion` to ensure that the calls are always intercepted.

Closes

### DIFF
--- a/packages/nvidia_nat_adk/src/nat/plugins/adk/callback_handler.py
+++ b/packages/nvidia_nat_adk/src/nat/plugins/adk/callback_handler.py
@@ -59,6 +59,7 @@ class ADKProfilerHandler(BaseProfilerCallback):
         # Original references to Google ADK Tool and LLM methods (for uninstrumenting if needed)
         self._original_tool_call = None
         self._original_llm_call = None
+        self._original_adk_llm_call = None
         self._instrumented = False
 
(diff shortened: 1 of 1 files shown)


willkill07:
@saglave when this is ready for review could you retarget the `release/1.5` branch by cherry-picking this onto `release/1.5` and setting the target branch on github to `release/1.5` ?

saglave:
@willkill07 Sure. Will do so. Just running some tests to make sure that the LLM calls aren't logged twice.

willkill07:
The release/1.5 ship has sailed, but whenever this is ready, please let me know

PR #1683, opened 2026-02-27
Fix Mem0 metadata validation error and improve auto_memory_wrapper example

ericevans-nv:
## Description

Fix a `ValidationError` in the Mem0 memory editor when the Mem0 API returns `null` for
the `metadata` field in search results. The `MemoryItem` model expects a `dict`, but
`res.pop("metadata", {})` returns `None` when the key exists with a `null` value, causing
Pydantic validation to fail.

Additionally:
- Remove the Phoenix tracing exporter from `config_zep.yml` to eliminate noisy connection
  errors for users without a local Phoenix server
- Add `config_mem0.yml` as an alternative example config for users who do not have access
  to a Zep API key
- Update the auto_memory_wrapper README to document Mem0 as an alternative memory backend

## Summary by CodeRabbit

* **New Features**
  * Added support for Mem0 as an alternative memory backend with example configs and setup guidance.

* **Bug Fixes**
  * Improved robustness of metadata handling in memory operations.

* **Documentation**
  * Updated examples and docs to include Mem0 options and run commands; removed deprecated Phoenix tracing config and fixed a tutorial reference link.

### DIFF
--- a/packages/nvidia_nat_mem0ai/src/nat/plugins/mem0ai/mem0_editor.py
+++ b/packages/nvidia_nat_mem0ai/src/nat/plugins/mem0ai/mem0_editor.py
@@ -91,7 +91,7 @@ class Mem0Editor(MemoryEditor):
         memories = []
 
         for res in search_result["results"]:
-            item_meta = res.pop("metadata", {})
+            item_meta = res.pop("metadata", None) or {}
 
             memories.append(
                 MemoryItem(conversation=res.pop("input", []),
--- a/docs/source/get-started/tutorials/create-a-new-workflow.md
+++ b/docs/source/get-started/tutorials/create-a-new-workflow.md
@@ -174,7 +174,7 @@ async def text_file_ingest_function(config: TextFileIngestFunctionConfig, builde
 
 ## Creating the Workflow Configuration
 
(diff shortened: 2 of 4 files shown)


packages/nvidia_nat_mem0ai/src/nat/plugins/mem0ai/mem0_editor.py
```diff
@@ -91,7 +91,7 @@ async def search(self, query: str, top_k: int = 5, **kwargs) \
         memories = []
 
         for res in search_result["results"]:
-            item_meta = res.pop("metadata", {})
+            item_meta = res.pop("metadata", None) or {}
```
willkill07, line 94: Just a comment -- this is weird that this fixes it -- but if metadata was none -- very defensive :) and good catch

willkill07 (approved):
Thanks for adding a `config_mem0.yml` example :)

AnuradhaKaruppiah (approved):

packages/nvidia_nat_mem0ai/src/nat/plugins/mem0ai/mem0_editor.py
ericevans-nv: Right. This was subtle because null was being returned in some cases, and not an empty dict. This caused errors since the type is expected to be a dict.

PR #1696, opened 2026-03-02
fix: preserve nested Pydantic models when unpacking multi-arg functions

MylesShannon:
Replace `model_dump()` with `getattr()` in `_convert_input_pydantic` to prevent nested Pydantic models from being flattened to dicts. `model_dump()` recursively serializes all nested models, so parameters like `list[PromptItem]` become `list[dict]`, causing `AttributeError` when functions access model attributes (e.g. p.text).

Using `getattr()` preserves the original types as declared in function signatures. Updated test expectations to match the corrected behavior.

---

This fixes a bug where `_convert_input_pydantic` in `FunctionInfo` uses `model_dump()` to unpack multi-argument function inputs into keyword arguments. The problem is that `model_dump()` recursively serializes everything, including nested Pydantic models, into plain dicts.

For example, say you have a NAT function like this:
```python
class PromptItem(BaseModel):
    text: str
    start: int | None = None
    len: int | None = 300
```

```python
async def render_prompts(author: str, prompts: list[PromptItem], ...):
```

NAT wraps this multi-arg function into a single-arg function with an auto-generated input schema. When it unpacks that schema to call the real function, `model_dump()` converts `list[PromptItem]` into `list[dict]`. Then when the function tries to access p.text, it gets `AttributeError: 'dict' object has no attribute 'text'`.

Any NAT function with multiple arguments that includes nested Pydantic models in its signature would hit this.

The fix swaps `**value.model_dump()` for `**{k: getattr(value, k) for k in type(value).model_fields}`, which unpacks the top-level fields without recursively flattening nested models. Only the top-level unpacking behavior changes. The nested values are passed through as-is, matching what the function signature actually declares.

The test changes are just updating expected values that were inadvertently asserting the broken behavior (they compared against `str(dict)` output instead of `str(PydanticModel)` output, which didn't crash but was technically wrong).

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved nested model structures and types when converting composite inputs for both streaming and non-streaming calls, ensuring consistent string representations and preventing unexpected changes in how complex inputs appear during invocation. Public interfaces unchanged.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/builder/function_info.py
+++ b/packages/nvidia_nat_core/src/nat/builder/function_info.py
@@ -460,8 +460,8 @@ class FunctionInfo:
 
                 async def _convert_input_pydantic(value: input_schema) -> final_single_fn_desc.output_type:
 
-                    # Unpack the pydantic model into the arguments
-                    return await saved_final_single_fn(**value.model_dump())
+                    # Unpack the pydantic model into the arguments, preserving nested model types
+                    return await saved_final_single_fn(**{k: getattr(value, k) for k in type(value).model_fields})
 
                 final_single_fn = _convert_input_pydantic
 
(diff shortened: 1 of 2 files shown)


willkill07:
/ok to test [c4e6295](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1696/commits/c4e62957c461aff97a688395dc70ced7002caa9f)

willkill07 (approved):

willkill07:
/ok to test [683ee01](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1696/commits/683ee012744c35aad2a78e4cc7a6937b1a085b7c)

PR #1708, opened 2026-03-02
Fix Simple Calculator MCP example to return consistent timezone results

dagardner-nv:
## Description
* This example uses the [mcp-server-time](https://pypi.org/project/mcp-server-time/) tool which requires a timezone as an input. 
* This PR fixes a bug where the LLM was either hallucinating a timezone or using the timezone of it's own datacenter as input. Resulting in the timezone and thus the time changing between calls.
* Adds a new tool `current_timezone` which returns the system's time in IANA format.
* Clean up handling of timezone objects for the `current_datetime` method.
* Add dependency on `tzlocal` (currently a transitive dependency) since [datetime.astimezone](https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone) does not return timezone names in IANA format, and specifically the [tzinfo.tzname](https://docs.python.org/3/library/datetime.html#datetime.tzinfo.tzname) method does not return the name in any particular format.

## Summary by CodeRabbit

* **New Features**
  * Added an explicit timezone query so clients can obtain the current timezone before requesting time-based data.
  * Time-based queries now prefer the detected timezone for more accurate results.
  * Expanded calculator tools in the MCP example: add, subtract, multiply, divide, compare.
  * Added an LLM configuration block to the MCP config for model settings.

* **Documentation**
  * Updated example README and MCP configs to reflect new tools and include lists.

* **Tests**
  * Added tests validating current time and timezone outputs.

* **Chores**
  * Added a dependency to support timezone detection.


Salonijain27 (approved):

examples/MCP/simple_calculator_mcp/README.md
```diff
@@ -80,6 +80,12 @@ function_groups:
     server:
       transport: streamable-http
       url: "http://localhost:9901/mcp"
+    include:
+      - calculator__add
+      - calculator__subtract
+      - calculator__multiply
+      - calculator__divide
+      - calculator__compare
```
willkill07, line 92: Is this aligning the config with the README?

willkill07 (approved):

examples/MCP/simple_calculator_mcp/README.md
dagardner-nv: Yes in the doc there's yaml doc mirroring the config, but it was missing this section.

Alternately we could pare it back a little bit with a note like "Truncated for simplicity, refer to `examples/MCP/simple_calculator_mcp/configs/config-mcp-client.yml` for the full configuration"

PR #1717, opened 2026-03-03
feat(agent): stream tool call chunks

MylesShannon:
Stream tool call chunks to frontend:
- Add ChoiceDeltaToolCall and ChoiceDeltaToolCallFunction models to
  ChoiceDelta in api_server.py for OpenAI-compatible tool call deltas
- Change _stream_fn return type from AsyncGenerator[str] to
  AsyncGenerator[ChatResponseChunk] and yield structured chunks for
  both content tokens and tool call deltas
- Add unit tests for tool call chunk serialization

Fix agent_node streaming accumulation bug:
- When agent_node accumulates AIMessageChunk objects via astream, the
  resulting chunk has tool_call_chunks but additional_kwargs["tool_calls"]
  (the wire format) is left empty, causing the second LLM call to fail
  with "messages with role 'tool' must be a response to a preceeding
  message with 'tool_calls'"
- Add _chunk_to_message helper that converts accumulated AIMessageChunk
  to AIMessage, using LangChain's convert_to_openai_messages to
  reconstruct the wire format rather than hardcoding it

Streaming handler hardening:
- Guard msg.content with isinstance(str) check for non-string content
- Fall back to msg.tool_calls when tool_call_chunks is missing (AIMessage)
- Normalize tool call index via enumerate to handle None values
- Serialize dict args to JSON string for AIMessage tool_calls
- Yield ChatResponseChunk instead of plain string in GraphRecursionError handler

## Summary by CodeRabbit

* **New Features**
  * Streaming now emits unified, structured chunks that include tool-call deltas alongside content, with per-chunk IDs, timestamps, and consistent formatting.
  * Final streamed chunks are converted into full message objects that preserve tool-call details and arguments.

* **Tests**
  * Added tests verifying tool-call chunk serialization, delta formatting, content-only behavior, argument preservation, and rehydration of tool-call data.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/data_models/api_server.py
+++ b/packages/nvidia_nat_core/src/nat/data_models/api_server.py
@@ -281,10 +281,25 @@ class ChoiceMessage(BaseModel):
     role: UserMessageContentRoleType | None = None
 
 
+class ChoiceDeltaToolCallFunction(BaseModel):
+    """Function details within a streamed tool call delta (OpenAI-compatible)."""
+    name: str | None = None
+    arguments: str | None = None
+
+
+class ChoiceDeltaToolCall(BaseModel):
+    """Tool call delta for streaming responses (OpenAI-compatible)."""
+    index: int
+    id: str | None = None
+    type: str | None = None
+    function: ChoiceDeltaToolCallFunction | None = None
+
+
 class ChoiceDelta(BaseModel):
     """Delta object for streaming responses (OpenAI-compatible)"""
     content: str | None = None
     role: UserMessageContentRoleType | None = None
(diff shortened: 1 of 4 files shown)


willkill07 (approved):

willkill07:
/ok to test [8316498](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1717/commits/8316498d1cdf4ebbef7c9e9b6fe9dece002f5121)

willkill07:
/ok to test [60d8d9a](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1717/commits/60d8d9a69f46205c0c1c9a6b30df8e840a03ac4c)

PR #1769, opened 2026-03-08
Add AtifBaseEvaluator for reusable concurrent ATIF custom evaluators

AnuradhaKaruppiah:
## Description

This PR introduces a reusable ATIF custom-evaluator base class to reduce boilerplate and prevent per-evaluator orchestration mistakes (for example serial execution instead of bounded concurrent execution).

## Summary by CodeRabbit

* **New Features**
  * New evaluator base enabling per-item evaluation with configurable concurrency, per-item scoring, and aggregated average scores.
  * Added optional case-insensitive normalization and tool-call counting; cosine-similarity scoring preserved.

* **Documentation**
  * Updated guides, examples, and snippets to reflect item-level evaluators, base-class orchestration, and concurrency configuration.

* **Tests**
  * Added tests covering per-item scoring, average-score aggregation, and bounded concurrency behavior.


packages/nvidia_nat_eval/tests/eval/evaluator/test_atif_base_evaluator.py
```diff
@@ -0,0 +1,76 @@
...
+        self._active += 1
+        self.peak_active = max(self.peak_active, self._active)
+        try:
+            await asyncio.sleep(0.01)
+            return EvalOutputItem(id=sample.item_id, score=1.0, reasoning={"score": 1.0})
+        finally:
+            self._active -= 1
+
+
+async def test_atif_base_evaluator_computes_average_score():
```
bbednarski9, line 57: additional test case with mod(samples, concurrency) != 0 could be useful to test semaphore completes entire set

packages/nvidia_nat_eval/src/nat/plugins/eval/evaluator/__init__.py
```diff
@@ -12,3 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from .atif_base_evaluator import AtifBaseEvaluator
+
+__all__ = ["AtifBaseEvaluator"]
```
bbednarski9: Inconsistent with how BaseEvaluator is imported directly from the Module path

```suggestion
from .atif_base_evaluator import AtifBaseEvaluator
from .base_evaluator import BaseEvaluator

__all__ = ["AtifBaseEvaluator", "BaseEvaluator"]
```

AnuradhaKaruppiah: good catch

bbednarski9:
I ran through the notebook. It only installs `nvidia_nat_eval` but the data model: `from nat.data_models.atif import ATIFTrajectory` comes from `nat_core`. So we likely need to add a uv pip install for nat core and update section 6 of the notebook to include that nat core is also needed.

I know it's kind of implied, but users wont be able to just shift-enter run through it without that error otherwise

bbednarski9 (approved):
see comments above - rubber stamp for when they are resolved.

AnuradhaKaruppiah:
> I ran through the notebook. It only installs `nvidia_nat_eval` but the data model: `from nat.data_models.atif import ATIFTrajectory` comes from `nat_core`. So we likely need to add a uv pip install for nat core and update section 6 of the notebook to include that nat core is also needed.
> 
> I know it's kind of implied, but users wont be able to just shift-enter run through it without that error otherwise

nvidia-nat-core is a dep in nvidia-nat-eval, I think the problem comes for the use of no-deps in the NB. Let me double check and remove thats. Good catch

packages/nvidia_nat_eval/tests/eval/evaluator/test_atif_base_evaluator.py
AnuradhaKaruppiah: good suggestion. i have added a test to ensure the "remainder" samples are evaluated,

PR #1773, opened 2026-03-09
Improves FastMCP dev experience and docs 

AnuradhaKaruppiah:
## Description

- Added reload filtering to nat fastmcp server dev:
  - New flags: --reload-include-glob, --reload-exclude-glob
  - Default excludes for noisy files (*.log, *.tmp, *.pyc, __pycache__, and similar)
  - Prevents restart loops when watched dirs contain active writes
- Added tests for CLI option wiring and include/exclude filtering behavior.
- Updated FastMCP example docs:
  - Clarified mcp-server-time dependency for config-mcp-client.yml runs
  - Improved --tool_names discoverability

## Summary by CodeRabbit

* **New Features**
  * Dev server adds repeatable --reload-include-glob / --reload-exclude-glob flags to limit which file changes trigger reloads, helping avoid noisy restarts.

* **Documentation**
  * CLI docs show reload-filtering usage and updated examples for publishing multiple tools.
  * Example README adds install guidance and how to expose selected tools.

* **Tests**
  * New tests cover glob-based filtering and CLI option behavior.

### DIFF
--- a/docs/source/run-workflows/fastmcp-server.md
+++ b/docs/source/run-workflows/fastmcp-server.md
@@ -75,7 +75,12 @@ nat fastmcp server dev --config_file examples/getting_started/simple_calculator/
   --watch-path examples/getting_started/simple_calculator/src
 ```
 
-If your watch path contains actively written files (such as logs, temp files, or cache artifacts), limit which changes trigger reload with include and exclude globs:
+By default, developer mode ignores common noisy files such as `*.log`, `*.tmp`, and `*.temp`.
+To further control which changes trigger reloads, use include and exclude globs:
+
+- `--reload-include-glob` narrows reloads to matching paths.
+- `--reload-exclude-glob` removes matches from that set.
+- When include globs are provided, they take precedence over default excludes.
 
 ```bash
 nat fastmcp server dev --config_file examples/getting_started/simple_calculator/configs/config.yml \
--- a/docs/source/run-workflows/fastmcp-server.md
+++ b/docs/source/run-workflows/fastmcp-server.md
@@ -127,7 +127,8 @@ You can publish a subset of tools using the `--tool_names` flag:
 
 ```bash
 nat fastmcp server run --config_file examples/getting_started/simple_calculator/configs/config.yml \
-  --tool_names calculator
+  --tool_names calculator__multiply \
+  --tool_names calculator__divide
 ```
 
 ### Mounting at Custom Paths
--- a/docs/source/run-workflows/fastmcp-server.md
+++ b/docs/source/run-workflows/fastmcp-server.md
@@ -75,6 +75,16 @@ nat fastmcp server dev --config_file examples/getting_started/simple_calculator/
   --watch-path examples/getting_started/simple_calculator/src
 ```
 
+If your watch path contains actively written files (such as logs, temp files, or cache artifacts), limit which changes trigger reload with include and exclude globs:
+
+```bash
+nat fastmcp server dev --config_file examples/getting_started/simple_calculator/configs/config.yml \
+  --watch-path examples/getting_started/simple_calculator/src \
+  --reload-include-glob "*.py" \
+  --reload-include-glob "*.yml" \
+  --reload-exclude-glob "*.log"
+```
+
 ### Generating MCP Client Configuration Snippets
 
 Use `nat fastmcp server install` to generate MCP client configuration snippets for a FastMCP server. This command does not modify your environment.
--- a/packages/nvidia_nat_fastmcp/src/nat/plugins/fastmcp/cli/utils.py
+++ b/packages/nvidia_nat_fastmcp/src/nat/plugins/fastmcp/cli/utils.py
@@ -16,6 +16,7 @@
 
 from __future__ import annotations
 
+from fnmatch import fnmatch
 from collections.abc import Iterable
 from collections.abc import Iterator
 from pathlib import Path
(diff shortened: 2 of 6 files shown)


packages/nvidia_nat_fastmcp/src/nat/plugins/fastmcp/cli/utils.py
```diff
@@ -18,16 +18,57 @@
...
 
+DEFAULT_RELOAD_EXCLUDE_GLOBS: tuple[str, ...] = (
+    "*.log",
+    "*.tmp",
+    "*.temp",
+    "*.swp",
+    "*.pyc",
+    "*.pyo",
+    "*__pycache__",
+    "*__pycache__/*",
```
mnajafian-nv:
```suggestion
# watchfiles' DefaultFilter already excludes __pycache__, *.pyc, *.pyo, *.swp,
# .git, node_modules, etc. These patterns cover files DefaultFilter misses.

DEFAULT_RELOAD_EXCLUDE_GLOBS: tuple[str, ...] = (
    "*.log",
    "*.tmp",
    "*.temp",
)
```

mnajafian-nv: watchfiles.watch() uses DefaultFilter by default, which already rejects events for __pycache__ dirs (via ignore_dirs), *.pyc/*.pyo (via \.py[cod]$), and *.swp (via \.sw.$). Those events never reach _filter_change_set, so 5 of 8 patterns here can never match anything.

I'd remove them and add a comment explaining what watchfiles already handles. Keeps the tuple honest about what it actually does.

willkill07 (approved):

packages/nvidia_nat_fastmcp/src/nat/plugins/fastmcp/cli/utils.py
```diff
@@ -18,16 +18,57 @@
...
+    for change_type, changed_path in changes:
+        if include_globs and not any(_glob_matches(changed_path, pattern) for pattern in include_globs):
+            continue
+        if exclude_globs and any(_glob_matches(changed_path, pattern) for pattern in exclude_globs):
+            continue
+        filtered_changes.add((change_type, changed_path))
+    return filtered_changes
+
 
 def iter_file_changes(
```
mnajafian-nv, line 64: suggestion
```
def iter_file_changes(
        paths: Iterable[Path],
        debounce_ms: int = 750,
        include_globs: Iterable[str] = (),
        exclude_globs: Iterable[str] = (),
) -> Iterator[set[tuple[Change, str]]]:
    """Yield filtered file change sets using watchfiles with debounce."""
    watch_paths = [str(path) for path in paths]
    include_patterns = tuple(pattern.strip() for pattern in include_globs if pattern.strip())
    user_exclude_patterns = tuple(pattern.strip() for pattern in exclude_globs if pattern.strip())
    if include_patterns:
        exclude_patterns = user_exclude_patterns
    else:
        exclude_patterns = DEFAULT_RELOAD_EXCLUDE_GLOBS + user_exclude_patterns
    for changes in watch(*watch_paths, debounce=debounce_ms):
        filtered_changes = _filter_change_set(changes, include_patterns, exclude_patterns)
        if filtered_changes:
            yield filtered_changes
```

mnajafian-nv: Right now, if a user runs:
--reload-include-glob "*.log"
they're explicitly saying "reload on log changes." But *.log is in DEFAULT_RELOAD_EXCLUDE_GLOBS, so the exclude check kills it silently. The user's intent is ignored with no warning.

When the user provides explicit include patterns, they've already narrowed the scope. Default excludes shouldn't fight that. I'd suggest only applying DEFAULT_RELOAD_EXCLUDE_GLOBS when no include globs are specified (the "watch everything" path). When includes are provided, only user-supplied excludes should apply.

packages/nvidia_nat_fastmcp/src/nat/plugins/fastmcp/cli/utils.py
```diff
@@ -18,16 +18,57 @@
...
+    "*.swp",
+    "*.pyc",
+    "*.pyo",
+    "*__pycache__",
+    "*__pycache__/*",
+)
+
+
+def _glob_matches(path: str, pattern: str) -> bool:
+    normalized_path = path.replace("\\", "/")
```
mnajafian-nv, line 43:
```suggestion
    """Check whether *path* matches a glob *pattern*.
    Matches against both the full path and the basename, so ``*.py``
    works regardless of directory depth.  Slashes are normalized to ``/``
    for cross-platform consistency.  Note: ``fnmatch``'s ``*`` matches
    path separators, unlike shell globbing.
    """
    normalized_path = path.replace("\\", "/")
```

mnajafian-nv: CodeRabbit flagged docstring coverage at 33% vs the 80% threshold. Both _glob_matches and _filter_change_set are missing docstrings. Even for private functions, a one-liner explaining the dual match behavior (full path + basename) and the fnmatch slash semantics would help future readers.

docs/source/run-workflows/fastmcp-server.md
```diff
@@ -75,6 +75,16 @@ nat fastmcp server dev --config_file examples/getting_started/simple_calculator/
   --watch-path examples/getting_started/simple_calculator/src
 ```
 
+If your watch path contains actively written files (such as logs, temp files, or cache artifacts), limit which changes trigger reload with include and exclude globs:
```
mnajafian-nv:
```suggestion
By default, the dev server ignores changes to common noisy files (`*.log`, `*.tmp`, `*.temp`) in addition to the patterns already excluded by `watchfiles` (`__pycache__`, `*.pyc`, `*.swp`, etc.).
To further control which changes trigger a reload, use include and exclude globs:

```

mnajafian-nv: Docs would benefit from mentioning what default excludes so they don't pass redundant flags or get confused when certain changes don't trigger reloads. A short sentence listing the defaults and explaining the interaction would help.

packages/nvidia_nat_fastmcp/tests/test_fastmcp.py
```diff
@@ -90,6 +96,44 @@ def test_fastmcp_cli_groups() -> None:
...
+        (Change.modified, "/tmp/__pycache__/worker.pyc"),
+        (Change.modified, "/tmp/worker.py"),
+    }
+    filtered = _filter_change_set(changes,
+                                  include_globs=(),
+                                  exclude_globs=("*.log", "*__pycache__", "*__pycache__/*", "*.pyc"))
+    assert filtered == {(Change.modified, "/tmp/worker.py")}
+
+
+def test_iter_file_changes_applies_include_and_exclude_globs(monkeypatch: pytest.MonkeyPatch) -> None:
```
mnajafian-nv, line 118: Do we benefit from adding a test case for when a user's include pattern overlaps with a default exclude? For example, `include_globs=("*.log",)` with no user exclude. Right now that path silently drops the log event because `*.log` is in `DEFAULT_RELOAD_EXCLUDE_GLOBS`. Whether that's the desired behavior or not, it should be tested explicitly.

mnajafian-nv (approved):
Nice quality-of-life feature for the FastMCP dev loop! I give this a conditional approval pending your review of the inline suggestions I made :)

packages/nvidia_nat_fastmcp/src/nat/plugins/fastmcp/cli/utils.py
AnuradhaKaruppiah: good point, addressed

packages/nvidia_nat_fastmcp/src/nat/plugins/fastmcp/cli/utils.py
AnuradhaKaruppiah: Updated iter_file_changes() in packages/nvidia_nat_fastmcp/src/nat/plugins/fastmcp/cli/utils.py:
if --reload-include-glob is provided, default excludes no longer silently block those includes
only user-provided excludes apply in that mode

packages/nvidia_nat_fastmcp/src/nat/plugins/fastmcp/cli/utils.py
AnuradhaKaruppiah: Added docstrings for _glob_matches, _filter_change_set, and expanded iter_file_changes with Args/Yields.

packages/nvidia_nat_fastmcp/tests/test_fastmcp.py
AnuradhaKaruppiah: Added test: include pattern overlapping default excludes (include_globs=("*.log",)) now behaves as intended.

docs/source/run-workflows/fastmcp-server.md
AnuradhaKaruppiah: good suggestion, addressed

PR #1775, opened 2026-03-10
Add centralized user identity management and custom WebSocket authentication

ericevans-nv:
## Description
This PR introduces a structured user identity system that provides consistent identity resolution across all credential methods — session cookies, JWT Bearer tokens, API keys, and HTTP Basic Auth — regardless of how or where credentials arrive (HTTP headers, WebSocket headers, or WebSocket auth messages).
### User Identity Abstraction
`UserInfo` is the single resolved identity object for every user in the system. It does not know or care how the user was identified — it only holds a deterministic `user_id` derived from whatever credential was provided. The same credential always produces the same UUID, making identity stable and predictable across requests, reconnections, and server restarts.
`UserManager` is a fully stateless identity resolver. It accepts any supported identity source — JWT claims, API keys, username/password credentials, or session cookies — auto-detects the credential type, resolves the user, and creates a deterministic `user_id` if one does not already exist. Callers never need to know what kind of credential was used. It is not responsible for session lifecycle, context variables, or workflow routing — it delegates persistence and session scoping to `SessionManager`. Its only job is to answer "who is this?" and hand back a stable identity.
### Identity from HTTP and WebSocket Connections
For HTTP requests and WebSocket upgrade headers, `UserManager.extract_user_from_connection()` inspects the connection and resolves identity automatically:
- **Session cookie** (`nat-session`) — used as-is as the identity key
- **`Authorization: Bearer <token>`** — if the token has 3 dot-separated parts, it is decoded as a JWT and identity is extracted from claims (`sub > email > preferred_username`); otherwise it is treated as an opaque API key
- **`X-API-Key: <key>`** — treated as an opaque API key (fallback when `Authorization` is not present)
- **`Authorization: Basic <base64>`** — decoded into `username:password`, and the raw base64 credential becomes the identity key

For WebSocket connections where headers or cookies are not available (e.g., browser clients), a new `auth_message` type allows clients to authenticate after connecting. The client sends an `auth_message` with JWT, API key, or username/password credentials before sending its first workflow message. The server validates the credentials, resolves a user identity, and responds with an `auth_response_message` containing the resolved `user_id` on success or structured error details using the `USER_AUTH_ERROR` error type on failure. The resolved identity persists for the duration of the WebSocket connection and applies automatically to all subsequent messages.
### Documentation
A new User Identity Resolution page documents the supported identity sources, resolution logic, and user_id derivation rules. The WebSocket API reference has been updated with the auth_message and auth_response_message schemas, payload examples for all three auth methods (JWT, API key, Basic), and the end-to-end authentication flow.
### Changes
- Add `UserInfo`, `JwtUserInfo`, and `BasicUserInfo` data models with deterministic UUID derivation and mutual-exclusivity validation
- Add `UserManager` with auto-detection for session cookies, JWT Bearer, API key Bearer, `X-API-Key` header, and HTTP Basic Auth credentials
- Add `AuthMethod` enum, `WebSocketAuthMessage`, and `WebSocketAuthResponseMessage` to the WebSocket message schema
- Add WebSocket `auth_message` flow for post-connection authentication with JWT, API key, or username/password payloads
- Extract shared JWT decode logic into `jwt_utils` module
- Align JWT identity claim precedence with RFC 7519 (`sub > email > preferred_username`)
- Rename `JwtUserInfo.first_name`/`last_name` to `given_name`/`family_name` per OIDC Core 1.0 standard claims
- Fix `SecretStr` serialization during Uvicorn worker config handoff using `SerializableSecretStr`/`OptionalSecretStr`
- Refactor `SessionManager.session()` to delegate credential extraction to `UserManager`
- Remove `connection_auth.py`, `UserManagerHolder`, and `ContextState.user_manager` context variable
- Remove the `security` field from WebSocket user messages
- Add user identity resolution documentation page (`components/auth/user-identity.md`)
- Update `websockets.md` with auth message schemas, examples, and end-to-end authentication flows
- Expand unit tests for serialization round-trip, claim precedence, cross-source `user_id` consistency, and auth payload resolution


docs/source/reference/rest-api/websockets.md
```diff
@@ -50,17 +52,14 @@ to the client.
     - Format: String for text messages and array for contents which can have attachments such as image, audio and videos. See above example.
     -   Attachments support OpenAI compatible chat objects such as (Default, Image, Audio, and Streaming)
 - `status`: Indicates the processing state of the message.
```
willkill07, line 54: If status is still optional, then we should still include text indicating this.

docs/source/reference/rest-api/websockets.md
```diff
@@ -50,17 +52,14 @@ to the client.
...
+    - For `auth_response_message`: `success`, `error`.
 - `timestamp`: Captures when the message was created or updated.
      - Format: ISO 8601 (e.g., `2025-01-13T10:00:00Z`).
  - `user`: Stores user information - OPTIONAL
     -   name: User name
     -   email: User email
     -   other info: Any other information
-- `security`: Stores security information such as `api_key`, auth token etc. - OPTIONAL
-    -   `api_key`: API key
-    - token: auth or access token
```
willkill07, line 63: Since this is being removed, PR should be marked "breaking".

I changed this for you.

packages/nvidia_nat_core/src/nat/data_models/config.py
```diff
@@ -310,6 +311,9 @@ class Config(HashableBaseModel):
     # Authentication Configuration
     authentication: dict[str, AuthProviderBaseConfig] = Field(default_factory=dict)
 
+    # Pre-defined users (keyed by username)
+    users: dict[str, UserInfo] = Field(default_factory=dict)
```
willkill07: Should this not be defined on a per-session basis? The only reason I ask is that the UserManager was effectively a `ContextVar`.

Also, there doesn't seem to be any sort of cleanup here which seems problematic for long-running services.

Followup: I don't see this being used anywhere... should be this be removed? Or are things not fully wired up?

packages/nvidia_nat_core/src/nat/runtime/user_manager.py
```diff
@@ -0,0 +1,251 @@
...
+            scopes=scopes,
+            issuer=claims.get("iss") if isinstance(claims.get("iss"), str) else None,
+            subject=claims.get("sub") if isinstance(claims.get("sub"), str) else None,
+            audience=audience,
+            expires_at=claims.get("exp") if isinstance(claims.get("exp"), int) else None,
+            issued_at=claims.get("iat") if isinstance(claims.get("iat"), int) else None,
+            client_id=(claims.get("azp") or claims.get("client_id")
+                       if isinstance(claims.get("azp"), str) or isinstance(claims.get("client_id"), str) else None),
+            claims=claims,
+        )
```
willkill07, line 273: Nit: all of this logic should probably be in a `from_jwt_claims()` function in UserInfo. The separation seems arbitrary and someone confusing because the UserManager has no reason to know about jwt details.

packages/nvidia_nat_core/src/nat/data_models/user_info.py
```diff
@@ -0,0 +1,159 @@
...
+            return self._api_key
+        if self._session_cookie is not None:
+            return self._session_cookie
+        return None
+
+    @classmethod
+    def _from_session_cookie(cls, cookie: str) -> "UserInfo":
+        instance: UserInfo = cls()
+        object.__setattr__(instance, "_session_cookie", cookie)
+        instance._set_user_id(cookie)
```
willkill07, line 169: nit: Am I missing something? Why wouldn't this work? I ask because `PrivateAttr` just makes it so it's not automatically dumped. You can still access via standard methods.
```suggestion
        instance: UserInfo = UserInfo()
        instance._session_cookie = cookie
        instance._set_user_id(cookie)
```

willkill07 (changes requested):
Overall this is a clean abstraction.

Changes requested due to:
- what seem to be important test failures
- logic organization
- dangling fields/logic
- documentation gap

packages/nvidia_nat_core/src/nat/data_models/user_info.py
ericevans-nv: You're right, `PrivateAttr` keeps fields out of serialization, but since the model has frozen attributes, attribute assignment is blocked, which is why object.__setattr__ is needed to set them.

packages/nvidia_nat_core/src/nat/runtime/user_manager.py
ericevans-nv: Fair enough.

docs/source/reference/rest-api/websockets.md
ericevans-nv: Yeah, unfortunately we may have shot ourselves in the foot here since I don't think anyone ever actually used that parameter, but it's probably best to make it a breaking change since it's being removed. It was never used in the code.

packages/nvidia_nat_core/src/nat/data_models/config.py
ericevans-nv: Since UserManager is stateless, per-session scoping is unnecessary. The SessionManager owns the ContextVar lifecycle (set on entry, reset in finally). Cleanup of per-user workflow builders is handled by SessionManager's periodic task. UserManager is now fully wired.

docs/source/reference/rest-api/websockets.md
willkill07, line 1: The JSON formatting changes are weird and non-standard. Any chance of reverting those back?

There's a lot of unnecessary changes to this file.

packages/nvidia_nat_core/src/nat/data_models/config.py
willkill07: I still don't understand the purpose of pre-seeding in the config.

Was this a requested feature? Because this seems orthogonal to the PR's title + goals.

ericevans-nv: This wasn't explicitly requested, but it follows the same pattern as the WebSocket auth_message—both create a user ID from credentials, just from different sources (YAML at startup vs. a WebSocket message on connection before any user message is sent). Pre-seeding builds per-user workflows eagerly for known users so their first request can access protected resources readily.

willkill07 (approved):

docs/source/components/auth/user-identity.md
```diff
@@ -0,0 +1,75 @@
...
+
+## User ID Derivation
+
+Each identity source produces a deterministic UUID v5 using a toolkit-specific namespace. The identity key varies by credential type:
+
+| **Identity Source** | **Identity Key** | **Standards Reference** |
+|---|---|---|
+| JWT | First non-empty value from `sub`, `email`, `preferred_username` | [RFC 7519 Section 4.1.2](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.2), [OpenID Connect Core 1.0 Section 5.1](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) |
+| API key | Raw API key string | — |
+| Basic Auth | `base64(username:password)` | [RFC 7617](https://www.rfc-editor.org/rfc/rfc7617) |
```
mnajafian-nv:
```suggestion
| Basic Auth | `base64(username:password)` ¹ | [RFC 7617](https://www.rfc-editor.org/rfc/rfc7617) |

¹ Because the password is part of the identity key, changing a password produces a new `user_id`. The user's prior per-user workflow state (conversation history, builders) becomes inaccessible.
```

packages/nvidia_nat_core/src/nat/data_models/user_info.py
```diff
@@ -0,0 +1,179 @@
...
+        """Return the first non-empty value using ``sub > email > preferred_username`` precedence.
+
+        ``sub`` is the stable, locally-unique identifier per RFC 7519 Section 4.1.2.
+        ``email`` and ``preferred_username`` are OIDC fallbacks (OpenID Connect Core 1.0 Section 5.1).
+        """
+        for key in ("sub", "email", "preferred_username"):
+            val: typing.Any = self.claims.get(key)
+            if val and isinstance(val, str) and val.strip():
+                return val.strip()
+        return None
```
mnajafian-nv, line 71:
```suggestion

    @property
    def identity_claim(self) -> str | None:
        """Return the first non-empty value using ``sub > email > preferred_username`` precedence.
        ``sub`` is the stable, locally-unique identifier per RFC 7519 Section 4.1.2.
        ``email`` and ``preferred_username`` are OIDC fallbacks (OpenID Connect Core 1.0 Section 5.1).
        """
        for key in ("sub", "email", "preferred_username"):
            val: typing.Any = self.claims.get(key)
            if val and isinstance(val, str) and val.strip():
                return val.strip()
        for attr_val in (self.subject, self.email, self.preferred_username):
            if attr_val and isinstance(attr_val, str) and attr_val.strip():
                return attr_val.strip()
        return None
```

mnajafian-nv: identity_claim only inspects self.claims (the raw dict). If someone constructs JwtUserInfo(subject="sub-123", claims={}) directly, identity_claim returns None even though self.subject is set. Since JwtUserInfo is a public Pydantic model, consider falling back to the typed fields.

packages/nvidia_nat_core/src/nat/data_models/user_info.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class BasicUserInfo(BaseModel):
```
mnajafian-nv, line 74: Documentation request:

BasicUserInfo derives credential = base64(username:password), and that credential is the identity key for UUID generation. This means if a user changes their password, they get a new user_id and lose access to their prior per-user state (conversation history, workflow builders, cached tokens). This may be intentional, but the user-identity.md documentation should explicitly call this out in the "User ID Derivation" table. It's a non-obvious consequence that will surprise deployers.

packages/nvidia_nat_core/src/nat/runtime/user_manager.py
```diff
@@ -0,0 +1,293 @@
...
+            token_value: str = payload.token.get_secret_value()
+            if not token_value:
+                raise ValueError("API key token is empty")
+            return UserInfo._from_api_key(token_value)
+
+        if isinstance(payload, BasicAuthPayload):
+            return UserInfo(basic_user=BasicUserInfo(
+                username=payload.username,
+                password=payload.password,
+            ))
```
mnajafian-nv, line 144:
```suggestion
        if isinstance(payload, BasicAuthPayload):
            return UserInfo(basic_user=BasicUserInfo(
                username=payload.username,
                password=payload.password,
            ))

        raise ValueError(f"Unsupported auth payload type: {type(payload).__name__}")
```

mnajafian-nv: If payload doesn't match any of the three known types, this function falls off the end and implicitly returns None. The caller in _process_auth_message then calls .get_user_id() on None -> AttributeError. Should raise explicitly.

docs/source/components/auth/user-identity.md
```diff
@@ -0,0 +1,75 @@
...
+Identity resolution provides the following capabilities:
+
+- **Automatic credential detection**: The server inspects incoming connections for known credential formats and resolves them into a user identity without any workflow-level configuration.
+- **Deterministic user IDs**: Each credential produces a stable UUID v5. The same credential always resolves to the same `user_id`, regardless of transport (HTTP or WebSocket) or credential format.
+- **Multiple credential types**: Session cookies, JWT Bearer tokens, opaque API keys, `X-API-Key` headers, HTTP Basic Auth, and WebSocket auth messages are all supported.
+- **Per-user workflow support**: When a workflow is configured as per-user, the resolved `user_id` is used to isolate workflow state per user. Each user gets their own workflow instance.
+
+:::{note}
+Identity resolution is an identity mapping step, not an authentication or authorization layer. JSON Web Tokens are decoded without signature verification to extract identity claims. Credential validation (for example, JWKS verification or OAuth flows) should be handled upstream or via an [authentication provider](./api-authentication.md).
+:::
```
mnajafian-nv:
```suggestion
:::{warning}
Identity resolution is an identity mapping step, not an authentication or authorization layer. **JWTs are decoded with `verify_signature=False`** — the server trusts whatever credential arrives. The resolved `user_id` controls access to per-user workflow state (conversation history, builders, cached tokens). In production, deploy an authenticating reverse proxy or auth middleware that validates JWTs before they reach NAT. Without upstream verification, any party that can send HTTP requests to Nemo Agent Toolkit can impersonate any user. For credential validation, see [Authentication Providers](./api-authentication.md).
:::
```

mnajafian-nv: The docstring correctly says "intended for identity extraction only," but the resolved user_id gates access to per-user isolated state (conversation history, builders, cached tokens). If user A sends user B's JWT, user A gets user B's state. This is by design (identity resolution, not authentication), but deployers need to know they must place an authenticating reverse proxy in front of NAT. Consider adding a logger.info on first invocation or promoting the note in user-identity.md from :::{note} to :::{warning}.

packages/nvidia_nat_core/src/nat/runtime/session.py
```diff
@@ -462,7 +454,6 @@ async def _get_or_create_per_user_builder(self, user_id: str) -> tuple["PerUserW
     @asynccontextmanager
     async def session(self,
```
mnajafian-nv, line 455: token_user_input and token_user_authentication are set before the try block, but extract_user_from_connection() can raise ValueError (e.g., malformed JWT) also before try. If that happens, the finally block never runs, and those context vars leak into the surrounding scope. Move all context var setup inside the try, or wrap the extraction in its own try/except.

I suggest

  ```
      token_user_input = None
        token_user_authentication = None
        token_workflow_parent_id = None
        token_workflow_parent_name = None
        token_user_id = None

        try:
            if user_input_callback is not None:
                token_user_input = self._context_state.user_input_callback.set(user_input_callback)

            if user_authentication_callback is not None:
                token_user_authentication = self._context_state.user_auth_callback.set(user_authentication_callback)

            if isinstance(http_connection, WebSocket):
                if user_id is None:
                    user_info: UserInfo | None = UserManager.extract_user_from_connection(http_connection)
                    if user_info is not None:
                        user_id = user_info.get_user_id()
                self.set_metadata_from_websocket(http_connection, user_message_id, conversation_id)

            if isinstance(http_connection, Request):
                if user_id is None:
                    user_info = UserManager.extract_user_from_connection(http_connection)
                    if user_info is not None:
                        user_id = user_info.get_user_id()
                token_workflow_parent_id, token_workflow_parent_name = \
                    await self.set_metadata_from_http_request(http_connection)

            token_user_id = self._context_state.user_id.set(user_id)

            if not user_id and self._is_workflow_per_user:
                raise ValueError(...)
```

File: packages/nvidia_nat_core/src/nat/front_ends/fastapi/message_handler.py [hunk 12]
```diff
@@ -96,6 +101,7 @@ def __init__(self,
         self._workflow_schema_type: str | None = None
         self._user_interaction: UserInteraction | None = None
         self._pending_observability_trace: ResponseObservabilityTrace | None = None
+        self._user_id: str | None = None
```
mnajafian-nv, line 104: _user_id is set by _process_auth_message but is not copied in _restore_execution_state(). If a WebSocket reconnects and the handler is swapped, the authenticated identity is lost. Either propagate _user_id in _restore_execution_state(), or document that auth-message identity is per-handler and clients must re-authenticate after reconnect.

packages/nvidia_nat_core/src/nat/runtime/user_manager.py
```diff
@@ -0,0 +1,293 @@
...
+            cookies: dict[str, str] = dict(connection.cookies) if connection.cookies else {}
+            return cookies.get(SESSION_COOKIE_NAME)
+
+        if isinstance(connection, WebSocket) and hasattr(connection, "scope") and "headers" in connection.scope:
+            for name, value in connection.scope.get("headers", []):
+                try:
+                    name_str: str = name.decode("utf-8").lower()
+                    value_str: str = value.decode("utf-8")
+                except Exception:
+                    continue
```
mnajafian-nv, line 164:
```suggestion
                except Exception:
                    logger.debug("Failed to decode WebSocket header, skipping", exc_info=True)
                    continue

```

mnajafian-nv: All three header extraction methods silently discard exceptions on header decode. This hides encoding bugs and makes auth failures hard to diagnose. At minimum, log at debug level.

packages/nvidia_nat_core/src/nat/data_models/api_server.py
```diff
@@ -109,13 +109,6 @@ class TextContent(BaseModel):
     text: str = "default"
 
 
-class Security(BaseModel):
-    model_config = ConfigDict(extra="forbid")
-
-    api_key: SerializableSecretStr = Field(default="default")
-    token: SerializableSecretStr = Field(default="default")
```
mnajafian-nv, line 116: The Security model and the security field on WebSocketUserMessage and WebSocketUserInteractionResponseMessage are removed. These models use extra="forbid", so any client currently sending a security block in their WebSocket messages will get a validation error. Even if "nobody used it," it was part of the public schema. The WebSocket docs or release notes should mention this removal.

mnajafian-nv (changes requested):
## Review Summary
Great work, the design direction is solid, stateless identity resolution, deterministic UUID v5 IDs, and the WebSocket `auth_message` flow fills a real gap. Test coverage is thorough. However, there are a few blocking items.
### Breaking Changes Need Migration Notes
- JWT claim precedence changed from `name > email > preferred_username > sub` to `sub > email > preferred_username` (`name` dropped entirely). Existing users with both `name` and `sub` in their JWTs will get a different `user_id`.
- User ID format changed from raw claim strings to UUID v5 : all existing per-user state becomes orphaned.
- `security` field removed from WebSocket messages (`extra="forbid"` means clients sending it will error).
Release notes should document these three changes and the state migration impact.
### Bugs (see inline comments)
- `identity_claim` ignores typed fields when `claims` dict is empty
- `_from_auth_payload` silently returns `None` on unknown payload type
- Context var leak in `session()` if `extract_user_from_connection` raises before `try`
- `_user_id` not preserved across handler reconnect
### Documentation
- `verify_signature=False` trust model should be a `:::{warning}`, not a `:::{note}` :  the resolved `user_id` gates per-user state access
- Basic Auth password-change = new `user_id` should be documented

packages/nvidia_nat_core/src/nat/front_ends/fastapi/message_handler.py
ericevans-nv: Unlike HTTP where each request is stateless and the user_id must be resolved every time to route to the correct per-user workflow, a WebSocket connection is inherently session-scoped. Each user_message starts a new workflow run, and _user_id is read at that point to create the session via session_manager.session(user_id=self._user_id). On reconnect, _restore_execution_state only activates when a conversation_id matches an already-running workflow — it copies the reference to that running task, not a new session. The reconnected handler resumes the in-flight workflow; it does not accept new user_message requests that would need _user_id. Once the workflow completes, the handler is cleaned up. There is no code path where the restored handler calls _run_workflow again, so _user_id is never needed after reconnect.

mnajafian-nv (approved):
LGTM. Thank you for addressing the comments.

PR #1856, opened 2026-04-09
fix(memmachein): preserve conversation message order in `add_items`

### LINKED ISSUE #1855: Flaky test: test_conversation_messages_preserved_in_order races on asyncio.gather ordering
## Summary

`packages/nvidia_nat_memmachine/tests/test_memmachine_api_calls.py::TestDataTransformation::test_conversation_messages_preserved_in_order` is a flaky test that fails non-deterministically under CI load (observed on Python 3.13, amd64).

## Root Cause

`MemMachineEditor.add_items` wraps each conversation message in `asyncio.to_thread(add_memory)` and dispatches all tasks concurrently via `asyncio.gather(*tasks)` ([`memmachine_editor.py:176-202`](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/packages/nvidia_nat_memmachine/src/nat/plugins/memmachine/memmachine_editor.py#L176-L202)). `asyncio.gather` makes no ordering guarantee — whichever thread completes first records its call first in the spy's list.

The test then asserts by **index position**:

```python
assert add_calls[0]['kwargs']['content'] == "First message"
assert add_calls[1]['kwargs']['content'] == "Second message"   # ← can be "Third message"
assert add_calls[2]['kwargs']['content'] == "Third message"
```

Under a loaded thread pool (Python 3.13 CI), thread 3 can complete before thread 2, causing `add_calls[1]` to hold `"Third message"` instead of `"Second message"`.

## Observed failure

```
FAILED packages/nvidia_nat_memmachine/tests/test_memmachine_api_calls.py::TestDataTransformation::test_conversation_messages_preserved_in_order
AssertionError: assert 'Third message' == 'Second message'
```

## Fix

The parallel dispatch in `add_items` is intentional for performance. The test should assert by **content lookup** (the same pattern already used in `test_add_conversation_calls_add_with_correct_parameters`) rather than by insertion index.

Proposed fix: replace positional index assertions with content-based lookups that verify all three messages are present with correct roles, regardless of completion order.

fede-kamel:
## Summary

- Fixes a race condition in \`MemMachineEditor.add_items\` that caused non-deterministic ordering of conversation messages and flaky CI failures on Python 3.13

## Root Cause

\`add_items\` wrapped each conversation message in \`asyncio.to_thread(add_memory)\` and dispatched all tasks concurrently via \`asyncio.gather(*tasks)\`. Thread pool tasks complete in nondeterministic order, so the API spy recorded calls by completion order rather than insertion order — causing \`test_conversation_messages_preserved_in_order\` to flip assertions under CI load.

## Fix

Refactored \`add_items\` to introduce an inner \`add_item\` coroutine per \`MemoryItem\`. Within that coroutine, conversation messages are \`await\`ed sequentially via \`asyncio.to_thread\`, preserving chronological order. Multiple \`MemoryItem\`s are still dispatched concurrently via \`asyncio.gather\` — so there is no performance regression on batch inserts.

\`\`\`python
async def add_item(memory_item: MemoryItem) -> None:
    ...
    for msg in conversation:
        await asyncio.to_thread(add_memory)   # sequential within one item

if items:
    await asyncio.gather(*(add_item(item) for item in items))  # concurrent across items
\`\`\`

The index-based assertions in the test are preserved as-is — they are now correct because the implementation guarantees order.

## Test plan

- [x] \`test_conversation_messages_preserved_in_order\` passes 50 consecutive runs on Python 3.11, 3.12, and 3.13 locally
- [x] Full test suite: 37 passed, 6 skipped (integration) on all three Python versions

Closes #1855

## Summary by CodeRabbit

* **Refactor**
  * Updated memory item upload processing to preserve the sequential order of conversation messages while maintaining concurrent processing of multiple memory items.

### DIFF
--- a/packages/nvidia_nat_memmachine/src/nat/plugins/memmachine/memmachine_editor.py
+++ b/packages/nvidia_nat_memmachine/src/nat/plugins/memmachine/memmachine_editor.py
@@ -118,6 +118,7 @@ class MemMachineEditor(MemoryEditor):
         """
 
         async def add_item(memory_item: MemoryItem) -> None:
+            """Upload a single MemoryItem, adding conversation messages sequentially."""
             # Make a copy of metadata to avoid modifying the original
             item_meta = memory_item.metadata.copy() if memory_item.metadata else {}
             conversation = memory_item.conversation
--- a/packages/nvidia_nat_memmachine/src/nat/plugins/memmachine/memmachine_editor.py
+++ b/packages/nvidia_nat_memmachine/src/nat/plugins/memmachine/memmachine_editor.py
@@ -111,11 +111,13 @@ class MemMachineEditor(MemoryEditor):
         Each MemoryItem is translated and uploaded through the MemMachine API.
 
         All memories are added to both episodic and semantic memory types.
-        """
-        # Run synchronous operations in thread pool to make them async
-        tasks = []
 
-        for memory_item in items:
+        Conversation messages within a single MemoryItem are added sequentially to
+        preserve chronological order. Separate MemoryItems (and non-conversation
+        memories) are still dispatched concurrently via asyncio.gather.
+        """
+
+        async def add_item(memory_item: MemoryItem) -> None:
             # Make a copy of metadata to avoid modifying the original
             item_meta = memory_item.metadata.copy() if memory_item.metadata else {}
             conversation = memory_item.conversation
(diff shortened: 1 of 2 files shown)


fede-kamel:
@willkill07 Could you add the appropriate labels (suggesting `bug` + `non-breaking`) and trigger CI when you get a chance? Thanks!

willkill07:
/ok to test [be60b01](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1856/commits/be60b01b43bb9edfe855d3ff3c13eeb76bf8c340)

willkill07 (approved):

fede-kamel:
@willkill07 seems like I got lucky here with CI :)

PR #1863, opened 2026-04-11
Filter AIMessage state updates from streaming output

MylesShannon:
## Description
LangGraph's stream_mode="messages" emits both AIMessageChunk (incremental tokens) and AIMessage (final state update) from the agent node. The _stream_fn was accepting both via isinstance(msg, (AIMessage, AIMessageChunk)), causing the full accumulated response to be emitted as a final chunk after all the individual tokens had already been streamed. Clients saw the complete response duplicated at the end of the SSE stream.

Filter to only AIMessageChunk so the state update is excluded.

Adds a regression test that confirms AIMessage objects are emitted by the graph stream (the duplicate source) and that filtering to AIMessageChunk excludes them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed streaming behavior so only chunked assistant message pieces are emitted during chunked streaming, preventing duplicate full assistant content from appearing in streamed agent responses.

* **Tests**
  * Added a regression test that validates chunk-only streaming excludes completed assistant messages, ensuring no duplicated prior assistant text in streamed output.


dagardner-nv (approved):
LGTM, feel free to merge once CI passes

PR #1864, opened 2026-04-12
fix: preserve tool_calls in _stream_llm for native tool calling

### LINKED ISSUE #1865: ReAct Agent: _stream_llm drops tool_calls when use_native_tool_calling=True
### Version

v1.6.0

_No response_

### Describe the bug.

When `use_native_tool_calling=True` is set on `ReActAgentWorkflowConfig`, the agent binds tools correctly via `bind_tools()` and the LLM returns `tool_calls` in its streamed response. However, `_stream_llm` in `base.py` reconstructs the `AIMessage` from content parts only:

```
content_parts = []
async for event in runnable.astream(inputs, config=self._runnable_config):
    content_parts.append(event.content)
return AIMessage(content="".join(content_parts), additional_kwargs=additional_kwargs)
```

Best I can tell, `tool_calls` and `tool_call_chunks` from the streamed chunks are never collected. This means the returned `AIMessage` always has `tool_calls=[]`, and so the native tool call check in `agent_node` (line 278) never triggers. The agent falls through to text parsing, which either fails with `ReActAgentParsingFailedError`, or treats the response as a final text answer.

```shell
1. Configure a ReAct agent with `use_native_tool_calling=True` and an LLM that supports tool calling
2. Send an interaction that should trigger the tool call
3. The agent returns a text response instead of calling the tool
4. Debug logging confirms "Binding tools to LLM for native tool calling" fires at init, but "Native tool call detected" never fires at runtime
```

### Relevant log output

_No response_

### Other/Misc.

The "Tool Calling Agent" already has the fix — PR #1717 added `_chunk_to_message` which accumulates chunks via LangChain's + operator (preserving `tool_call_chunks`) and reconstructs the wire format via `convert_to_openai_messages`. This same fix needs to be applied to `_stream_llm` in `base.py` so the ReAct agent benefits as well.

### Code of Conduct

getglad:
`_stream_llm` reconstructed `AIMessage` from `content` parts only, dropping `tool_calls` from streamed chunks. This broke `use_native_tool_calling` on the ReAct agent - the agent treated every response as a final text answer because `output_message.tool_calls` was always empty.

Use `LangChain` chunk accumulation (chunk + chunk) which preserves `tool_call_chunks`, then convert via `_chunk_to_message`. Move `_chunk_to_message` from `tool_calling_agent` to `base` so both agent types share the implementation.

## Description

Closes https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues/1865

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming message reconstruction to consistently preserve tool-call data.
  * Restored correct behavior when no stream chunks are produced by returning an empty message.
  * Standardized streamed-message aggregation to avoid lost or malformed metadata.

* **Tests**
  * Added streaming tests covering tool-call preservation, empty-stream handling, and new chunk-based stream shapes.


dagardner-nv:
@getglad thanks for the PR.
A few things:
* This repo requires all commits to be signed-off with the git `-s`/`--signoff` flag. Click on the DCO bot in the CI run for instructions on applying this after the fact.

getglad:
@dagardner-nv - the DCO should be handled now. Let me know if I can help address anything else.

dagardner-nv (approved):

PR #1876, opened 2026-04-16
fix(tool): stop leaking exception details in chat_completion error response

ColinM-sys:
## Summary
The `_chat_completion` error handler in `packages/nvidia_nat_core/src/nat/tool/chat_completion.py` concatenated `str(e)` from the caught exception into the user-facing response string (`"...Error: {str(e)}"`). Since the response is returned over the network to API callers, that leaked internal details — stack frame class names, DB driver messages revealing schema, HTTP client errors revealing endpoint URLs / key prefixes, and file paths.

Move the exception detail to server-side logging (`logger.exception`) and return only the user-safe apology + query echo. Operators triage via logs; callers no longer see internal state.

## What changed
- `packages/nvidia_nat_core/src/nat/tool/chat_completion.py`
  - Added module-level `logger = logging.getLogger(__name__)`
  - Replaced `Error: {str(e)}` suffix with server-side `logger.exception("chat completion failed")`
  - Replaced silent `except Exception: pass` on the last-message extraction with `logger.exception(...)` for observability

## CWE
CWE-209 — Information Exposure Through an Error Message

## Why this matters
Caller-visible error responses in an LLM / agent pipeline often get echoed into chat UIs, logged downstream, or stored in conversation history. Leaking database schemas, endpoint URLs, or file paths in those surfaces is a small but real hardening gap. Happy-path output is unchanged.

## Test plan
- [x] Logger pattern matches NAT core convention (module-level `logger = logging.getLogger(__name__)`)
- [x] Happy-path response unchanged
- [x] Regression tests added covering error-path sanitization and server-side logging


packages/nvidia_nat_core/src/nat/tool/chat_completion.py
```diff
@@ -114,17 +118,25 @@ async def _chat_completion(chat_request_or_message: ChatRequestOrMessage) -> Cha
...
             last_content = ""
             try:
                 msg = GlobalTypeConverter.get().convert(chat_request_or_message, to_type=ChatRequest)
                 if msg.messages:
                     last = msg.messages[-1].content
                     last_content = last if isinstance(last, str) else str(last)
             except Exception:
-                pass
+                logger.exception("failed to extract last user message for error response")
+            del e  # explicitly drop the reference; do not leak it into the response
```
dagardner-nv: This isn't needed, just let `e` fall out of scope

packages/nvidia_nat_core/tests/nat/tools/test_chat_completion_error_response.py
```diff
@@ -0,0 +1,131 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
dagardner-nv:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

packages/nvidia_nat_core/src/nat/tool/chat_completion.py
ColinM-sys: Good catch, removed in cb4b3cbc.

packages/nvidia_nat_core/tests/nat/tools/test_chat_completion_error_response.py
ColinM-sys: Applied in cb4b3cbc.

dagardner-nv:
@ColinM-sys tests are failing, please run tests locally, refer to `docs/source/resources/contributing/testing/running-tests.md` if needed.

dagardner-nv (approved):

dagardner-nv:
@ColinM-sys thanks for the contribution!

PR #1877, opened 2026-04-16
fix(a2a): reject non-localhost bind without server_auth by default

ColinM-sys:
## Summary
\`A2AFrontEndConfig.validate_security_configuration\` previously **warned but proceeded** when \`host\` was non-localhost and \`server_auth\` was \`None\`. A warning is easy to miss in production and the resulting deployment exposes an unauthenticated A2A endpoint to the network.

Promote the check from a log warning to a validator error. Operators who legitimately run A2A behind an external auth layer (ingress, service mesh, mTLS, network policy) can opt in with a new \`allow_unauthenticated_network_bind: bool\` field, which defaults to \`False\`.

## CWE
CWE-306 — Missing Authentication for Critical Function

## Behavior change matrix

| host                     | server_auth | allow_unauthenticated_network_bind | result       |
|--------------------------|-------------|------------------------------------|--------------|
| localhost/127.0.0.1/::1  | any         | any                                | ✅ allow     |
| non-localhost            | configured  | any                                | ✅ allow     |
| non-localhost            | None        | \`True\`                           | ✅ allow + warn |
| non-localhost            | None        | \`False\` (default)                | ❌ reject    |

## Why this matters
LLM-driven agent endpoints exposed without authentication are a common mistake and a high-consequence one — any caller can invoke the agent, consume resources, and trigger downstream tool calls. Making the insecure case require an explicit acknowledgement turns the deployment-time misconfiguration from silent into loud.

## Escape hatch
Deployments that *do* have authentication out-of-process (Istio, nginx-ingress with OAuth2 proxy, mTLS, cluster-local-only service, etc.) can set:

\`\`\`yaml
front_end:
  _type: a2a
  host: 0.0.0.0
  allow_unauthenticated_network_bind: true
\`\`\`

## Files changed
- \`packages/nvidia_nat_a2a/src/nat/plugins/a2a/server/front_end_config.py\` — add \`allow_unauthenticated_network_bind\` field; validator now raises \`ValueError\` when misconfigured
- \`packages/nvidia_nat_a2a/tests/server/test_front_end_config_security.py\` — new regression tests for all four rows of the matrix above
- \`packages/nvidia_nat_a2a/tests/server/test_agent_card_generation.py\` — update the existing 0.0.0.0 public-URL test to set the new flag so coverage keeps passing

## Test plan
- [x] \`localhost\`, \`127.0.0.1\`, \`::1\` without auth: allowed (parametrized test)
- [x] \`0.0.0.0\` without auth, no opt-in: rejected with clear error message
- [x] named host without auth, no opt-in: rejected
- [x] \`0.0.0.0\` without auth, \`allow_unauthenticated_network_bind=True\`: allowed + warning logged
- [x] Existing \`test_agent_card_url_uses_public_base_url_when_configured\` updated to set the flag
- [x] Syntax-checked (couldn't run pytest locally — my env is missing \`pytest-asyncio\`; CI will run it)

## Summary by CodeRabbit

* **Security**
  * Strengthened network-binding validation: non-localhost hosts now require authentication unless explicitly allowed; misconfigurations that previously only warned are rejected by default.

* **New Features**
  * Added an opt-in flag to permit unauthenticated network binding (default: false); enabling it permits non-localhost binds without auth and is logged.

* **Tests**
  * Added tests covering localhost vs non-localhost binding, opt-in behavior, logging, and validation error cases.

### DIFF
--- a/packages/nvidia_nat_a2a/src/nat/plugins/a2a/server/front_end_config.py
+++ b/packages/nvidia_nat_a2a/src/nat/plugins/a2a/server/front_end_config.py
@@ -121,17 +121,46 @@ class A2AFrontEndConfig(FrontEndBaseConfig, name="a2a"):
                      "opaque token validation (via RFC 7662 introspection)."),
     )
 
+    # Explicit acknowledgement required to bind to a non-localhost interface without server_auth.
+    # Default is False: the validator below will reject the config rather than silently warning
+    # and proceeding. Operators who intentionally run A2A on a non-localhost interface behind
+    # an external auth layer (ingress, service mesh, network policy) can set this to True to
+    # override the check.
+    allow_unauthenticated_network_bind: bool = Field(
+        default=False,
+        description=("Explicit acknowledgement to allow binding to a non-localhost interface without "
+                     "server_auth configured. Required when host is not localhost/127.0.0.1/::1 and "
+                     "server_auth is None. Set this only when an external authentication layer "
+                     "(ingress, service mesh, mTLS, network policy) protects the A2A endpoint. "
+                     "Default: False (reject the config to prevent accidental unauthenticated exposure)."),
+    )
+
     @model_validator(mode="after")
     def validate_security_configuration(self):
-        """Validate security configuration to prevent accidental misconfigurations."""
-        # Check if server is bound to a non-localhost interface without authentication
+        """Validate security configuration to prevent accidental misconfigurations.
+
+        Reject any configuration that binds the A2A server to a non-localhost interface
+        without either (a) server_auth configured or (b) an explicit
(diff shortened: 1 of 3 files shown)


AnuradhaKaruppiah:
@ColinM-sys Is there a GH issue for this that was used for discussion with maintainers?

dagardner-nv:
Thanks for creating a NAT PR, however this change would caused a behavioral change, and we do document authentication. Please open an issue for this feature so that we can discuss the requirements and possible solution.

ColinM-sys:
Closing in favor of #1883 to discuss the design first per your feedback. The behavioral-change concern is fair — happy to revisit once we've landed on an approach that balances hardening with backward compatibility. Thanks for the direction, @dagardner-nv.

PR #1878, opened 2026-04-16
fix(fastapi): reject CORS wildcard origin combined with allow_credentials

ColinM-sys:
## Summary
`FastApiFrontEndConfig.CrossOriginResourceSharing` silently accepted `allow_origins=["*"]` with `allow_credentials=True`. Per the CORS spec (Fetch §3.2, MDN CORS), browsers reject responses that set `Access-Control-Allow-Origin: *` when credentials are allowed — so the combination produces a server that looks permissive but is broken in every browser, and signals operator intent that would be unsafe if the browser check were ever bypassed.

Promote the check to a `model_validator` that rejects the combination at config-load time with an actionable error pointing at the two correct fixes.

## CWE
CWE-942 — Permissive Cross-domain Policy with Untrusted Domains

## Behavior change

| allow_origins | allow_origin_regex | allow_credentials | result |
|---|---|---|---|
| `["*"]` | — | `True` | ❌ rejected |
| `["*", "https://app.ex.com"]` | — | `True` | ❌ rejected |
| `["*"]` | — | `False` | ✅ allowed |
| `["https://app.ex.com"]` | — | `True` | ✅ allowed (intended) |
| — | `.*` / `.+` / `^.*$` / `^.+$` | `True` | ❌ rejected |
| — | `^https://(app\|admin)\.example\.com$` | `True` | ✅ allowed |

## Why this matters
The combination is cheap to misconfigure and carries production risk. Failing fast at config-load time prevents shipping a service that is simultaneously (a) broken in browsers and (b) a vulnerability waiting to surface via any non-browser client that doesn't enforce the browser-side check.

## Files changed
- `packages/nvidia_nat_core/src/nat/front_ends/fastapi/fastapi_front_end_config.py` — add `model_validator` on `CrossOriginResourceSharing`
- `packages/nvidia_nat_core/tests/nat/front_ends/fastapi/test_fastapi_front_end_config.py` — new `TestCorsWildcardCredentialsValidator` class

## Test plan
- [x] Wildcard origin + credentials → rejected
- [x] Wildcard origin alone → allowed
- [x] Explicit origin allowlist + credentials → allowed (intended pattern)
- [x] Match-everything regex (`.*`, `.+`, `^.*$`, `^.+$`) + credentials → rejected (parametrized)
- [x] Bounded regex + credentials → allowed
- [x] Mixed wildcard + explicit list + credentials → rejected

### DIFF
--- a/packages/nvidia_nat_core/src/nat/front_ends/fastapi/fastapi_front_end_config.py
+++ b/packages/nvidia_nat_core/src/nat/front_ends/fastapi/fastapi_front_end_config.py
@@ -24,6 +24,7 @@ from pydantic import BaseModel
 from pydantic import Field
 from pydantic import SerializeAsAny
 from pydantic import field_validator
+from pydantic import model_validator
 
 from nat.data_models.component_ref import ObjectStoreRef
 from nat.data_models.evaluator import EvalInputItem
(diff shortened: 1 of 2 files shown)


packages/nvidia_nat_core/src/nat/front_ends/fastapi/fastapi_front_end_config.py
```diff
@@ -218,6 +219,36 @@ class CrossOriginResourceSharing(BaseModel):
...
+            Per the CORS spec (Fetch §3.2, MDN CORS), a response cannot set
+            Access-Control-Allow-Origin: * while Access-Control-Allow-Credentials
+            is true — browsers refuse such responses. Accepting the combination
+            in the config silently produces a server that looks permissive but
+            is broken in every browser, AND signals operator intent that would
+            be genuinely unsafe if the browser check were ever bypassed
+            (CVE-class misconfiguration). Fail fast with a clear error.
+
+            CWE-942 — Permissive Cross-domain Policy with Untrusted Domains.
+            """
```
dagardner-nv, line 235: I would slim down this doc-string since:
1) It's not a public method
2) The same information is repeated in the body of the value error that is raised.

```suggestion
            """Reject configs that pair wildcard origins with credentials."""
```

packages/nvidia_nat_core/src/nat/front_ends/fastapi/fastapi_front_end_config.py
```diff
@@ -218,6 +219,36 @@ class CrossOriginResourceSharing(BaseModel):
...
+                        "'allow_origins=[\"*\"]'. Per the CORS spec, browsers reject responses that "
+                        "set Access-Control-Allow-Origin: * when credentials are allowed. "
+                        "Specify an explicit list of trusted origins (e.g. ['https://app.example.com']) "
+                        "or disable credentials.")
+                if self.allow_origin_regex and self.allow_origin_regex.strip() in (".*", ".+", "^.*$", "^.+$"):
+                    raise ValueError(
+                        "CORS misconfiguration: 'allow_credentials=True' cannot be combined with a "
+                        "match-everything 'allow_origin_regex' ({!r}). Use an explicit regex that "
+                        "matches only trusted origins (e.g. r'^https://(app|admin)\\.example\\.com$') "
+                        "or disable credentials.".format(self.allow_origin_regex))
```
dagardner-nv, line 249: I'm not sure if this is correct as `allow_origin_regex` appears to be specific to FastAPI's `CORSMiddleware` implementation.

dagardner-nv (commented):
The code in question is really a YAML wrapper around FastAPI's [CORSMiddleware](https://fastapi.tiangolo.com/tutorial/cors/#use-corsmiddleware).

My question then is, if this is a bug, is it not a FastAPI bug?
Given that this code works as-is:
```py
from fastapi import FastAPI
from fastapi.middleware.cors import CORSMiddleware

app = FastAPI()
app.add_middleware(
    CORSMiddleware,
    allow_origins=['*'],
    allow_credentials=True)
```

ColinM-sys:
Fair point. FastAPI intentionally allows this combination and NAT is ultimately just passing the config through to CORSMiddleware. The right fix is either upstream in FastAPI or in documentation rather than a hard guardrail at the NAT config layer that could break legitimate server-to-server use cases where CORS does not apply. Closing this one.

PR #1879, opened 2026-04-16
fix(cache-middleware): bound cache size and enforce safe fuzzy threshold

ColinM-sys:
## Summary
\`CacheMiddleware\` previously had two hardening gaps:

1. **Unbounded cache** (dict-backed) — sustained unique inputs grew the cache without limit, a memory-exhaustion DoS surface, and combined with fuzzy matching a long-lived surface for cross-request confusion.
2. **Unrestricted \`similarity_threshold\`** — any float in [0, 1] was accepted. Low values make cache poisoning trivial: a crafted input whose difflib ratio exceeds the threshold against a legitimate cached key hijacks the legitimate user's response. At 0.5, most short/structured inputs collide on first try.

## What changed
- Back the cache with an \`OrderedDict\`. Evict the oldest entry (LRU) when the cache exceeds \`max_entries\`. A cache hit \`move_to_end()\`s the entry so frequently-useful results are preferred on eviction.
- Add \`max_entries\` parameter (default 1024, must be positive int).
- Reject \`similarity_threshold < 0.85\` at construct time with an actionable error. 1.0 remains recommended.
- Non-numeric threshold values now raise instead of failing later.

## CWE
- CWE-400 — Uncontrolled Resource Consumption
- CWE-345 — Insufficient Verification of Data Authenticity (cache-key confusion)

## Breaking change
Configs that pass \`similarity_threshold < 0.85\` will fail at construct time. Intentional — any such value is a cache-poisoning foot-gun. The error message points at the two safe choices:
  - Use \`1.0\` (exact match, recommended)
  - Use \`>= 0.85\` (fuzzy, reasonable edit tolerance)

## Test plan
Existing tests at sub-0.85 thresholds were updated to 0.85/0.9 so they keep exercising fuzzy mode without tripping the new floor.

New coverage in two classes:

**\`TestSimilarityThresholdFloor\`**:
- [x] Below-floor values (0.0, 0.3, 0.5, 0.7, 0.84) rejected with "outside the safe range"
- [x] At-or-above-floor values (0.85, 0.9, 0.95, 1.0) accepted
- [x] Threshold > 1.0 rejected
- [x] Non-numeric threshold rejected with "must be a number"

**\`TestMaxEntriesLruEviction\`**:
- [x] Default \`max_entries\` is positive
- [x] Zero or negative \`max_entries\` rejected
- [x] Cache stays at \`max_entries\` after sustained unique inserts (size = 3 after 10 inserts with max_entries=3; newest 3 retained)
- [x] Cache hit promotes entry to MRU: A→B→C, hit A, insert D, B (oldest) is evicted; A and D remain

Syntax-checked locally; pytest deferred to CI (my env is missing \`pytest-asyncio\`).

## Summary by CodeRabbit

* **New Features**
  * Cache gains a configurable max-entries limit (default 1024) with LRU eviction and promotion on hits.
  * Configuration exposes a new max_entries field.

* **Improvements**
  * Fuzzy-match logic refined for more consistent matching and clearer insertion logging.
  * Various dependency and packaging constraints updated.

* **Tests**
  * Added tests covering LRU eviction and adjusted similarity-threshold expectations.

* **Documentation**
  * Docs and notebooks updated to version 1.7 links.

* **Chores**
  * CI/workflow pins and build/script logging simplified; small container image and tooling updates.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/middleware/cache/cache_middleware.py
+++ b/packages/nvidia_nat_core/src/nat/middleware/cache/cache_middleware.py
@@ -30,6 +30,7 @@ from __future__ import annotations
 
 import json
 import logging
+from collections import OrderedDict
 from collections.abc import AsyncIterator
 from typing import Any
 
@@ -42,6 +43,21 @@ from nat.middleware.function_middleware import FunctionMiddlewareContext
 from nat.middleware.middleware import InvocationContext
 
 logger = logging.getLogger(__name__)
+
+# Lower bound on fuzzy-match similarity to reduce the cache-poisoning surface.
+# A threshold below this makes it trivial to craft an input that is "similar
+# enough" to a legitimate user's cached key to hijack their response (for the
+# current process, which is how the in-memory cache is scoped). 0.85 is the
+# smallest value that we're comfortable shipping as an unconditional default;
+# operators with strict needs should use 1.0 (exact match only).
+_MIN_FUZZY_THRESHOLD = 0.85
+
+# Default bound on cache size. The previous implementation used an unbounded
+# dict which, under sustained unique input, grew without limit — a memory-
+# exhaustion DoS and, combined with fuzzy matching, a long-lived surface for
+# cross-request confusion. OrderedDict-backed LRU evicts the oldest entry
+# when the cache exceeds this bound.
+_DEFAULT_MAX_CACHE_ENTRIES = 1024
 
 
 class CacheMiddleware(FunctionMiddleware):
@@ -67,19 +83,51 @@ class CacheMiddleware(FunctionMiddleware):
             computation.
     """
 
-    def __init__(self, *, enabled_mode: str, similarity_threshold: float) -> None:
+    def __init__(
+        self,
+        *,
+        enabled_mode: str,
+        similarity_threshold: float,
+        max_entries: int = _DEFAULT_MAX_CACHE_ENTRIES,
+    ) -> None:
         """Initialize the cache middleware.
 
         Args:
             enabled_mode: Either "always" or "eval". If "eval", only caches
                 when Context.is_evaluating is True.
-            similarity_threshold: Similarity threshold between 0 and 1.
-                If 1.0, performs exact matching. Otherwise uses fuzzy matching.
+            similarity_threshold: Similarity threshold in [_MIN_FUZZY_THRESHOLD, 1.0].
+                If 1.0, performs exact matching. Lower values enable difflib-
+                based fuzzy matching. Values below _MIN_FUZZY_THRESHOLD are
+                rejected to prevent cache-poisoning where a crafted input
+                collides with a legitimate user's cached key.
+            max_entries: Maximum number of cache entries. When exceeded, the
+                oldest entry is evicted (LRU). Defaults to
+                _DEFAULT_MAX_CACHE_ENTRIES. Must be >= 1.
+
+        Raises:
+            ValueError: If similarity_threshold is outside [_MIN_FUZZY_THRESHOLD, 1.0]
+                or max_entries is not a positive integer.
         """
+        if not isinstance(similarity_threshold, (int, float)):
+            raise ValueError(
+                f"similarity_threshold must be a number, got {type(similarity_threshold).__name__}")
+        if similarity_threshold < _MIN_FUZZY_THRESHOLD or similarity_threshold > 1.0:
+            raise ValueError(
+                f"similarity_threshold={similarity_threshold} is outside the safe range "
+                f"[{_MIN_FUZZY_THRESHOLD}, 1.0]. Lower values make cache-poisoning trivial — "
+                "a crafted input can collide with a legitimate user's cached key. Use 1.0 "
+                "for exact matching (recommended), or a value >= "
+                f"{_MIN_FUZZY_THRESHOLD} for fuzzy matching.")
+        if not isinstance(max_entries, int) or max_entries < 1:
+            raise ValueError(f"max_entries must be a positive integer, got {max_entries!r}")
+
         super().__init__(is_final=True)
         self._enabled_mode = enabled_mode
         self._similarity_threshold = similarity_threshold
-        self._cache: dict[str, Any] = {}
+        # OrderedDict gives O(1) LRU: move_to_end() on hit, popitem(last=False)
+        # to evict the oldest when we exceed max_entries.
+        self._cache: OrderedDict[str, Any] = OrderedDict()
+        self._max_entries = max_entries
 
     # ==================== Abstract Method Implementations ====================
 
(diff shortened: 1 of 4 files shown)


dagardner-nv:
@ColinM-sys I think this PR has some value, but it really needs a GitHub issue to be opened to discuss the design of these changes (no need to close the PR). 

At a first glance, I'm not sure that `similarity_threshold` requires a minimum value but instead a mention of the potential consequences are of a lower value.

I don't believe the type-checking around the `similarity_threshold` attribute is required as this should be enforced by Pydantic, as the intended use-case is to construct this class from the `CacheMiddlewareConfig` class, anyone attempting to construct an instance of `CacheMiddleware` another way is on their own.

ColinM-sys:
Opened #1888 for the design discussion. Makes sense on the type-checking since Pydantic handles that at the config layer. On the threshold floor I think there is still value in at least documenting the cache poisoning risk of lower values even if a hard minimum is not appropriate for a core library. Happy to revise once the issue discussion settles on a direction.

packages/nvidia_nat_core/src/nat/middleware/cache/cache_middleware.py
```diff
@@ -199,20 +216,31 @@ async def function_middleware_invoke(self,
...
         logger.debug("Cache miss for function %s", context.name)
         result = await call_next(*args, **kwargs)
 
-        # Phase 3: Postprocess - cache the result for future use
+        # Phase 3: Postprocess - cache the result for future use. Enforce the
+        # LRU bound BEFORE insert so the new entry always lands in a cache of
+        # size <= max_entries, preventing unbounded memory growth (DoS).
         self._cache[input_str] = result
-        logger.debug("Cached result for function %s", context.name)
+        self._cache.move_to_end(input_str)
```
dagardner-nv, line 237: Is this call to `self._cache.move_to_end(input_str)` needed? If it was just inserted on line 236, shouldn't it already be at the end?

packages/nvidia_nat_core/src/nat/middleware/cache/cache_middleware.py
```diff
@@ -44,6 +45,14 @@
 logger = logging.getLogger(__name__)
 
 
+# Default bound on cache size. The previous implementation used an unbounded
+# dict which, under sustained unique input, grew without limit — a memory-
+# exhaustion DoS and, combined with fuzzy matching, a long-lived surface for
+# cross-request confusion. OrderedDict-backed LRU evicts the oldest entry
+# when the cache exceeds this bound.
+_DEFAULT_MAX_CACHE_ENTRIES = 1024
```
dagardner-nv, line 53: I don't think this is needed as a global const, the default value should be in the config class, since we only construct `CacheMiddleware` from the config object, the constructor similarly doesn't need a default value.

packages/nvidia_nat_core/src/nat/middleware/cache/cache_middleware_config.py
```diff
@@ -31,14 +32,35 @@ class CacheMiddlewareConfig(FunctionMiddlewareBaseConfig, name="cache"):
...
+        le=1.0,
+        description=(
+            "Similarity threshold in [0, 1.0]. Use 1.0 for exact matching (recommended). "
+            "Lower values enable fuzzy matching via difflib; note that difflib is quadratic "
+            "in the worst case, so large caches with low thresholds may have a performance "
+            "cost. Values near 0 increase the risk of cache collisions where different "
+            "inputs return the same cached response."),
+    )
+    max_entries: int = Field(
+        default=_DEFAULT_MAX_CACHE_ENTRIES,
```
dagardner-nv, line 61:
```suggestion
        default=1024,
```

Remove the const global, 1024 isn't a number that needs to be a const.

packages/nvidia_nat_core/tests/nat/middleware/test_cache_middleware.py
```diff
@@ -62,7 +62,7 @@ def test_default_initialization(self):
 
     def test_custom_initialization(self):
         """Test custom initialization."""
-        middleware = CacheMiddleware(enabled_mode="always", similarity_threshold=0.8)
+        middleware = CacheMiddleware(enabled_mode="always", similarity_threshold=0.9)
```
dagardner-nv, line 65: Revert, I don't think this is needed

packages/nvidia_nat_core/tests/nat/middleware/test_cache_middleware.py
```diff
@@ -109,7 +109,7 @@ async def mock_next_call(*args, **kwargs):
 
     async def test_fuzzy_match_caching(self, middleware_context):
         """Test fuzzy matching with similarity_threshold < 1.0."""
-        middleware = CacheMiddleware(enabled_mode="always", similarity_threshold=0.8)
+        middleware = CacheMiddleware(enabled_mode="always", similarity_threshold=0.9)
```
dagardner-nv, line 112: Same, not needed

packages/nvidia_nat_core/tests/nat/middleware/test_cache_middleware.py
```diff
@@ -267,8 +267,7 @@ async def mock_next_call(*args, **kwargs):
 
     def test_similarity_computation_for_different_thresholds(self):
         """Test similarity computation for different thresholds."""
-        # This is more of a unit test for the similarity logic
-        middleware = CacheMiddleware(enabled_mode="always", similarity_threshold=0.5)
+        middleware = CacheMiddleware(enabled_mode="always", similarity_threshold=0.9)
```
dagardner-nv, line 270: Revert

packages/nvidia_nat_core/tests/nat/middleware/test_cache_middleware.py
```diff
@@ -278,14 +277,14 @@ def test_similarity_computation_for_different_thresholds(self):
...
-        # Very similar
+        # Very similar (one char shorter, ~0.95 ratio)
         assert middleware._find_similar_key("hello worl") == test_key  # noqa
         # Too different - use a completely different string
         assert middleware._find_similar_key("xyz123abc") is None  # noqa
 
     async def test_multiple_similar_entries(self, middleware_context):
         """Test behavior with multiple similar cached entries."""
-        middleware = CacheMiddleware(enabled_mode="always", similarity_threshold=0.7)
+        middleware = CacheMiddleware(enabled_mode="always", similarity_threshold=0.85)
```
dagardner-nv, line 287: Revert

willkill07:
@ColinM-sys do you still have interest in updating this PR?

ColinM-sys:
Sorry for the delay — I thought this had been picked back up after incorporating all of dagardner's feedback from #1888 in the April 20 push. All changes are already in (dropped threshold floor to 0, Pydantic handles bounds at config layer, replaced SequenceMatcher loop with get_close_matches, max_entries wired through register.py). Ready for review whenever.

dagardner-nv:
> Sorry for the delay — I thought this had been picked back up after incorporating all of dagardner's feedback from #1888 in the April 20 push. All changes are already in (dropped threshold floor to 0, Pydantic handles bounds at config layer, replaced SequenceMatcher loop with get_close_matches, max_entries wired through register.py). Ready for review whenever.

I left review feedback on April 21st.

willkill07:
/ok to test [b57ef41](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1879/commits/b57ef41c3050af197add4854a98ce99a7b9071a1)

willkill07:
/ok to test [bed0a8b](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1879/commits/bed0a8bafd5fbdd688e44f85184d1b332b076e0b)

willkill07 (approved):

PR #1880, opened 2026-04-16
fix(github-files-tool): add optional allowed_repos allowlist

ColinM-sys:
## Summary
\`github_files_tool\` previously accepted any URL matching \`github.com/<org>/<repo>/blob/...\` — meaning a prompt-injected agent (or a user following a crafted prompt) could have the tool fetch arbitrary public source from any repo on GitHub. That reaches into:

- Secrets / misconfigured credentials committed to public forks
- Private-research PoC repos that were later made public
- Any cross-org code the agent's operator never intended to expose

Add an optional \`allowed_repos\` field to \`GithubFilesGroupConfig\`. Entries are either \`'org/repo'\` (exact) or \`'org/*'\` (wildcard), matched case-insensitively. When set, URLs outside the allowlist return a clear \`"not in the configured allowed_repos list"\` error instead of fetching. **Default remains \`None\` (permissive) for backward compatibility**, with the docstring on the field calling out the security implication so operators can decide.

## CWE
CWE-284 — Improper Access Control (broad exposure scope on an external data source)

## Behavior matrix

| \`allowed_repos\` | URL fetched? |
|---|---|
| \`None\` (default) | ✅ any github.com repo (backward compat) |
| \`[]\` | ❌ deny-all (explicit) |
| \`["NVIDIA/NeMo-Agent-Toolkit"]\` | ✅ exact match only |
| \`["NVIDIA/*"]\` | ✅ any repo under NVIDIA org |
| \`["NVIDIA/*", "apache/spark"]\` | ✅ OR-combined |

Matching is case-insensitive; malformed entries (missing \`/\`, non-strings, empty) are ignored without false-positive matches.

## Why not change the default to strict
Flipping the default breaks every existing config that uses the tool without setting the field. Opt-in allowlist with a prominent docstring is the backward-compatible hardening: operators who want scope get it, operators who rely on current behavior are unchanged.

## Files changed
- \`packages/nvidia_nat_core/src/nat/tool/github_tools.py\`
  - \`GithubFilesGroupConfig.allowed_repos: list[str] | None\` (default None)
  - New \`_repo_path_is_allowed\` helper
  - \`get()\` closure returns a scoped-out error message when the repo_path fails the allowlist
- \`packages/nvidia_nat_core/tests/nat/tools/test_github_files_tool_allowlist.py\` — direct unit coverage of the helper + config shape (12 parametrized cases)

## Test plan
- [x] \`allowed_repos=None\` → any repo allowed
- [x] Explicit \`"org/repo"\` and \`"org/*"\` wildcard matches (parametrized)
- [x] Case-insensitive matching (operator can write \`"NVIDIA/Repo"\`, URL can use lowercase)
- [x] Multiple allowlist entries OR-combined
- [x] Malformed entries (\`""\`, \`None\`, non-strings, missing \`/\`) ignored without false-positives
- [x] Malformed \`repo_path\` rejected
- [x] Empty list = explicit deny-all
- [x] Syntax-checked locally; pytest deferred to CI

End-to-end coverage of the \`get()\` closure needs a mocked httpx transport and is out of scope for this defensive hardening PR.

## Summary by CodeRabbit

* **New Features**
  * Added an optional repository allowlist to restrict GitHub file access to specific repositories, supporting exact repo matches and org-level wildcards (e.g., org/*) with case-insensitive matching.

* **Tests**
  * Added comprehensive tests for allowlist behavior: defaults, wildcard and exact-match semantics, case-insensitivity, defensive handling of malformed entries, and empty-allowlist blocking.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/tool/github_tools.py
+++ b/packages/nvidia_nat_core/src/nat/tool/github_tools.py
@@ -372,6 +372,45 @@ async def github_tool(config: GithubGroupConfig, _builder: Builder):
 
 class GithubFilesGroupConfig(FunctionBaseConfig, name="github_files_tool"):
     timeout: int = Field(default=5, description="Timeout in seconds for HTTP requests")
+    allowed_repos: list[str] | None = Field(
+        default=None,
+        description=(
+            "Optional allowlist of GitHub repositories the tool may fetch from. "
+            "Each entry is either a full 'org/repo' or an 'org/*' wildcard. "
+            "When None (the default), any public GitHub repository is reachable — "
+            "which means a prompt-injected agent can read arbitrary public source "
+            "code (secrets in public forks, private PoC repos, etc.). Operators who "
+            "need a specific scope (e.g., only their own org) should set this to "
+            "['my-org/*'] or an explicit repo list. Matching is case-insensitive."),
+    )
+
+
+def _repo_path_is_allowed(repo_path: str, allowed_repos: list[str] | None) -> bool:
+    """Check whether a repo_path ('org/repo') matches the configured allowlist.
+
+    Returns True if:
+      - allowed_repos is None (no restriction — backward compatible default), or
+      - repo_path matches an explicit 'org/repo' entry (case-insensitive), or
+      - repo_path matches an 'org/*' wildcard entry (case-insensitive).
+
+    Returns False for any malformed entry or non-matching repo_path.
+    """
+    if allowed_repos is None:
+        return True
+    if not repo_path or "/" not in repo_path:
+        return False
+    repo_lower = repo_path.lower()
+    repo_org = repo_lower.split("/", 1)[0]
+    for entry in allowed_repos:
+        if not isinstance(entry, str) or "/" not in entry:
+            continue
+        entry_lower = entry.lower()
+        if entry_lower == repo_lower:
+            return True
+        entry_org, entry_repo = entry_lower.split("/", 1)
+        if entry_repo == "*" and entry_org == repo_org:
+            return True
+    return False
 
 
 @register_function(config_type=GithubFilesGroupConfig)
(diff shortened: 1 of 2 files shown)


packages/nvidia_nat_core/src/nat/tool/github_tools.py
```diff
@@ -372,6 +372,80 @@ async def commit(updated_file_list: GithubCommitCodeModelList) -> str:
...
+    if allowed_repos is None:
+        return True
+
+    split = _split_org_repo(repo_path)
+    if split is None:
+        return False
+    repo_org_lower = split[0].lower()
+    repo_lower = f"{split[0]}/{split[1]}".lower()
+
+    for entry in allowed_repos:
```
dagardner-nv, line 434: I would parse these in `github_files_tool` when the tool is first registered since they do not change across invokations.

packages/nvidia_nat_core/src/nat/tool/github_tools.py
```diff
@@ -372,6 +372,80 @@ async def commit(updated_file_list: GithubCommitCodeModelList) -> str:
...
+            # — ignore rather than accidentally matching.
+            continue
+        entry_org_lower = entry_split[0].lower()
+        entry_repo_lower = entry_split[1].lower()
+        # Exact match: 'org/repo'
+        if f"{entry_org_lower}/{entry_repo_lower}" == repo_lower:
+            return True
+        # Wildcard match: 'org/*'
+        if entry_repo_lower == "*" and entry_org_lower == repo_org_lower:
+            return True
```
dagardner-nv, line 447: In both cases you need to `entry_org_lower == repo_org_lower` (implied on line 443).


```suggestion
        if entry_org_lower == repo_org_lower:
            if entry_repo_lower == "*" or entry_repo_lower == repo_lower:
                return True
        return False
```

dagardner-nv (commented):
Overall I think this has some merit. I don't think the current implementation represents a security risk unless the tool is running with additional github credentials which grant increased github visibility. 

Otherwise unless I'm missing something, displaying the content of a publicly available file on github to the user which they could easily fetch themselves doesn't represent a security vulnerability.

ColinM-sys:
Thanks for the review @dagardner-nv. You're right — the security framing was overstated. Without a concrete operator use case driving the need for repo restriction, the practical value is limited. Closing this one.

PR #1897, opened 2026-04-27
feat(cli): add consent-gated runtime telemetry for NAT CLI commands

yczhang-nv:
## Description

Adds aggregate, anonymous CLI-usage telemetry to the nat Python CLI. Posture is consent-gated: 

- A first-run prompt asks the user whether to allow telemetry, defaulting to yes (Enter accepts, type n to opt out)
- The decision is persisted across invocations.
- Non-interactive sessions (CI / cron / piped / stderr-captured) never prompt and default OFF
- `NAT_TELEMETRY_ENABLED` env var overrides both.

### Consent flow
Order of precedence for whether telemetry is active:

- `NAT_TELEMETRY_ENABLED` env var (any value — 1/true/yes enables, anything else disables). Sets the answer for this shell session and bypasses both the persisted file and the prompt.
- Persisted decision at ~/.config/nat/telemetry.toml (TOML, written once when the user answers the first-run prompt; can be re-set anytime via nat configure telemetry --enable | --disable). The persisted record carries a prompt_version; if we materially change what the prompt discloses (new collected field, new endpoint), bumping PROMPT_VERSION forces a re-prompt for users who consented under the old language.
- Interactive prompt (stdin, stdout, and stderr must all be TTYs) on first run when neither of the above is set. Default answer is yes: pressing Enter (or typing y/yes) → enabled. Typing n/no, garbage, hitting EOF, or Ctrl-C → disabled. The prompt explicitly lists what is collected and what is not.
- Default OFF in non-interactive sessions (no TTY on any of the three standard streams → no opportunity to ask → no data sent).

### CLI subcommand: nat configure telemetry
Three modes:

```bash
nat configure telemetry --enable — persist consent as enabled
nat configure telemetry --disable — persist consent as disabled
nat configure telemetry --status (default) — show current effective state and its source
```
The status command flags when an env var is overriding the persisted decision, so users aren't surprised when their persisted choice doesn't take effect. The first-run prompt is suppressed when running any of these three subcommands — they exist to manage the consent decision, so prompting first would break the read-only contract of --status and conflict with explicit --disable / --enable.

## Summary by CodeRabbit

* **New Features**
  * Opt-in runtime telemetry for the NAT CLI (disabled by default).
  * New `nat configure telemetry` command: enable, disable, or show status.
  * First-run consent prompt for interactive sessions and env-var override via NAT_TELEMETRY_ENABLED.
  * Telemetry emission for CLI invocations, with local debugging (stdout endpoint) and dry-run mode.

* **Documentation**
  * README updated with telemetry details, consent flow, available controls, and data collection clarity.

* **Tests**
  * Extensive unit and integration tests covering consent flow, CLI integration, event payloads, handler, and configuration.


packages/nvidia_nat_core/src/nat/utils/telemetry/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+
+See `nat.utils.telemetry.config` for environment variables (defaults: opt-out
+disabled, telemetry enabled). To turn telemetry off for a single invocation:
+
+    NAT_TELEMETRY_ENABLED=false nat run ...
+
+To inspect events locally without making network calls:
+
+    NAT_TELEMETRY_ENDPOINT=stdout nat run ...
+"""
```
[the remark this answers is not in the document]
yczhang-nv: The PR desc is out-dated. Now it defaults to enable with interactive env.

packages/nvidia_nat_core/src/nat/cli/commands/configure/telemetry.py
```diff
@@ -0,0 +1,111 @@
...
+        _persist_and_verify(ConsentState.DISABLED)
+        click.echo(f"Telemetry disabled. Persisted to {_consent_file_path()}.")
+        _warn_on_env_var_override()
+        return
+
+    # Status (default)
+    env_value = os.getenv("NAT_TELEMETRY_ENABLED")
+    persisted = read_persisted_consent()
+    if env_value is not None:
+        active = env_value.strip().lower() in ("1", "true", "yes")
```
dagardner-nv: This looks very similar to the check in `packages/nvidia_nat_core/src/nat/utils/telemetry/consent.py::resolve_initial_consent`

Maybe move to cli/utils.py?

packages/nvidia_nat_core/src/nat/utils/telemetry/handler.py
```diff
@@ -0,0 +1,302 @@
...
+    async def __aenter__(self) -> NATTelemetryHandler:
+        await self.astart()
+        return self
+
+    async def __aexit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> None:
+        await self.astop()
+
+    # ------------------------------------------------------------- internals
+
+    def _run_sync(self, coro: Any) -> Any:
```
dagardner-nv, line 183: Since we currently depend on `nest-asyncio2` and explicitly apply it in `packages/nvidia_nat_core/src/nat/cli/entrypoint.py`, I'm not sure if this is needed (unless it's being invoked prior to that).

dagardner-nv (approved):
Approved, left a few comments, be sure to verify the notebooks still work.

q) For tests which shell out and run `nat run` or `nat serve`? Is telemetry disabled as they are non-interactive, or does it take into account if the developer has a `telemetry.toml` file? What about in CI where this file won't exist?

packages/nvidia_nat_core/src/nat/cli/commands/configure/telemetry.py
yczhang-nv: Good point, I extracted the helper function into `packages/nvidia_nat_core/src/nat/utils/telemetry/consent.py` since I feel more natural‎

packages/nvidia_nat_core/src/nat/utils/telemetry/handler.py
yczhang-nv: That's reasonable, but I'll suggest keep this ~10 lines of code. This part of code is mostly mirrored from the implementation in [DataDesigner repo](https://github.com/NVIDIA-NeMo/DataDesigner/blob/main/packages/data-designer-engine/src/data_designer/engine/models/telemetry.py) so just want to keep the implementation aligned. Also, other modules can import and use it without going through `nat.cli.entrypoint`, like we already done in `tests/nat/utils/telemetry/test_handler.py`.

yczhang-nv:
@dagardner-nv Thanks for the comments! I double-checked the notebooks are still working. 

> q) For tests which shell out and run nat run or nat serve? Is telemetry disabled as they are non-interactive, or does it take into account if the developer has a telemetry.toml file? What about in CI where this file won't exist?

That's a good catch. I turned off telemetry for tests in the session-level `conftest` to ensure running unit tests won't emit any telemetry data. In CI the telemetry is turned off since it's not interactive env.

PR #1902, opened 2026-04-28
chore: dependency cleanup - drops optuna chain and openinference from core

bbednarski9:
## Description

This PR reduces the base dependency surface of `nvidia-nat-core` and moves dependency ownership closer to the packages that actually use those integrations.

27MB saved for nvidia-nat-core installs with an additional 35MB for nvidia-nat-eval  without full

- Removes `optuna` and `openinference-semantic-conventions` from `nvidia-nat-core`.
- Adds `openinference-semantic-conventions` to `nvidia-nat-opentelemetry`, where it is used.
- Makes `aioboto3` optional for `nvidia-nat-eval` under `full` and `test` extras.
- Lazy-loads eval S3 upload dependencies so importing eval runtime does not require `aioboto3`.
- Replaces the core Optuna `Trial` type dependency with a local protocol for Optuna-compatible trial objects.
- Moves `test_evaluate_callbacks.py` into the eval package and updates it to use non-deprecated `nat.plugins.eval` imports.
- Updates `uv.lock` files to reflect the dependency boundary changes.

## Testing

```bash
git diff --cached --check
uv run pytest packages/nvidia_nat_core/tests/nat/data_models/test_optimizable.py packages/nvidia_nat_config_optimizer/tests/test_parameter_optimizer.py packages/nvidia_nat_config_optimizer/tests/test_optimizable_utils.py packages nvidia_nat_eval/tests/eval/test_dependency_guidance.py packages/nvidia_nat_eval/tests/eval/test_evaluate_callbacks.py packages/nvidia_nat_eval/tests/eval/utils/test_output_uploader.py packages/nvidia_nat_eval/tests/eval/test_evaluate.py -q
uv run pytest packages/nvidia_nat_opentelemetry/tests/observability -q
``` 

Focused eval/core/optimizer suite: 111 passed
OpenTelemetry suite: 93 passed

⚠️ Breaking change: aioboto3 moved from base to [full] extra

nvidia-nat-eval previously installed aioboto3 (transitively boto3/botocore) as a base dependency. This branch moves it to the [full] extra, removing ~35 MB and 22 transitive packages (incl. the full aiohttp async stack and botocore AWS service models) from the default install.

Who is affected

Users running pip install nvidia-nat-eval (without [full]) whose eval configs use the dataset.s3.* download path. On develop this worked out of the box; after this change, boto3 is not present and the S3 download will raise ModuleNotFoundError at runtime.

What stays the same

- Local datasets are unaffected; signed-URL and S3 remote datasets require optional full eval dependencies if their transport dependency is not otherwise installed.
- The [full] extra still pulls everything previously available.
- nvidia-nat[eval] users (the recommended top-level install) are unaffected.

Migration
```text
pip install 'nvidia-nat-eval[full]'
# or
uv pip install 'nvidia-nat[eval]'
```

The runtime now raises a clear ModuleNotFoundError with this install hint at the point of failure, rather than failing at import time.

Other dep moves in this PR (no behavioral break expected)

- optuna removed from nvidia-nat-core base; nvidia-nat-config-optimizer keeps it pinned. Core was refactored to use a Protocol so no top-level optuna import remains.
- openinference-semantic-conventions moved from nvidia-nat-core to nvidia-nat-opentelemetry base. Core no longer imports it.

These moves only affect users who were importing those packages directly while depending solely on nvidia-nat-core — a pattern that effectively requires they already had the package installed by other means.

Closes

## Summary by CodeRabbit

* **Chores**
  * Optional dependencies reorganized: S3 and related libraries moved to optional "full"/test installs to reduce default install size
  * Lazy-loading of optional libraries for S3 and signed-URL downloads to avoid import-time failures and provide clear pip-install hints when missing
  * Core search API typing decoupled from a specific optimization backend for broader compatibility

* **Tests**
  * New and expanded tests covering missing-optional-dependency behavior and search type-hint resolution


packages/nvidia_nat_core/src/nat/data_models/optimizable.py
```diff
@@ -78,8 +90,8 @@ def validate_search_space_parameters(self):
 
         return self
 
-    # Helper for Optuna Trials
-    def suggest(self, trial: Trial, name: str):
+    # Helper for Optuna-compatible trial objects
+    def suggest(self, trial: _TrialLike, name: str):
```
dagardner-nv, line 95: Why not just enclose Trial in quotes. 

I think we can still import optuna under a `if typing.TYPE_CHECKING:` block

```suggestion
    def suggest(self, trial: "Trial", name: str):
```

packages/nvidia_nat_eval/src/nat/plugins/eval/utils/output_uploader.py
```diff
@@ -20,15 +20,26 @@
...
-from botocore.exceptions import NoCredentialsError
 from tqdm import tqdm
 
 from nat.data_models.common import get_secret_value
 from nat.data_models.evaluate_config import EvalOutputConfig
 
 logger = logging.getLogger(__name__)
 
+S3_UPLOAD_INSTALL_HINT = ("S3 output upload requires optional dependencies that are not installed. "
+                          "Install with: pip install \"nvidia-nat-eval[full]\"")
```
dagardner-nv:
```suggestion
                          "Install with: uv pip install \"nvidia-nat-eval[full]\"")
```

dagardner-nv (approved):

packages/nvidia_nat_core/src/nat/data_models/optimizable.py
bbednarski9: Good point. Updated to use a TYPE_CHECKING import plus the quoted Trial annotation so core avoids a runtime optuna import while keeping the public type tied to Optuna.

packages/nvidia_nat_eval/pyproject.toml
```diff
@@ -55,17 +55,19 @@ dependencies = [
   # Keep package version constraints as open as possible to avoid conflicts with other packages. Always define a minimum
   # version when adding a new package. If unsure, default to using `~=` instead of `==`. Does not apply to nvidia-nat packages.
   # Keep sorted!!!
```
AnuradhaKaruppiah, line 57:
```suggestion
# Keep package version constraints as open as possible to avoid conflicts with other packages. Always define a minimum
# version when adding a new package. If unsure, default to using `~=` instead of `==`. Does not apply to nvidia-nat packages.
# Keep sorted!!!
#
# Keep the base nvidia-nat-eval dependencies minimal and ATIF-only. Dependencies required for running evaluation via
# the `nat eval` CLI must be added to the `full` optional dependency group.
```

packages/nvidia_nat_eval/pyproject.toml
```diff
@@ -55,17 +55,19 @@ dependencies = [
...
   "nvidia-nat-atif == {version}",
+  "tqdm~=4.67",
 ]
 
 [tool.setuptools_dynamic_dependencies.optional-dependencies]
 full = [
+  "aioboto3>=11.0.0",
   "datasets~=4.4",
   "nvidia-nat-core == {version}",
   "openpyxl~=3.1",
```
AnuradhaKaruppiah, line 70:
```suggestion
  "openpyxl~=3.1",
  "tqdm~=4.67",
```

packages/nvidia_nat_eval/pyproject.toml
```diff
@@ -55,17 +55,19 @@ dependencies = [
   # Keep package version constraints as open as possible to avoid conflicts with other packages. Always define a minimum
   # version when adding a new package. If unsure, default to using `~=` instead of `==`. Does not apply to nvidia-nat packages.
   # Keep sorted!!!
-  "aioboto3>=11.0.0",
   "nvidia-nat-atif == {version}",
+  "tqdm~=4.67",
```
AnuradhaKaruppiah:
```suggestion
```

AnuradhaKaruppiah (approved):

packages/nvidia_nat_core/src/nat/data_models/optimizable.py
```diff
@@ -29,6 +29,9 @@
 
 T = TypeVar("T", int, float, bool, str)
 
+if TYPE_CHECKING:
+    from optuna import Trial
+
```
willkill07: Shouldn't this be leveraging the protocol rather than attempting to import from optuna?

packages/nvidia_nat_eval/src/nat/plugins/eval/utils/output_uploader.py
bbednarski9: thx, addressed

willkill07 (approved):

packages/nvidia_nat_core/src/nat/data_models/optimizable.py
bbednarski9: Alright im going with a compromise. Returning to the Protocol model as this is how i first designed it. But adding a unit test jusstttt to be safe and make sure our protocol doesnt drift

PR #1851, opened 2026-04-08
Feature: token streaming support for ReAct Agent

thepatrickchin:
## Description

Previously, ReAct's `_stream_llm` explicitly passed `self._runnable_config` to the LLM, overriding LangGraph's injected runtime config and dropping its streaming callbacks. This was fixed by merging both configs so LangGraph can observe tokens.

The ReAct Agent was also missing  `_stream_fn`, so the framework had no streaming entry point and always fell back to the non-streaming path. This was fixed by propagating the injected node config through `agent_node` down to `_stream_llm`, and registering a `_stream_fn` via `FunctionInfo.create` that yields ChatResponseChunk tokens from `graph.astream`

Closes #1850 

## By Submitting this PR I confirm:
- I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/docs/source/resources/contributing/index.md).
- We require that all contributors "sign-off" on their commits. This certifies that the contribution is your original work, or you have rights to submit it under the same license, or a compatible license.
  - Any contribution which contains commits that are not Signed-Off will not be accepted.
- When the PR is ready for review, new or existing tests cover these changes.
- When the PR is ready for review, the documentation is up to date with these changes.

## Summary by CodeRabbit

* **New Features**
  * Workflows now provide real-time streaming responses alongside existing single-response handling; agents accept an optional runtime config for streaming.

* **Improvements**
  * Streaming merges runtime configuration so callbacks and settings are applied consistently.
  * Prompt enforces exact "Final Answer:" formatting.
  * Streaming buffers until the final-answer marker, filters tool-call chunks, and reports recursion/streaming errors more clearly.

* **Tests**
  * Added tests for token streaming, final-answer buffering, filtering, error handling, and config propagation.

PR #1978, opened 2026-05-20
fix(server): emit canonical {"value": ...} envelope from /generate/full

matthewgrossman:
## Description

Restores the wire format that `/generate/full` emits for the workflow's output, which silently regressed between 1.6 and 1.7.

**The contract** (documented by the eval client `nat.plugins.eval.runtime.remote_workflow.py:79` and pinned by `test_remote_evaluate.py`'s server fixture):

```
data: {"value": "<final answer>"}
```

**What broke.** PR #1851 (`feat: token streaming support for ReAct Agent`) added a `_stream_fn` to `react_agent` that yields `ChatResponseChunk` (OpenAI shape). Once a stream_fn is registered, `generate_streaming_response_full` takes the streaming branch and wraps each chunk in `ResponsePayloadOutput`, whose `get_stream_data()` dumps the chunk's full OpenAI envelope. There is no top-level `value` field, so the eval client's `chunk_data.get("value")` returns `None` and every eval scores 0. The producer (`react_agent`) and consumer (eval client) ship in the same NAT release and disagree on the wire shape.

`tool_calling_agent` is exposed to the same regression for the same reason — both yield `ChatResponseChunk` from their `_stream_fn`. The fix is in the shared `ResponsePayloadOutput.get_stream_data()`, so both code paths get covered.

**The fix.** `ResponsePayloadOutput.get_stream_data()` now normalizes any payload — string, primitive, `ChatResponseChunk`, `ChatResponse`, other `BaseModel` — into the canonical `data: {"value": "<str>"}\n\n` envelope. Scoped to `/generate/full`; `/v1/chat/completions` is unaffected because that path yields `ChatResponseChunk` directly through `ResponseBaseModelOutput.get_stream_data()`, never wrapped in `ResponsePayloadOutput`. WebSocket consumers do their own payload coercion in `MessageValidator.convert_data_to_message_content()` and don't call `get_stream_data()` either.

**Tests.** Parametrized unit tests pin the wire format per payload type. A new integration test in `test_remote_evaluate.py` round-trips real `ResponsePayloadOutput` lines through the real `EvaluationRemoteWorkflowHandler`, so a future change that desynchronizes producer and consumer fails CI rather than silently scoring zero on every eval.

## How to verify

The bug surface is two pure functions on NAT data models — the producer (`ResponsePayloadOutput.get_stream_data`) and the consumer (`chunk_data.get("value")` in the eval client). You can reproduce both the bug and the fix with no FastAPI server, no LLM, and no external services:

```python
import json
from nat.data_models.api_server import ResponsePayloadOutput, ChatResponseChunk

# What react_agent's _stream_fn yields after PR #1851:
chunk = ChatResponseChunk.create_streaming_chunk("21")

# What /generate/full puts on the wire:
sse_line = ResponsePayloadOutput(payload=chunk).get_stream_data()

# What nat.plugins.eval.runtime.remote_workflow extracts:
data = json.loads(sse_line[len("data: "):-2])
print("eval client extracts:", repr(data.get("value")))
```

Expected output:

| | Output |
|---|---|
| On `release/1.7` (without this PR) | `eval client extracts: None` |
| With this PR applied | `eval client extracts: '21'` |

The full `nvidia_nat_core` and `nvidia_nat_eval` test suites pass on this branch with no regressions, including the new parametrized unit tests and the producer/consumer integration test added here.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/data_models/api_server.py
+++ b/packages/nvidia_nat_core/src/nat/data_models/api_server.py
@@ -15,6 +15,7 @@
 
 import abc
 import datetime
+import json
 import typing
 import uuid
 from abc import abstractmethod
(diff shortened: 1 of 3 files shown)


matthewgrossman:
closing in favor of https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/1980, branched off of release/1.7

PR #1982, opened 2026-05-20
Generate a list of NIM models used in workflow YAML files in `pre-commit`

dagardner-nv:
## Description
* Generate a list of models used in example workflows in CI located at `ci/.nim_models_used.json`
* model list file is maintained by a new `pre-commit` hook
* Allows for simple tracking of the NIM models we use.
* Update `ci/scripts/model_health_check.py` to use logging rather than print statements
* Ignore the locally hosted nim yaml file since the model names are different than their build.nvidia.com counter-parts.
* Add logic to ignore symlink paths and files not tracked by git

## Release Notes

* **New Features**
  * Model deprecation tracking now integrated into health checks and Slack reporting
  * Dry-run mode now generates structured JSON reports for model catalogs

* **Improvements**
  * Configurable logging for health check operations
  * Enhanced model catalog tracking and organization

* **Chores**
  * Python runtime requirement updated to 3.11

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NeMo-Agent-Toolkit/pull/1982?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


yczhang-nv (approved):
LGTM

mnajafian-nv (commented):
This direction makes sense to me. A checked-in generated manifest is a good way to give build.nvidia.com a stable source of truth and help NAT get notified before model removals break examples or docs.

A few things I think we should tighten before this becomes the consumed contract:

- I would keep `ci/.nim_models_used.json` if build.nvidia.com has already wired to that path. At this point, path stability matters more than renaming.
- I like keeping `llms` and `embedders` split. The replacement path and blast radius are different enough that the distinction is useful.
- I would align the count field with the downstream contract. The Slack example and Smita’s note mention `num_configs`, but the PR currently generates `usage_count`. Since the value is counting config files, I think this should be `num_configs` unless build.nvidia.com explicitly prefers `usage_count`.
- I would add a top-level `schema_version` now. Once another team is parsing this file, we need a simple way to evolve it without breaking their cron/parser.
- I would state the inventory scope explicitly. From the implementation, this appears to cover model references discovered from repo YAML configs, not arbitrary mentions in docs/READMEs. That scope is reasonable, but it should be clear so nobody assumes full documentation coverage.
- The pre-commit hook seems appropriate because it runs in dry-run mode and generates from repo-local files. It should stay deterministic and offline. No build.nvidia.com/NIM service calls in pre-commit.
- I would not require config paths in v1 if build.nvidia.com only needs model/count, but paths would be the next useful field for NAT-side remediation.

Once the file contract is aligned with build.nvidia.com, I’m happy to do a code-level pass.

dagardner-nv:
@mnajafian-nv I renamed `usage_count` to `num_configs`, and added a top-level `schema_version` entry.

> I would keep `ci/.nim_models_used.json` if build.nvidia.com has already wired to that path. At this point, path stability matters more than renaming.

This is still open for change, they just signed-off on the rough idea, and are waiting for the PR to be merged prior to writing any code.

> I would not require config paths in v1 if build.nvidia.com only needs model/count, but paths would be the next useful field for NAT-side remediation.

I looked at adding these, but it just feels redundant since it's easy to search the repo for a model name and quickly find any references to it.

mnajafian-nv:
Thank you, the changes addressed the main contract points I was worried about. `schema_version` + `num_configs` makes the file much clearer for build.nvidia.com to consume.

One small hardening item: since this is an external contract, I think the generator should always emit the same top-level shape in dry-run/output-json mode, even when one category is empty or no NIM models are found.

Ideally the output is always:

{
  "schema_version": "1.0",
  "llms": [],
  "embedders": []
}

That avoids stale checked-in data if model references are removed and keeps the downstream parser simple.

dagardner-nv:
> Thank you, the changes addressed the main contract points I was worried about. `schema_version` + `num_configs` makes the file much clearer for build.nvidia.com to consume.
> 
> One small hardening item: since this is an external contract, I think the generator should always emit the same top-level shape in dry-run/output-json mode, even when one category is empty or no NIM models are found.
> 
> Ideally the output is always:
> 
> { "schema_version": "1.0", "llms": [], "embedders": [] }
> 
> That avoids stale checked-in data if model references are removed and keeps the downstream parser simple.

Done

ci/scripts/model_health_check.py
```diff
@@ -73,26 +94,28 @@ def find_nim_models(examples_dir: Path) -> tuple[dict[str, list[str]], dict[str,
     for pattern in ("*.yml", "*.yaml"):
         config_paths.extend(examples_dir.rglob(pattern))
 
+    git_files = get_git_files()
+
     for config_path in sorted(config_paths):
+        relative_path = str(config_path.resolve().relative_to(REPO))
```
mnajafian-nv: This path normalization can still fail before the new “skip untracked config” logic gets a chance to run. `--examples-dir` is a public CLI option, and if it points outside `REPO` then `config_path.resolve().relative_to(REPO)` raises `ValueError`. The same applies to any YAML path that resolves outside the repo via a symlink.

The previous implementation guarded `relative_to(REPO)` and fell back to a string path for reporting. Since this PR is intentionally tightening path handling, I think we should keep that robustness here and skip/log paths that cannot be expressed relative to `REPO`.

Suggested shape:

```python
try:
    relative_path = str(config_path.resolve().relative_to(REPO))
except ValueError:
    _logger.warning("Skipping config outside repository root: %s", config_path)
    continue
```

.pre-commit-config.yaml
```diff
@@ -61,5 +61,15 @@ repos:
         args: ["-q", "--config", "ci/markdown-link-check-config.json"]
         exclude: "^(packages/nvidia_nat_core/src/nat/meta/pypi\\.md|CHANGELOG\\.md)$"
 
+  - repo: local
+    hooks:
+      - id: generate-models-list
+        name: Generate Models List
+        entry: python ./ci/scripts/model_health_check.py --log-level=WARNING --dry-run --output-json ci/.nim_models_used.json
```
mnajafian-nv, line 68: This local hook inherits `default_language_version: python3`, so the interpreter is whatever `python3` resolves to on the contributor machine. The repo requires Python >=3.11, but that is not guaranteed by this hook configuration. On machines where `python3` is older, the hook can fail before doing any useful work because `model_health_check.py` uses modern type syntax such as `Path | None`.

Can we pin this hook to a supported interpreter, for example `language_version: python3.11`, or route it through the repo-managed environment? CI is fine because it runs in the expected image; this is about keeping local pre-commit reliable.

.pre-commit-config.yaml
```diff
@@ -61,5 +61,15 @@ repos:
...
+        language: python
+        files: "^examples/.*\\.(yml|yaml)$"
```
mnajafian-nv: This hook only runs when an example YAML file changes. Since `ci/.nim_models_used.json` is generated by `ci/scripts/model_health_check.py`, changes to the generator itself, `EXCLUDE_YAMLS`, or the JSON schema could land without pre-commit checking that the generated file was refreshed.

Can we include the generator script in this hook trigger as well, or make the hook `always_run: true` if the runtime cost is acceptable? This would keep the generated artifact tied to the logic that produces it.

mnajafian-nv (approved):
Thanks , this is a useful addition and the generated model inventory should make NIM usage easier to track.

I left a few comments focused on local/pre-commit robustness and keeping the generated artifact tied to the generator logic. The main one I’d like fixed before merge is guarding the `relative_to(REPO)` path normalization so custom `--examples-dir` usage or symlinked paths outside the repo do not crash before the skip logic runs.

Approved from my side assuming those comments are addressed before merge.

.pre-commit-config.yaml
dagardner-nv: I updated the files to include the `model_health_check.py` script itself in addition to any yaml files.

mnajafian-nv (approved):
Thanks, the changes addressed my generated-file trigger comment. I also see the path guard and Python 3.11 pre-commit update, so comments look addressed from my side.

PR #1995, opened 2026-05-29
experimental(agent): Add experimental coding-agent adapters with NeMo-Relay telemetry

yczhang-nv:
## Description

Adds experimental NVIDIA NeMo Agent Toolkit workflow adapters for five coding-agent harnesses:

- Claude Code
- Codex
- Cursor
- Hermes
- OpenClaw

Each adapter is packaged as an installable example with a `nat.components` entry point, a primitive workflow `_type`, Relay-enabled configs, Phoenix tracing configs, smoke eval configs, and README guidance.

This also adds an experimental Relay telemetry bridge used by the Relay-backed adapters, plus Phoenix trace screenshots for the verified workflows.

## Summary by CodeRabbit

* **New Features**
  * Added five experimental agent adapter examples: Claude Code, Codex, Cursor, Hermes, and OpenClaw for extended workflow integration scenarios.
  * Introduced relay telemetry bridge for mapping NeMo Relay telemetry events into NAT intermediate steps.
  * Added comprehensive configuration templates and evaluation datasets for each adapter.

* **Documentation**
  * Added setup and usage documentation for all experimental agent adapters with end-to-end workflow examples.

* **Tests**
  * Added telemetry bridge tests covering event mapping, parentage resolution, and file injection.

* **Chores**
  * Updated path checking rules and project dependencies for new experimental examples.


examples/experimental/openclaw_agent_adapter/configs/config-relay-phoenix-eval.yml
```diff
@@ -0,0 +1,57 @@
...
+      dir: ./.tmp/nat/examples/openclaw_agent_adapter/relay_phoenix_eval
+      write_atif_workflow_output: true
+    dataset:
+      _type: json
+      file_path: examples/experimental/openclaw_agent_adapter/data/eval-smoke.json
+  evaluators:
+    avg_workflow_runtime:
+      _type: avg_workflow_runtime
+      max_concurrency: 1
+    avg_num_llm_calls:
```
mnajafian-nv: Small question on the eval surface here: should `avg_num_llm_calls` / `avg_tokens_per_llm_end` be expected to reflect the OpenClaw-internal model activity for this example? I’m asking because this adapter does not import Relay/OpenClaw spans back into the NAT trajectory the way the Claude/Codex/Cursor/Hermes adapters do, and the README also says this path does not import a Relay `events.jsonl` file. From the current code, these evaluators look like they would only see NAT-side trajectory steps, not the OpenClaw-internal LLM spans the example is otherwise showcasing. If that reading is right, it may be worth either dropping these evaluators here or adding a short note about what they do and don’t measure in this example.

examples/experimental/openclaw_agent_adapter/src/nat_openclaw_agent_adapter/register.py
```diff
@@ -0,0 +1,300 @@
...
+    description: str = Field(default="OpenClaw CLI Agent Workflow",
+                             description="The description of this function's use.")
+
+    command: str = Field(default="openclaw", description="OpenClaw CLI command or absolute path.")
+    working_directory: str = Field(default=".", description="Directory used as the OpenClaw subprocess cwd.")
+    agent_id: str = Field(default="main", description="OpenClaw agent id.")
+    session_key: str | None = Field(default="nat-openclaw", description="Optional durable OpenClaw session key.")
+    model: str | None = Field(default=None, description="Optional OpenClaw model reference.")
+    thinking: str | None = Field(default=None, description="Optional OpenClaw thinking level.")
+    local: bool = Field(default=True, description="Force embedded local execution instead of Gateway mode.")
```
mnajafian-nv: Could we make the default here line up with the documented example path? Right now `local` defaults to `true`, which appends `--local` and bypasses Gateway, but the README describes the Relay integration as Gateway-based because the `nemo-relay` plugin runs there. I noticed all of the checked-in sample configs already override this to `local: false`, so I wondered if the default should match that too to make the workflow type less surprising when someone copies it outside the provided config files.

examples/experimental/openclaw_agent_adapter/README.md
```diff
@@ -0,0 +1,309 @@
...
+Number of LLMs: 0
+Number of Embedders: 0
+Number of Memory: 0
+Number of Object Stores: 0
+Number of Retrievers: 0
+Number of TTC Strategies: 0
+Number of Authentication Providers: 0
+
+Workflow Result:
+[pyproject.toml](/Users/yuchenz/Desktop/Work/Project/nemo-agent-toolkit/examples/experimental/openclaw_agent_adapter/pyproject.toml:44) exposes a `nat.components` entry point:
```
mnajafian-nv: Could we swap this absolute `/Users/yuchenz/...` path for a relative path or plain code text? The rest of the example reads nicely as repo-portable documentation, and this stood out as contributor-local output that other readers won’t be able to reproduce.

examples/README.md
```diff
@@ -104,6 +105,13 @@ _Additional information can be found in the [Agents README](./agents/README.md).
 - **[`alert_triage_agent`](advanced_agents/alert_triage_agent/README.md)**: Production-ready intelligent alert triage system using LangGraph that automates system monitoring diagnostics with tools for hardware checks, network connectivity, performance analysis, and generates structured triage reports with root cause categorization **[🟨 Intermediate]**
 - **[`vulnerability_analysis_blueprint`](advanced_agents/vulnerability_analysis_blueprint/README.md)**: Blueprint documentation for vulnerability analysis agents **[🛑 Advanced]**
 
+### Experimental
+- **[`claude_code_agent_adapter`](experimental/claude_code_agent_adapter/README.md)**: Experimental primitive agent workflow backed by the Claude Code Agent SDK with a conservative local smoke-test config. **[🛑 Advanced]**
```
mnajafian-nv: Tiny wording suggestion here: would “CLI-backed” be more accurate than “SDK-backed” for the Claude Code and Codex entries in this section? In the example implementations themselves, both adapters are structured around shelling out to the local `claude` / `codex` commands, so “SDK” set me up for a slightly different integration model than what the code is actually doing.

examples/experimental/openclaw_agent_adapter/configs/config-relay-phoenix-eval.yml
yczhang-nv: That's a good catch. Yes the OpenClaw integration is slightly different than the other four, since we're using the Relay-OpenClaw plugin here. I removed the evaluators as you suggested.

examples/experimental/openclaw_agent_adapter/src/nat_openclaw_agent_adapter/register.py
yczhang-nv: Yes. I updated the default value to `False` to match the documentation.

examples/experimental/openclaw_agent_adapter/README.md
yczhang-nv: Removed my local path from the example

examples/README.md
yczhang-nv: Good catch, the SDK-backed method was no longer there as all examples are using the CLI now. Documentation updated.

examples/experimental/openclaw_agent_adapter/src/nat_openclaw_agent_adapter/register.py
```diff
@@ -0,0 +1,300 @@
...
+                "sure the selected OpenClaw/Codex model is available for the account.")
+    return ""
+
+
+async def _run_openclaw_agent(prompt: str, config: OpenClawAgentWorkflowConfig) -> str:
+    cwd = Path(config.working_directory).resolve()
+    openclaw_prompt = _add_gateway_path_context(prompt, cwd, config)
+    command = _build_openclaw_command(config, openclaw_prompt)
+
+    try:
```
mnajafian-nv, line 249: Non-blocking nit: would it make sense to add the same upfront `working_directory` validation here that the other adapters now have before subprocess launch? It would make the error path more consistent across these examples and give a clearer config error when the workspace path is invalid.

mnajafian-nv (approved):
LGTM with one inline nit :) great work!

Salonijain27 (approved):

examples/experimental/openclaw_agent_adapter/src/nat_openclaw_agent_adapter/register.py
yczhang-nv: Thanks, fixed in the latest commit

PR #1998, opened 2026-06-01
Fix LangSmith integration tests

dagardner-nv:
## Description
* Fix LangSmith ID int->uuid conversions, based upon the existing comments this appears to have been a LangSmith behavior change.
* Don't prefix dataset names with "Benchmark Dataset ", just use the name provided by the user as-is.
* Fix LangSmith integration tests
  - `EvalCallbackManager.on_dataset_loaded` expects a list of `EvalInputItem` not dicts
  - `EvalCallbackManager.on_eval_complete` doesn't create runs, instead associate the `EvalResultItem` with a run by setting the `root_span_id`
* Optionally avoid cleaning up LangSmith generated items (useful for debugging)

## Summary by CodeRabbit

* **Bug Fixes**
  * Run IDs are now derived deterministically and dataset naming uses the raw dataset name for consistency.
  * Prompt repository names are sanitized to produce valid, stable handles.

* **Tests**
  * Rewrote and expanded LangSmith integration tests with async polling, new helpers/fixtures, and coverage for run/feedback linking and prompt cleanup.

* **Chores**
  * CI fixture optionally preserves projects and logs deletion errors.


dagardner-nv:
@coderabbitai review

packages/nvidia_nat_langchain/tests/langsmith/test_langsmith_integration.py
```diff
@@ -92,6 +180,7 @@ async def test_eval_callback_creates_dataset_runs_and_feedback(
                     actual_output="4",
                     scores={"accuracy": 1.0},
                     reasoning={"accuracy": "Exact match"},
+                    root_span_id=run_map['r1'].id.int,
```
mnajafian-nv: Could we make this test cover the per-item association contract more directly?

Right now the assertion below passes as soon as any run has feedback. Would it be better to assert that each expected run ID receives the expected `accuracy` feedback, and optionally that each run is linked to the expected dataset example? That would catch cases where one `EvalResultItem` is dropped or both scores attach to the same LangSmith run.

packages/nvidia_nat_langchain/src/nat/plugins/langchain/langsmith/langsmith_evaluation_callback.py
```diff
@@ -53,11 +54,9 @@ def _humanize_dataset_name(name: str) -> str:
 def _span_id_to_langsmith_run_id(span_id: int) -> str:
     """Derive LangSmith run_id from OTEL span_id.
 
-    LangSmith deterministically maps OTEL span_ids to run UUIDs:
-    the first 8 bytes are zeroed, the last 8 bytes are the span_id.
+    LangSmith deterministically maps OTEL span_ids to run UUIDs.
     """
-    hex_str = format(span_id, "016x")
-    return f"00000000-0000-0000-{hex_str[:4]}-{hex_str[4:]}"
+    return str(uuid.UUID(int=span_id))
```
mnajafian-nv, line 59: Could we clarify the ID contract here?

The PR description mentions LangSmith IDs may appear as UUID objects or integer values, but this callback path only accepts integer `root_span_id` values. That may be the right NAT contract since `EvalResultItem.root_span_id` is currently typed/generated as an `int`.

If UUID objects can reach this callback, should the helper and the eager/retry/fallback guards accept `int | uuid.UUID`? If only integer IDs are valid here, could we make that explicit so future changes do not accidentally rely on UUID-object support?

mnajafian-nv (approved):
Thank you. Approving to unblock, assuming the two inline comments are addressed before merge. The implementation direction looks good; my comments are mainly about making the contract and regression coverage tighter.

packages/nvidia_nat_langchain/src/nat/plugins/langchain/langsmith/langsmith_evaluation_callback.py
dagardner-nv: Currently we're using the `int` constructor argument for `uuid.UUID` which will raise a `TypeError` rather than cast another UUID instance, so if `span_id` was an instance of `UUID` it would raise an exception. 

We could of course add an isinstance check here and handle it, but my first thought was that since this method isn't a part of the public API it wasn't needed.

PR #2000, opened 2026-06-01
Fix running `pytest` from the root of the repo

dagardner-nv:
## Description
* This allows running `pytest` from the root of the repo and testing those packages inluced 
* Redefine helper methods into fixtures

## Summary by CodeRabbit

* **Tests**
  * Migrated test infrastructure to use pytest fixtures for consistent test data provisioning across test suites.

* **Refactor**
  * Consolidated shared test utilities and graph builders into centralized fixtures, eliminating duplicate code across test modules.
  * Reorganized test configuration to reduce redundant test helper definitions.


packages/nvidia_nat_test/src/_utils/nat_app_test_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Shared test helpers for nvidia_nat_app."""
+
+from __future__ import annotations
+
+from nat_app.graph.access import AccessSet
```
mnajafian-nv: Quick package-boundary question: since this helper lives in `nvidia-nat-test` but imports `nat_app`, do we want to keep this helper local to `packages/nvidia_nat_app/tests` instead?

I think the current version works for the `most` test environment because both packages are installed together, but keeping app-specific helpers inside the app test tree may avoid making `nvidia-nat-test` carry an implicit dependency on `nvidia-nat-app`. Not blocking from my side, just wanted to check the intended ownership here.

packages/nvidia_nat_test/src/_utils/nat_app_test_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def build(self, original, result):
+        return result
+
+
+def make_node(
+    name: str,
+    reads: set[str] | None = None,
+    writes: set[str] | None = None,
+    confidence: str = "full",
```
mnajafian-nv: Tiny typing nit: since `NodeAnalysis.confidence` is typed as `Literal["full", "partial", "opaque"]`, should this helper use the same type? That would keep the test helper contract aligned with the production model.

mnajafian-nv (approved):
LGTM! The pytest-root fix makes sense, and moving the no-arg graph builders into fixtures is a cleaner pytest pattern. I left one package-boundary question around the new app-specific helper in `nvidia-nat-test`, but nothing blocking from my side.

packages/nvidia_nat_test/src/_utils/nat_app_test_helpers.py
dagardner-nv: This is a tricky one, `nvidia_nat_app` does depend on `nvidia-nat-test` for tests, but yes, the opposite is not true. 

The main issue is that currently in the `develop` branch there's code in `packages/nvidia_nat_app/tests` that does:
```py
from tests.conftest import MinimalAdapter as _SimpleAdapter
```

Which assumes that there is a top-level `tests` package which is owned by `nvidia_nat_app`, this works when we run `pytest` directly inside the `nvidia_nat_app` directory but not anywhere else. 

The `packages/nvidia_nat_test/src/_utils` is typically where helper code like this would go.

I think with a little more work, I could refactor these as fixtures to avoid this.

packages/nvidia_nat_test/src/_utils/nat_app_test_helpers.py
dagardner-nv: > Tiny typing nit: since NodeAnalysis.confidence is typed as Literal["full", "partial", "opaque"], should this helper use the same type? That would keep the test helper contract aligned with the production model.

I moved the method as-is, but yes I think that could be an improvement

PR #2003, opened 2026-06-04
fix(agent): Fix ReWOO nested evidence placeholder substitution

yczhang-nv:
### Summary

- Recursively replace ReWOO evidence placeholders inside nested dict/list tool inputs before tool execution.
- Preserve evidence output types for exact placeholder matches while retaining string interpolation behavior for partial matches.
- Avoid mutating planner-provided tool inputs during placeholder substitution.

### Why
Before the fix ReWOO agent was sending raw placeholders like `#E1` and `#E2` into dependent tool calls such as `{"numbers": ["#E1", "#E2"]}`, causing schema validation failures.

## Summary by CodeRabbit

* **New Features**
  * Enhanced ReWOO Agent placeholder handling to support nested structures and richer tool inputs.

* **Bug Fixes**
  * Prevented unintended mutation of original input values during placeholder substitution; token-aware replacements avoid false matches.

* **Tests**
  * Added tests validating nested placeholder replacement, immutability of inputs, and correct token matching.

### DIFF
--- a/packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/rewoo_agent/agent.py
+++ b/packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/rewoo_agent/agent.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import asyncio
+import copy
 import json
 import logging
 import re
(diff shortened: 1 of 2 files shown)


packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/rewoo_agent/agent.py
```diff
@@ -208,34 +209,55 @@ def _parse_planner_dependencies(steps: list[ReWOOPlanStep]) -> tuple[dict[str, R
...
 
-        elif isinstance(tool_input, str):
-            tool_input = tool_input.replace(placeholder, str(tool_output))
+        if isinstance(tool_input, str):
+            if tool_input in replacements:
+                return copy.deepcopy(replacements[tool_input])
 
-        else:
-            assert False, f"Unexpected type for tool_input: {type(tool_input)}"
+            for placeholder, tool_output in replacements.items():
```
mnajafian-nv: Small edge case question on the placeholder replacement semantics: should this handle placeholders whose names share prefixes, like `#E1` and `#E10`? Since this loop does plain substring replacement in insertion order, a string such as `"Compare #E10 and #E1"` can replace `#E1` inside `#E10` first and turn it into something like `"Compare <E1-output>0 and <E1-output>"`. Because ReWOO placeholders are sequential, this becomes realistic once a plan has 10+ evidence steps. Would it make sense to make this token-aware, or at least process placeholders longest-first, and add a small regression test for the `#E1` / `#E10` case?

mnajafian-nv (approved):
Conditional approval from my side, subject to addressing the inline comment. The PR looks good overall and the nested placeholder fix is well-scoped; I’d just like to make sure the shared-prefix placeholder case is covered before final approval.

packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/rewoo_agent/agent.py
yczhang-nv: Thanks, the latest commit implemented the token-aware method

PR #2008, opened 2026-06-05
fix(examples): Keep Haystack indexing chunks within NIM embedder token limit

zhongxuanwang-nv:
## Summary

The `haystack_deep_research_agent` example fails to index its bundled documents on startup. The indexing pipeline splits documents by sentence (`split_length=10`), which produces chunks larger than the default NIM embedder's 512-token limit (`nvidia/nv-embedqa-e5-v5`). Embedding then fails with `Input length 576 exceeds maximum allowed token size 512`, aborting the entire indexing run — OpenSearch is left empty and the RAG tool returns nothing (web search still works).

Reported in NAT-309.

## Root cause

`pipelines/indexing.py` built the splitter as:

```python
DocumentSplitter(split_by="sentence", split_length=10, split_overlap=2)
```

Ten sentences of prose/markdown easily exceed 512 tokens. The bundled `data/` produces a 309-word / 576-token chunk that the embedder rejects, and the failure propagates up and skips indexing.

## Fix

- Split by **word** with a bounded length (`split_length=200`, `split_overlap=30`) so every chunk stays well under the 512-token limit. This mirrors Haystack's own canonical `nv-embedqa-e5-v5` indexing example.
- Set `truncate="END"` on `NvidiaDocumentEmbedder` as a safety net: any still-oversized chunk is trimmed by the endpoint instead of failing the run.
- Add a short troubleshooting note to the example README.

The change is intentionally minimal — two lines in `pipelines/indexing.py` plus one README line. No config/API surface changes.

## Validation

Verified against the live NIM endpoint (`nvidia/nv-embedqa-e5-v5`):

| Config | Largest chunk | Result |
| --- | --- | --- |
| before (`sentence` / 10) | 309 words -> 576 tokens | FAIL: `Input length 576 exceeds maximum allowed token size 512` |
| after (`word` / 200 / 30 + `truncate=END`) | 220 words | PASS: all chunks embed |

Running the actual `run_startup_indexing()` end-to-end (in-memory store, no OpenSearch needed) now reports `Indexing complete. Documents written: 10` (was 0). The existing `test_config_yaml_loads_and_has_keys` smoke test still passes.

## By submitting this PR I confirm:

- Existing tests cover the workflow config; the fix was additionally validated against the live NIM embedder.
- Documentation (example README troubleshooting) is updated.

## Summary by CodeRabbit

* **Documentation**
  * Added troubleshooting for "Embedding Token Limit Exceeded" with guidance to reduce indexing chunk size.

* **Bug Fixes**
  * Updated indexing to produce larger, word-based chunks and enable embedding input truncation to avoid exceeding model token limits.

* **Tests**
  * Added a regression test ensuring large documents are split into chunks that stay within embedder token limits.


zhongxuanwang-nv:
/ok to test [c0419d7](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2008/commits/c0419d772e15a79ef560c76d9e9f8efa534bd6af)

docs/source/versions1.json
```diff
@@ -1,7 +1,11 @@
 [
     {
-        "version": "1.8",
+        "version": "1.9",
```
mnajafian-nv, line 3: Should this version-list change live in a separate PR? I am having trouble connecting the `1.8 -> 1.9` docs switch to the Haystack indexing bug fix, and I would prefer to keep this release metadata churn out of a targeted example bugfix.

examples/frameworks/haystack_deep_research_agent/README.md
```diff
@@ -201,6 +201,7 @@ You can customize the workflow by:
 2. **Missing API Keys**: Verify that both NVIDIA_API_KEY and SERPERDEV_API_KEY are set
 3. **No Documents Found**: Check that PDF files are placed in the data directory and the path is correct
 4. **Web Search Fails**: Verify your SerperDev API key is valid and has remaining quota
+5. **Embedding Token Limit Exceeded** (`Input length X exceeds maximum allowed token size 512`): indexing chunks exceed the embedder's 512-token limit. Lower `split_length` in `pipelines/indexing.py`.
```
mnajafian-nv: Could we reword this troubleshooting note so it does not mention `pipelines/indexing.py` as a raw relative path? The current wording looks like it is tripping our path checks in CI.

packages/nvidia_nat_core/src/nat/meta/pypi.md
```diff
@@ -23,19 +23,19 @@ NeMo Agent Toolkit is a flexible library designed to seamlessly integrate your e
 
 ## Key Features
 
-- [**Framework Agnostic:**](https://docs.nvidia.com/nemo/agent-toolkit/1.8/extend/plugins.html) Works with any agentic framework, so you can use your current technology stack without replatforming.
-- [**Reusability:**](https://docs.nvidia.com/nemo/agent-toolkit/1.8/extend/sharing-components.html) Every agent, tool, or workflow can be combined and repurposed, allowing developers to leverage existing work in new scenarios.
-- [**Rapid Development:**](https://docs.nvidia.com/nemo/agent-toolkit/1.8/tutorials/index.html) Start with a pre-built agent, tool, or workflow, and customize it to your needs.
-- [**Profiling:**](https://docs.nvidia.com/nemo/agent-toolkit/1.8/workflows/profiler.html) Profile entire workflows down to the tool and agent level, track input/output tokens and timings, and identify bottlenecks.
+- [**Framework Agnostic:**](https://docs.nvidia.com/nemo/agent-toolkit/1.9/extend/plugins.html) Works with any agentic framework, so you can use your current technology stack without replatforming.
```
mnajafian-nv, line 26: Could we drop the `1.9` doc-link updates from this PR and keep the scope to the indexing fix? These packaging/docs link changes look unrelated to the reported Haystack chunking issue and make the review surface quite a bit larger.

examples/frameworks/haystack_deep_research_agent/src/nat_haystack_deep_research_agent/pipelines/indexing.py
```diff
@@ -36,13 +36,15 @@ def _build_indexing_pipeline(document_store, embedder_model: str) -> Pipeline:
     p = Pipeline()
     p.add_component("joiner", DocumentJoiner())
     p.add_component("cleaner", DocumentCleaner())
+    # Bound chunks under the embedder's 512-token limit; truncate="END" trims any
+    # that still exceed it (sentence splits could overflow).
     p.add_component(
```
mnajafian-nv, line 41: Would it make sense to add a focused regression test alongside this change? The fix looks reasonable, but right now the PR description is carrying the proof and the automated coverage still seems limited to config smoke / broad E2E.

mnajafian-nv (approved):
Conditional approval after the inline comments are resolved and CI is green.

docs/source/versions1.json
zhongxuanwang-nv: Yes... This is a very big issue — thank you so much for catching this!!

zhongxuanwang-nv:
/ok to test [15d6b41](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2008/commits/15d6b41bb26a3859547208fcd469136415d136cd)

PR #2009, opened 2026-06-06
fix(fastapi): fix Generate endpoint for WebSocket non-streaming and HTTP streaming

zhongxuanwang-nv:
## Summary

A ReAct-style workflow whose streaming output type is `ChatResponseChunk` (e.g. `simple_calculator`) broke two of the four Generate transport combos in the chat UI:

- **WebSocket + Generate (non-streaming)** raised a red `ValueError: Cannot convert type ChatResponseChunk to OutputArgsSchema. No match found.` The WebSocket handler always forced `streaming=True`, so a non-streaming request stream-converted chunks to the single-output schema (which has no converter).
- **HTTP + Generate (streaming)** ran the tools but never showed a final answer: the endpoint emitted raw OpenAI `ChatResponseChunk` chunks (plus a `[DONE]` sentinel) instead of the `{"value": ...}` envelope the Generate consumers expect.

Chat Completions and the other two Generate combos already worked and are unchanged.

## Changes

- Derive the WebSocket `streaming` flag from the schema type (`*_stream` streams; `generate`/`chat` aggregate a single result).
- Render the non-streaming Generate answer as plain text over WebSocket (unwrap the auto-derived `OutputArgsSchema` `value`).
- Wrap Generate streaming output in the `{"value": ...}` envelope and skip the OpenAI `[DONE]` sentinel; Chat endpoints keep the OpenAI shape.

## Test plan

- [x] `pytest packages/nvidia_nat_core/tests/nat/server/test_unified_api_server.py --run_integration`
- [x] `front_ends` test suite green
- [ ] Manual UI check of all four Generate settings combos (WebSocket/HTTP × streaming/non-streaming)

## Summary by CodeRabbit

* **New Features**
  * Configurable output wrapping for streaming endpoints; POST streaming now emits wrapped {"value": ...} envelopes while chat streaming preserves OpenAI-style chunks.
  * Streaming behavior now follows the selected workflow schema (only streaming schema types produce streams).

* **Bug Fixes**
  * Message validation now unwraps single-field payloads with a top-level value into plain-text responses.

* **Tests**
  * Added integration and unit tests for streaming envelope formats and WebSocket payload unwrapping.


zhongxuanwang-nv:
/ok to test [e035f3c](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2009/commits/e035f3c67c0ba36275eae10bc0f57f61899d52c5)

zhongxuanwang-nv:
/ok to test [b935de8](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2009/commits/b935de87686c91712ee62b4b19ca133b165f539c)

packages/nvidia_nat_core/tests/nat/server/test_unified_api_server.py
```diff
@@ -886,6 +917,54 @@ class Rich(BaseModel):
     assert decoded == {"value": "21", "tokens": 42}
 
 
+async def test_response_payload_output_websocket_unwraps_value():
+    """Non-streaming generate renders the answer (``42``) over WebSocket, not a ``{"value": "42"}`` blob."""
+
+    class OutputArgsSchema(BaseModel):
+        value: str
+
```
mnajafian-nv, line 926: Should we add a companion WebSocket regression test for the `{value, ...extras}` case here? We already preserve sibling fields for richer `ResponsePayloadOutput` payloads on the SSE side, so I think it would be good to pin whether WebSocket non-streaming generate is supposed to preserve that richer shape or intentionally collapse it to plain text.

packages/nvidia_nat_core/src/nat/front_ends/fastapi/message_validator.py
```diff
@@ -152,11 +152,14 @@ async def convert_data_to_message_content(self, data_model: BaseModel) -> BaseMo
...
             if (isinstance(data_model, ResponsePayloadOutput)):
-                if hasattr(data_model.payload, 'model_dump_json'):
-                    text_content: str = data_model.payload.model_dump_json()
-                else:
-                    text_content: str = str(data_model.payload)
-                return SystemResponseContent(text=text_content)
+                payload = data_model.payload
+                # Unwrap the auto-derived ``OutputArgsSchema`` value so non-streaming generate
+                # responses render the answer (``42``) not a ``{"value": "42"}`` blob.
+                if isinstance(payload, BaseModel) and "value" in type(payload).model_fields:
```
mnajafian-nv: Could we make this unwrap condition more specific than any `BaseModel` with a `value` field? As written, this looks like it would also flatten legitimate structured outputs such as `{"value": "21", "tokens": 42}` down to just `21` over WebSocket, which seems broader than the PR intent of only unwrapping the auto-derived `OutputArgsSchema`.

packages/nvidia_nat_core/src/nat/front_ends/fastapi/response_helpers.py
```diff
@@ -70,7 +72,13 @@ async def generate_streaming_response(payload: typing.Any,
                                       streaming: bool,
                                       step_adaptor: StepAdaptor = StepAdaptor(StepAdaptorConfig()),
                                       result_type: type | None = None,
-                                      output_type: type | None = None) -> AsyncGenerator[ResponseSerializable]:
+                                      output_type: type | None = None,
+                                      wrap_output_in_payload: bool = False) -> AsyncGenerator[ResponseSerializable]:
+
+    # Generate endpoints wrap the workflow output in ResponsePayloadOutput (the {"value": ...} envelope)
```
mnajafian-nv, line 78: Can we clarify in tests whether generate is expected to preserve structured payloads consistently across HTTP streaming and WebSocket non-streaming? I think the new wrapping behavior here and the new unwrapping behavior in `message_validator.py` may diverge for models that expose `value` plus extra fields.

mnajafian-nv (approved):
Main fix looks right and the transport coverage is much better now. I left a few inline comments, but the only approval-blocking one for me is narrowing the WebSocket unwrap logic so we do not accidentally flatten richer payloads that happen to expose a top-level  field. If that gets tightened up and covered by a regression test, I’m good with this.

mnajafian-nv (approved):
Conditional approval, upon addressing inline suggestions and green CI

packages/nvidia_nat_core/tests/nat/server/test_unified_api_server.py
zhongxuanwang-nv: Hey Maryam, thank you for your comments!! I added a test reusing the same Rich(value="21", tokens=42) fixture as the SSE ..._preserves_extras test. The WS non-streaming path intentionally collapses to plain text (text == "21", tokens dropped).

zhongxuanwang-nv:
/ok to review [f52061b](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2009/commits/f52061b1f054389817b66ccb850a69eff412441f)

zhongxuanwang-nv:
/ok to test [eef6c3e](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2009/commits/eef6c3e41fbab788b7e35458f97f813ba050cc56)

PR #2013, opened 2026-06-08
Remove integration package for RagaAI Catalyst

dagardner-nv:
## Description
* This service has been down/unresponsive for a few weeks now.
* Temporarily disable codecov for now to work-around https://github.com/codecov/codecov-action/issues/1956

## Summary by CodeRabbit

* **Removals**
  * RagaAI Catalyst telemetry exporter and related observability support removed.

* **Documentation**
  * Deleted Catalyst integration guides, examples, config snippets, and references across docs and READMEs; updated telemetry listings and install notes.

* **Chores**
  * Removed the optional "ragaai" dependency extra; minor license/redirect formatting and model-config count tweaks; allowlist entry cleaned up.

* **Tests**
  * Replaced Catalyst-specific integration test with a local file-based telemetry lineage test.

* **CI**
  * Coverage upload to Codecov temporarily disabled in CI.


Salonijain27 (approved):

willkill07 (approved):

pyproject.toml
```diff
@@ -75,7 +75,6 @@ phoenix = ["nvidia-nat-phoenix == {version}"]
 profiler = ["nvidia-nat-profiler == {version}"]
 rag = ["nvidia-nat-rag == {version}"]
 ragas = ["nvidia-nat-ragas == {version}"]
-ragaai = ["nvidia-nat-ragaai == {version}"]
```
mnajafian-nv, line 78: Should we also update the example manifests that still depend on `nvidia-nat[...ragaai...]` as part of this removal? With the root extra going away here, I think leaving those references behind would keep a couple of example install paths broken after merge.

packages/nvidia_nat_core/tests/nat/data_models/test_common.py
```diff
@@ -172,22 +172,16 @@ class WeaveExporter(TelemetryExporterBase, name="weave"):
         class PhoenixExporter(BatchConfigMixin, CollectorConfigMixin, TelemetryExporterBase, name="phoenix"):
             pass
 
-        class CatalystExporter(BatchConfigMixin, TelemetryExporterBase, name="catalyst"):
```
mnajafian-nv, line 175: Would it make sense to keep this third mixin-based exporter case and just rename it to something neutral instead of deleting it? This looks like generic `TypedBaseModel` inheritance regression coverage rather than Catalyst-specific behavior, so I would prefer not to shrink that coverage if we do not need to.

docs/source/run-workflows/observe/observe.md
```diff
@@ -76,7 +76,6 @@ The following table lists each exporter with its supported features and configur
 | Provider | Integration Documentation | Supported Features |
 | -------- | ------------------------- | ------------------ |
 | [Arize AX](https://arize.com/docs/ax/) | [Observing with Arize AX](?provider=Arize-AX#provider-integration-guides){.external} | Logging, Tracing |
-| [Catalyst](https://docs.raga.ai/ragaai-catalyst) | [Observing with Catalyst](?provider=Catalyst#provider-integration-guides){.external} | Logging, Tracing |
```
mnajafian-nv, line 79: Should we also remove the remaining Catalyst install commands and provider mentions from this page in the same PR? I want to make sure the public observe docs do not still point users at a deleted integration after this lands.

docs/source/get-started/installation.md
```diff
@@ -51,7 +51,6 @@ To install these first-party plugin libraries, you can use the full distribution
 - `nvidia-nat[openpipe-art]` or `nvidia-nat-openpipe-art` - [Agent Reinforcement Trainer](https://art.openpipe.ai/getting-started/about) Conflicts with `nvidia-nat[adk]` and `nvidia-nat[crewai]`.
 - `nvidia-nat[opentelemetry]` or `nvidia-nat-opentelemetry` - [OpenTelemetry](https://opentelemetry.io/) (includes the `arize_ax` exporter for [Arize AX](https://arize.com/docs/ax/integrations/opentelemetry/opentelemetry-arize-otel))
 - `nvidia-nat[phoenix]` or `nvidia-nat-phoenix` - [Arize Phoenix](https://arize.com/docs/phoenix)
-- `nvidia-nat[ragaai]` or `nvidia-nat-ragaai` - [RagaAI Catalyst](https://raga.ai/) Conflicts with `nvidia-nat[adk]` and `nvidia-nat[strands]`.
```
mnajafian-nv, line 54: Could we also clean up the other `ragaai` mentions on this page while we are here? Right now this diff removes the extra from the supported list, but the surrounding conflict notes still read like `nvidia-nat[ragaai]` is a supported public extra.

docs/source/extend/custom-components/telemetry-exporters.md
```diff
@@ -78,7 +78,6 @@ Examples of existing telemetry exporters include:
 - **OpenTelemetry Collector**: Exports traces to OpenTelemetry-compatible services
 - **Patronus**: Exports traces to Patronus via OTLP
 - **Galileo**: Exports traces to Galileo via OTLP
-- **RagaAI Catalyst**: Exports traces to RagaAI Catalyst
```
mnajafian-nv, line 81: Could we also remove the remaining `RagaAI` mention in the surrounding pre-built integrations text while we are here? I think otherwise this page will still describe it as a supported integration even after the package removal.

pyproject.toml
mnajafian-nv, line 78: Should we also remove the stale `tool.uv.conflicts` entries that still mention `ragaai` in the root `pyproject.toml`? Since this PR removes the extra itself, keeping the resolver conflict metadata around looks like it would leave the packaging contract only partially cleaned up.

mnajafian-nv (approved):
Great work! Conditional approval upon reviewing inline suggestions.

pyproject.toml
dagardner-nv: good catch

PR #2016, opened 2026-06-08
fix(example): Fix `Agno` personal finance workflow for Agno 2.x and switch to NIM

yczhang-nv:
## Description

Updating the Agno personal finance example for Agno 2.x and removing the workflow’s dependency on OpenAI models by switching to NIM for more consistent result.

## Summary by CodeRabbit

* **New Features**
  * Personal Finance workflow now uses an NVIDIA NIM LLM
  * Agent context enhanced: current datetime added and conversation history included (3 runs)

* **Documentation**
  * Setup instructions updated to use NVIDIA API keys
  * Agno framework reference clarified with official project link

* **Tests**
  * Added workflow initialization test with a mock builder; tests updated to use NVIDIA API keys

* **Chores**
  * Minor CI metadata update


examples/frameworks/agno_personal_finance/tests/test_agno_personal_finance_workflow.py
```diff
@@ -18,8 +18,28 @@
...
+
+
+async def test_workflow_initializes_with_agno_v2_constructor_args():
+    from nat.builder.function_info import FunctionInfo
+    from nat_agno_personal_finance.agno_personal_finance_function import AgnoPersonalFinanceFunctionConfig
+    from nat_agno_personal_finance.agno_personal_finance_function import agno_personal_finance_function
+
+    config = AgnoPersonalFinanceFunctionConfig(llm_name="mock_llm", tools=[])
+
+    async with agno_personal_finance_function(config, MockBuilder()) as fn_info:
```
bbednarski9: The new unit test catches the Agno 2.x constructor-arg regression, which is useful. It may be worth having MockBuilder assert that get_llm() and get_tools() are called with LLMFrameworkEnum.AGNO and the expected config names. Right now the test would still pass if this workflow accidentally requested the wrong wrapper type.

bbednarski9 (approved):

examples/frameworks/agno_personal_finance/tests/test_agno_personal_finance_workflow.py
yczhang-nv: Thanks Bryan! Fixed in the latest commit

PR #2021, opened 2026-06-08
fix: handle empty web ingest responses

mnajafian-nv:
## Summary

The CUDA docs ingest path could crash when one scraped page produced empty content. The script cached successful pages, appended a `None` cache path for the empty page, and then passed that `None` path into `BSHTMLLoader`.

This PR makes the web ingest path tolerate partial scrape/cache failures so the valid pages still ingest and the empty result is skipped.

No public GitHub issue.

## Root Cause

`scripts/langchain_web_ingest.py` appended every `cache_html(...)[1]` result into `filenames` without checking whether caching produced a real file path. `scripts/web_utils.py` also logged invalid cache input using the Python `input` builtin, which made the failure message misleading.

The ingest script also scraped the full URL list even after detecting cached files, so cached pages could be fetched again unnecessarily.

## Fix

- Follow redirects during scrape requests and classify empty response bodies as scrape failures.
- Scrape only URLs that are not already present in the local cache.
- Skip cache results that do not produce a file path before loading documents.
- Log invalid cache payloads with the actual payload instead of the Python `input` builtin.
- Add focused regression coverage for invalid cache logging, empty scrape responses, redirect handling, and mixed cached/uncached ingest behavior.

## Validation

- `uv run --extra test pytest -q examples/RAG/simple_rag/tests/test_langchain_web_ingest.py`
- `uv run --extra test ruff check scripts/web_utils.py scripts/langchain_web_ingest.py examples/RAG/simple_rag/tests/test_langchain_web_ingest.py`
- `uv run pre-commit run --all-files`

## Summary by CodeRabbit

* **New Features**
  * Page caching now prevents re-scraping cached URLs, scraping only uncached pages.

* **Bug Fixes**
  * Improved same-origin redirect handling during scraping.
  * Enhanced input validation and logging for caching and scraping.
  * More robust exception handling for disk write failures.

* **Tests**
  * Added comprehensive tests covering scraping, caching, invalid inputs, same‑origin vs cross‑origin redirects, and caching behavior.


examples/RAG/simple_rag/tests/test_langchain_web_ingest.py
```diff
@@ -0,0 +1,227 @@
...
+    failed_url = "https://failed.example.com/reference"
+
+    cached_file, cached_dir = web_utils.get_file_path_from_url(cached_url, str(tmp_path))
+    fresh_file, fresh_dir = web_utils.get_file_path_from_url(uncached_url, str(tmp_path))
+    Path(cached_dir).mkdir(parents=True, exist_ok=True)
+    Path(cached_file).write_text("<html>cached</html>", encoding="utf-8")
+
+    scraped_urls: list[list[str]] = []
+
+    async def fake_scrape(urls: list[str]):
```
bbednarski9, line 249: examples/RAG/simple_rag/tests/test_langchain_web_ingest.py:201: The new test exercises the main regression path well, but fake_scrape() returns the failed page in scraped_pages and also returns a separate failure entry. That is useful for testing defensive skipping, but it does not exactly match the new scrape() contract, where empty-content pages move only to failures. I’d consider adding or adjusting one assertion to cover the real contract through web_utils.scrape() plus main(), or rename the case to make clear it is testing defensive behavior against malformed scrape output.

scripts/web_utils.py
```diff
@@ -51,18 +54,19 @@ async def scrape(urls: list | str, headers: dict = None):
         tasks = [_wrap_request(client.get(
             url,
             headers=headers,
+            follow_redirects=True,
```
bbednarski9: `scrape()` now follows redirects for arbitrary user-provided URLs. That fixes common doc-site redirects, but it also widens SSRF exposure: an external URL can redirect the request to localhost, private network ranges, or metadata IPs. Can we either keep redirects disabled by default, or validate every final redirect target before following it?

bbednarski9 (changes requested):

mnajafian-nv:
Addressed the remaining review feedback in `58012fd`:

- Removed the redundant `exc_info=True` from `_wrap_request()` because `logger.exception()` already includes exception context.
- Fixed the `containnig` typo in the `cache_html()` docstring.
- Updated the scrape tests to match the real success/failure contract.
- Replaced broad redirect following with same-origin-only redirect handling and added regression coverage for blocked cross-origin redirects.

Validation rerun:
- `uv run --extra test pytest -q examples/RAG/simple_rag/tests/test_langchain_web_ingest.py`
- `uv run --extra test ruff check scripts/web_utils.py scripts/langchain_web_ingest.py examples/RAG/simple_rag/tests/test_langchain_web_ingest.py`
- `uv run pre-commit run --all-files`

scripts/web_utils.py
mnajafian-nv: Thanks for flagging this. I replaced `follow_redirects=True` with explicit redirect handling that keeps redirects disabled at the httpx layer and only follows redirects when the target remains same-origin with the original URL. Cross-origin redirects, including redirects to localhost or metadata IPs, are returned as scrape failures instead of being fetched.

I added coverage for both same-origin redirects and blocked cross-origin redirects. Validation: focused pytest passes, ruff passes, and `uv run pre-commit run --all-files` passes.

bbednarski9 (approved):

mnajafian-nv:
Addressed the valid CodeRabbit feedback in 18e4e556: removed `exc_info=True` before re-raise and moved `DummyResponse` fields into `__init__`. Kept `@pytest.mark.asyncio` intentionally because root pytest-asyncio runs in strict mode and NAT tests consistently use explicit async markers. Validated with focused pytest, ruff, diff check, and pre-commit.

PR #2022, opened 2026-06-08
chore: Update dependencies

dagardner-nv:
## Description
* Set minimum of:
  - lanchain: v1.3.3
  - langsmith: v0.8
  - lxml-html-clean: v0.4.4
  - pywjt: v2.13

* Set minimum version of banks to 2.4.2
* Set minimum v


Salonijain27 (approved):

packages/nvidia_nat_llama_index/pyproject.toml
```diff
@@ -68,6 +68,9 @@ dependencies = [
   "llama-index-readers-file>=0.5.6,<1.0.0",
   "llama-index>=0.14.12,<1.0.0",
   "nltk>=3.9.4,<4.0.0",
+  # Transitive dep
+  # Remove once llama-index-core updates banks
+  "banks>=2.4.2"
```
mnajafian-nv: Could we add an upper bound here, for example `banks>=2.4.2,<3.0.0`, to match the surrounding dependency style and avoid this transitive workaround admitting a future breaking `banks` major release?

packages/nvidia_nat_langchain/pyproject.toml
```diff
@@ -72,6 +72,7 @@ dependencies = [
   "langchain-openai>=1.1.6,<2.0.0",
   "langchain-tavily>=0.2.18,<1.0.0",
   "langgraph>=1.0.5,<2.0.0",
+  "langsmith~=0.8",
```
mnajafian-nv, line 74: Could we clarify whether the LangSmith minimum is intended to apply only to the LangChain package, or repo-wide? Some package locks that still expose LangSmith-related test helpers remain on `langsmith==0.7.5`, so I want to make sure the release dependency boundary matches the PR intent.

uv.lock
```diff
@@ -4745,14 +4749,14 @@ html-clean = [
 
 [[package]]
 name = "lxml-html-clean"
-version = "0.4.3"
+version = "0.4.4"
```
mnajafian-nv, line 4737: Could we confirm whether this is meant to enforce a minimum `lxml-html-clean` version or only refresh the root lock? I do not see a matching `pyproject.toml` constraint, so non-root resolution could still choose an older compatible version.

mnajafian-nv (approved):
LGTM, Conditional approval upon Green CI and addressing inline suggestions.

packages/nvidia_nat_langchain/pyproject.toml
dagardner-nv: That's weird, the only example that called out langsmith as a dependency is `examples/observability/simple_calculator_observability/pyproject.toml`. and I updated that one.

dagardner-nv: I'll need to dig into this more.

PR #2043, opened 2026-06-09
chore: Remove DBNL appears to have gone away

dagardner-nv:
## Description
* This service appears to have gone away. 
* http://docs.dbnl.com/ is down
* https://www.distributional.com/ only has a vague "Something new is coming
Check back soon for updates." message

## Summary by CodeRabbit

* **Documentation**
  * Removed DBNL telemetry exporter integration guides, provider docs, example README sections, and example observability walkthroughs referencing DBNL.
  * Cleansed docs references and linkcheck entries related to DBNL.

* **Chores**
  * Discontinued support for the DBNL telemetry exporter and removed its example configuration.
  * Adjusted an internal CI model configuration count.


packages/nvidia_nat_opentelemetry/src/nat/plugins/opentelemetry/register.py
```diff
@@ -312,45 +312,3 @@ async def arize_ax_telemetry_exporter(config: ArizeAxTelemetryExporter, builder:
         drop_on_overflow=config.drop_on_overflow,
         shutdown_timeout=config.shutdown_timeout,
     )
```
mnajafian-nv, line 313: Could we also remove the now-unused `OptionalSecretStr` import near the top of this file? With `DBNLTelemetryExporter` deleted, Ruff flags that import as `F401`, so this should keep the removal PR lint-clean.

mnajafian-nv (approved):
LGTM, thanks, just a small inline note.

PR #2060, opened 2026-06-11
feat(middleware): add HITLMiddleware for human-in-the-loop function interception

ericevans-nv:
## Summary

- Introduces `HITLMiddleware`, an abstract `DynamicFunctionMiddleware` subclass that presents configurable `HumanPrompt`s before and after any intercepted function call
- Adds `InvocationAction.SKIP` to `InvocationContext` so pre-invoke can signal the framework to bypass the function call; checked in both `function_middleware_invoke` and `function_middleware_stream`
- `HITLMiddlewareConfig` provides optional `pre_invoke_prompt` / `post_invoke_prompt` fields and inherits all function-targeting fields from `DynamicMiddlewareConfig`

## Summary by CodeRabbit

* **New Features**
  * Human-in-the-Loop (HITL) middleware that can prompt users before and/or after function calls.
  * Middleware can explicitly skip execution and control streaming behavior; post-invoke decisions are applied per streamed chunk.

* **Documentation**
  * Added a comprehensive guide for implementing, configuring, and registering HITL middleware (including YAML usage and examples).

* **Tests**
  * Added tests covering prompt presentation, skip/proceed behavior, and streaming/post-invoke semantics.


rpelevin:
This is a useful direction. The part I would make explicit is that HITL middleware is not just prompt plumbing; it becomes an invocation-scoped execution boundary.

For pre-invoke SKIP, I would keep the contract visible to the caller:

- the wrapped function is not called in either invoke or stream mode;
- a skipped invocation is distinguishable from a function that legitimately returns None;
- the terminal record carries function name, invocation id if available, prompt id or prompt version, decision outcome, responder source, decided_at, and skip reason;
- modified args are bound to the same decision that allowed the invocation.

For post-invoke prompts, I would keep a separate record: the function ran, produced an output, and that output was then accepted, replaced, or cleared. That avoids mixing "the callable did not run" with "the callable ran and its result was rejected or transformed."

The regression tests I would add are:

1. pre-invoke skip never calls the function and has the same terminal semantics in single and streaming paths.
2. a real None return is distinguishable from a skipped invocation.
3. modified args cannot be applied under a stale prompt response.
4. post-invoke output replacement preserves original versus effective output state without reclassifying execution as skipped.

That keeps the middleware small while making the control boundary clear: which exact invocation was proposed, who decided, what changed, and whether the callable actually ran.

ericevans-nv:
> This is a useful direction. The part I would make explicit is that HITL middleware is not just prompt plumbing; it becomes an invocation-scoped execution boundary.
> 
> For pre-invoke SKIP, I would keep the contract visible to the caller:
> 
> * the wrapped function is not called in either invoke or stream mode;
> * a skipped invocation is distinguishable from a function that legitimately returns None;
> * the terminal record carries function name, invocation id if available, prompt id or prompt version, decision outcome, responder source, decided_at, and skip reason;
> * modified args are bound to the same decision that allowed the invocation.
> 
> For post-invoke prompts, I would keep a separate record: the function ran, produced an output, and that output was then accepted, replaced, or cleared. That avoids mixing "the callable did not run" with "the callable ran and its result was rejected or transformed."
> 
> The regression tests I would add are:
> 
> 1. pre-invoke skip never calls the function and has the same terminal semantics in single and streaming paths.
> 2. a real None return is distinguishable from a skipped invocation.
> 3. modified args cannot be applied under a stale prompt response.
> 4. post-invoke output replacement preserves original versus effective output state without reclassifying execution as skipped.
> 
> That keeps the middleware small while making the control boundary clear: which exact invocation was proposed, who decided, what changed, and whether the callable actually ran.



> This is a useful direction. The part I would make explicit is that HITL middleware is not just prompt plumbing; it becomes an invocation-scoped execution boundary.
> 
> For pre-invoke SKIP, I would keep the contract visible to the caller:
> 
> * the wrapped function is not called in either invoke or stream mode;
> * a skipped invocation is distinguishable from a function that legitimately returns None;
> * the terminal record carries function name, invocation id if available, prompt id or prompt version, decision outcome, responder source, decided_at, and skip reason;
> * modified args are bound to the same decision that allowed the invocation.
> 
> For post-invoke prompts, I would keep a separate record: the function ran, produced an output, and that output was then accepted, replaced, or cleared. That avoids mixing "the callable did not run" with "the callable ran and its result was rejected or transformed."
> 
> The regression tests I would add are:
> 
> 1. pre-invoke skip never calls the function and has the same terminal semantics in single and streaming paths.
> 2. a real None return is distinguishable from a skipped invocation.
> 3. modified args cannot be applied under a stale prompt response.
> 4. post-invoke output replacement preserves original versus effective output state without reclassifying execution as skipped.
> 
> That keeps the middleware small while making the control boundary clear: which exact invocation was proposed, who decided, what changed, and whether the callable actually ran.

The middleware contract is transparent interception, the caller gets back the function's effective result and should never need to know a middleware ran. Returning SKIPPED leaks the middleware's existence into the caller's API contract. It also changes the effective return type of the intercepted function. Any value other than the function's declared return type breaks the contract the caller already depends on.

HITLMiddleware is intentionally minimal: it defines the interception boundary and the decision contract, nothing more. Observability, audit logging, or any behavior around the decision belongs in the concrete subclass, where implementers can customize it to their requirements without affecting callers or the abstract interface.

rpelevin:
I agree with that caller-contract boundary. The function caller should not have to handle a new sentinel value just because middleware intercepted the invocation, and the declared return type should stay stable.

The distinction I would preserve is control-plane visibility, not necessarily caller-visible return shape.

For pre-invoke skip, the function should not run, but the skip can be represented as middleware state or an invocation record rather than as the function's return value. That gives three separate facts:

- caller-facing result follows the framework's chosen middleware contract;
- the control/audit path can distinguish skipped invocation from a real None return;
- the decision record captures proposed function, proposed args, prompt or decision id, outcome, responder source, decided_at, and skip reason.

That also keeps post-invoke prompts separate. If the function ran and the output was replaced or cleared, the record should say function_ran=true and preserve original versus effective output. If pre-invoke denied or skipped, function_ran=false.

The tests I would add:

1. pre-invoke skip does not call the function in invoke or stream mode, while the caller-facing contract remains stable.
2. a real None return and a skipped invocation are distinguishable in the middleware/control record.
3. modified args are bound to the exact prompt response; a stale response cannot apply to a later invocation.
4. post-invoke output replacement records that the function ran and preserves original versus effective output.
5. a concrete subclass can emit audit/logging/export records without changing the abstract interface or caller return contract.

That keeps HITLMiddleware minimal while still making the execution boundary inspectable: what invocation was proposed, who decided, whether args changed, whether the callable ran, and what effective result continued.

ericevans-nv:
> I agree with that caller-contract boundary. The function caller should not have to handle a new sentinel value just because middleware intercepted the invocation, and the declared return type should stay stable.
> 
> The distinction I would preserve is control-plane visibility, not necessarily caller-visible return shape.
> 
> For pre-invoke skip, the function should not run, but the skip can be represented as middleware state or an invocation record rather than as the function's return value. That gives three separate facts:
> 
> * caller-facing result follows the framework's chosen middleware contract;
> * the control/audit path can distinguish skipped invocation from a real None return;
> * the decision record captures proposed function, proposed args, prompt or decision id, outcome, responder source, decided_at, and skip reason.
> 
> That also keeps post-invoke prompts separate. If the function ran and the output was replaced or cleared, the record should say function_ran=true and preserve original versus effective output. If pre-invoke denied or skipped, function_ran=false.
> 
> The tests I would add:
> 
> 1. pre-invoke skip does not call the function in invoke or stream mode, while the caller-facing contract remains stable.
> 2. a real None return and a skipped invocation are distinguishable in the middleware/control record.
> 3. modified args are bound to the exact prompt response; a stale response cannot apply to a later invocation.
> 4. post-invoke output replacement records that the function ran and preserves original versus effective output.
> 5. a concrete subclass can emit audit/logging/export records without changing the abstract interface or caller return contract.
> 
> That keeps HITLMiddleware minimal while still making the execution boundary inspectable: what invocation was proposed, who decided, whether args changed, whether the callable ran, and what effective result continued.

Any additional tracking or observability is a concrete subclass concern — the abstract interface stays clean. A concrete implementation can override the pre/post hooks to record decisions, track state, or enforce any binding logic it needs.

rpelevin:
Agree that the abstract interface should stay clean. If observability is owned by concrete subclasses, the key contract is not a new return type or audit API; it is stable hook context.

The subclass needs enough context in the pre and post hooks to bind its own record without changing what the caller sees. For pre-invoke, that means it can consistently see:

- invocation id or equivalent correlation key;
- function name and proposed args, or a digest of them;
- prompt or decision id;
- decision outcome and responder source;
- decided_at;
- whether args were modified;
- whether the callable ran;
- skip or denial reason when it did not run.

For post-invoke, it should be able to record that the callable did run, plus original versus effective output when the output is replaced or cleared.

That keeps the split clean:

- the abstract middleware defines interception and hook timing;
- the concrete subclass owns records, logging, export, and binding rules;
- the caller receives the normal effective result, not audit metadata.

The tests I would add are:

1. a concrete subclass sees the same invocation correlation key through pre-invoke, invoke or stream, and post-invoke.
2. pre-invoke skip lets the subclass record function_ran=false without returning a caller-visible sentinel.
3. modified args are visible to the subclass as proposed versus effective args, or equivalent digests.
4. post-invoke replacement records function_ran=true and preserves original versus effective output.
5. disabling the concrete subclass leaves no audit side effects and does not change caller return semantics.

That gives implementers room to build audit or enforcement behavior in subclasses while keeping HITLMiddleware minimal and transparent to function callers.

Yash-xoxo:
Keeping the caller contract clean is non-negotiable, and audit/binding logic lives in concrete subclasses.

To close the loop on what the abstract interface actually needs to expose, here's a concrete proposal for the hook context objects:

For `_on_pre_invoke_response`:
- `function_name: str`
- `proposed_args: dict` — snapshot of args at decision time
- `prompt_id: str` — correlates the response back to the prompt version that produced it
- `decided_at: datetime`
- `responder_source: str` — human, system, timeout, etc.
- `outcome: Literal["proceed", "skip"]`
- `modified_args: dict | None` — set only when args were changed; None means unmodified, not skipped
- `skip_reason: str | None`

For `_on_post_invoke_response`:
- same correlation fields (`function_name`, `prompt_id`, `decided_at`, `responder_source`)
- `function_ran: bool` — always True here, but explicit for subclass parity
- `original_output: Any`
- `effective_output: Any` — same object as `original_output` if unchanged
- `outcome: Literal["accepted", "replaced", "cleared"]`

The abstract class never writes a record itself — it just guarantees the subclass sees these fields. Concrete classes can ignore what they don't need.

This also covers the stale-args concern: since `modified_args` is bound to a specific `prompt_id`, a subclass can enforce that it's only applied when the prompt response and the live invocation share the same correlation key.

rpelevin:
This field shape is close. I would add one constraint to make the stale-args boundary enforceable: prompt_id should identify the prompt or decision, but it should not be the whole replay boundary.

For pre-invoke, the hook context should carry both:

- invocation_id or equivalent per-call correlation key;
- prompt_id or decision_id for the human or system response;
- proposed_args or proposed_args_digest at prompt time;
- effective_args or effective_args_digest after modification;
- function_name;
- outcome, responder_source, decided_at, and skip_reason.

Then a concrete subclass can enforce that modified_args only applies when the live invocation correlation key and prompt or decision id both match. If either side changes, it should fail closed or require a new prompt cycle. That avoids the subtle case where a prompt response is valid in isolation but gets replayed onto a later call with the same prompt shape.

For post-invoke, I like keeping function_ran explicit. I would also attach original_output and effective_output to the same invocation correlation key so replacement or clearing cannot be confused with a different call's result.

The regression tests I would add:

1. modified_args from one invocation cannot be applied to a later invocation with a different correlation key.
2. same prompt_id with changed proposed_args is rejected unless a new prompt or decision is issued.
3. stream and non-stream paths expose the same correlation key to the subclass.
4. pre-invoke skip records function_ran=false without changing caller return semantics.
5. post-invoke replacement records function_ran=true and preserves original versus effective output under the same invocation key.

That keeps the abstract class clean, while giving concrete subclasses enough structure to make stale decisions non-replayable.

docs/source/build-workflows/advanced/middleware.md
```diff
@@ -404,6 +404,96 @@ Timeout middleware extends `DynamicFunctionMiddleware`, enabling interception of
...
+#### Abstract interface
+
+Both abstract methods receive the `InteractionResponse` from the user and the current `InvocationContext`. They return `InvocationContext | None`:
+
+- **`_on_pre_invoke_response(response, context)`**: Called after the pre-invoke prompt is answered. Return `None` to proceed with the call unchanged. To skip the call entirely, set `context.action = InvocationAction.SKIP` and return `context`. To modify arguments, update `context.modified_args` and return `context`.
+- **`_on_post_invoke_response(response, context)`**: Called after the post-invoke prompt is answered. Return `None` to accept the output as-is. To replace or clear the output, update `context.output` and return `context`.
+
+#### Configuration
+
+`HITLMiddlewareConfig` is the base configuration class. It already declares `name="hitl"`, which maps to `_type: hitl` in YAML. Subclass it with a different `name` only if you need a distinct `_type` value. It provides two prompt fields and inherits all function-targeting fields from `DynamicMiddlewareConfig`.
```
mnajafian-nv: Could we tighten this contract before merging? Since `HITLMiddleware` is abstract and there is no built-in `hitl` middleware registered in `nat.middleware.register`, documenting `HITLMiddlewareConfig` as mapping to `_type: hitl` makes that YAML type look runnable when it is not. I think the cleaner contract is either to remove `name="hitl"` from the abstract base and require concrete subclasses to declare their own `name`, or to add a concrete registered middleware for `_type: hitl` if that type is intended to work out of the box.

docs/source/build-workflows/advanced/middleware.md
```diff
@@ -404,6 +404,96 @@ Timeout middleware extends `DynamicFunctionMiddleware`, enabling interception of
...
+```
+
+#### Implementing a custom HITL middleware
+
+Subclass `HITLMiddleware` to implement the decision logic. Subclass `HITLMiddlewareConfig` with a new `name` only when you need a distinct `_type`:
+
+```python
+from nat.data_models.interactive import InteractionResponse
+from nat.middleware.hitl import HITLMiddleware, HITLMiddlewareConfig
+from nat.middleware.middleware import InvocationAction, InvocationContext
```
mnajafian-nv: Should we route these new middleware-authoring symbols through `nat.plugin_api`, or explicitly mark this as a specialized/provisional implementation API? The public plugin API docs tell external plugin authors to import authoring contracts from `nat.plugin_api`, but this example introduces `HITLMiddleware`, `HITLMiddlewareConfig`, and `InvocationAction` from implementation modules. If this is intended as the supported extension path, I would expect the new symbols to be exported, pinned in the plugin API tests, and added to the public API docs.

packages/nvidia_nat_core/tests/nat/middleware/test_hitl_middleware.py
```diff
@@ -0,0 +1,287 @@
...
+        yield "x"
+        yield "y"
+
+    with _patch_user_prompt(_response("no")):
+        chunks: list[str] = [
+            chunk async for chunk in middleware.function_middleware_stream("input", call_next=call_next,
+                                                                           context=function_context)
+        ]
+
+    assert chunks == [None, None]
```
mnajafian-nv: Could we make this streaming contract explicit and align the test name/docstring with the assertion? Right now the test says clearing `context.output` yields no chunks, but the asserted behavior is `[None, None]` because `function_middleware_stream` always yields `ctx.output` after `post_invoke`. If `None` is meant to suppress a chunk, the implementation should skip yielding it; if `None` is meant to be a valid replacement chunk, the test should say that clearly so stream consumers do not get surprised by `None` values.

mnajafian-nv (approved):
LGTM, with conditional approval pending resolution of the inline suggestion, Thanks!

rpelevin:
I agree with making these merge blockers explicit. I would treat the three comments as one contract boundary:

1. Abstract base versus runnable middleware. If HITLMiddleware is only an authoring base, the base config should not make _type: hitl look runnable. Either remove the concrete name from the abstract base and require concrete subclasses to declare their own _type, or add a real default registered middleware and test that _type: hitl loads end to end. The docs should not imply a YAML type exists unless the registry can resolve it.

2. Public extension API. If external authors are expected to subclass this, the stable imports should come through nat.plugin_api and be pinned by public API tests: HITLMiddleware, HITLMiddlewareConfig, InvocationAction, and the context/response types needed to implement the hooks. If the surface is provisional, say that directly and keep implementation-module imports out of the public authoring example.

3. Stream post-invoke semantics. I would avoid bare None doing double duty as both "replacement value" and "suppress output". The test and implementation should choose one explicit contract:
- post hook returns a suppress/clear action and no chunk is yielded; or
- None is a valid replacement chunk and the docs/test name say consumers may receive it.

For HITL, I would prefer the first contract. A reviewer or subclass author should be able to tell whether the human decision suppressed output, replaced output, or passed it through without inferring that from an ambiguous None in the stream.

That keeps the merge boundary tight: runnable YAML types match registry behavior, the extension surface is either public or explicitly provisional, and stream consumers do not have to reverse-engineer human decision semantics from emitted chunks.

rpelevin:
Nice merge. The final shape lands the important contract split cleanly: the abstract HITL middleware is an authoring surface, runnable YAML types stay tied to concrete registered middleware, supported extension symbols are promoted through the plugin API, and stream suppression is explicit instead of leaking ambiguous None chunks.

The next boundary I would preserve as concrete HITL subclasses appear is decision-record binding.

For pre-invoke decisions, the subclass should be able to record:

- invocation id or per-call correlation key;
- function name;
- proposed args digest at prompt time;
- effective args digest after modification;
- decision id or prompt id;
- outcome and responder source;
- decided_at;
- whether the callable ran;
- skip or denial reason.

For post-invoke decisions, the record should say that the callable did run, then bind original output, effective output, and whether the human/system decision accepted, replaced, cleared, or suppressed it.

The regression tests I would keep close to this surface:

1. abstract base cannot be loaded as a runnable YAML middleware unless a concrete type is registered.
2. public plugin API exports are pinned for external middleware authors.
3. stream clear/suppress emits no chunk, while replacement output remains explicit.
4. modified args cannot be replayed onto a different invocation id.
5. skipped invocation remains distinguishable in control/audit state from a real None return.

That keeps the merge durable: caller contracts stay clean, extension authors get a stable surface, and the decision stays bound to the exact invocation rather than becoming hidden prompt state.

PR #2064, opened 2026-06-12
fix(agent): flatten list-style message content for Bedrock/Anthropic

### LINKED ISSUE #2063: Tool Calling and ReAct agents drop Anthropic/Bedrock list-style message content (empty streamed reply, stringified list on /chat)
### Version

1.7.0

PyPi

### Describe the bug.

The LangChain `tool_calling_agent` and `react_agent` assume `message.content` is a string. Anthropic models - including via AWS Bedrock (`langchain_aws.ChatBedrockConverse`) return `content` as a list of typed blocks, e.g. `[{"type": "text", "text": "..."}]`.
 
 Because of this mismatch:
 - Streaming (`_stream_fn`, used by `/chat/stream` and the websocket UI) gates on `if isinstance(msg.content, str)`. With list content this is always False, so no tokens are emitted and the client receives an EMPTY reply even though the agent ran its tools and produced an answer.
 - Single-shot (`_response_fn`, used by `/chat` and `nat run`) returns `str(output_message.content)`, so the client receives a stringified Python list like `"[{'type': 'text', 'text': 'Hello'}]"` instead of `"Hello"`.
 
Expected: the agent returns the model's text answer on both the streaming and single-shot paths, regardless of whether the provider returns string or list-style content.
 
 Affected (same `str`-only assumption):
 - packages/nvidia_nat_langchain/.../agent/tool_calling_agent/register.py (_stream_fn, _response_fn)
 - packages/nvidia_nat_langchain/.../agent/react_agent/register.py (_stream_fn, _response_fn)
 (the pattern also appears in router_agent, auto_memory_wrapper, and rewoo_agent.)
 
 OpenAI/NIM/Gemini are unaffected because they return string content.

```shell
# config.yml — any Anthropic model on Bedrock reproduces it
 cat > config.yml <<'YAML'
 llms:
   bedrock_llm:
     _type: aws_bedrock
     model_name: "us.anthropic.claude-3-5-sonnet-20241022-v2:0"
     region_name: "us-east-1"
     max_tokens: 1024
 
 functions:
   current_datetime:
     _type: current_datetime
 
 workflow:
   _type: tool_calling_agent
   llm_name: bedrock_llm
   tool_names:
     - current_datetime
 YAML
 
 # 1) Single-shot: returns a stringified list instead of the answer
 nat run --config_file config.yml --input "What is the current time? Use the tool."
 
 # 2) Streaming (what the UI uses): returns an EMPTY reply
 nat serve --config_file config.yml --host 0.0.0.0 --port 8000 &
 curl -sN -X POST http://localhost:8000/chat/stream \
   -H "Content-Type: application/json" \
   -d '{"messages":[{"role":"user","content":"What is the current time? Use the tool."}]}'
```

### Relevant log output

 # /chat (single-shot) — content is a stringified list, not the answer:
 {"choices":[{"message":{"role":"assistant", "content":"[{'type': 'text', 'text': 'The current time is ...', 'index': 0}]"}}]}
 
 # /chat/stream — only empty deltas are emitted, then [DONE]:
 data: {"choices":[{"index":0,"delta":{"content":"","role":null,"tool_calls":null}, "finish_reason":"stop"}], ... "object":"chat.completion.chunk"}
 data: [DONE]
 
 </pre></details>

### Other/Misc.

Root cause: agents consume message content as `str`; Anthropic/Bedrock return a list of content blocks. Verified against `au.anthropic.claude-opus-4-8` on Bedrock and confirmed the same code on `develop` HEAD (so a version bump does not fix it).
 
I have a fix prepared (shared `_extract_message_text` helper in the agent base module, applied to the tool_calling and ReAct agents' streaming + single-shot paths, with unit tests) and will open a PR referencing this issue.

### Code of Conduct

ctrlsam:
## Description
 Closes #2063
 
 The tool calling and ReAct agents consumed message content as a string: the streaming paths gated on `isinstance(msg.content, str)` and the single-shot paths called `str(output_message.content)`. Anthropic models, including via AWS Bedrock (`ChatBedrockConverse`), return content as a list of typed blocks (for example `[{"type": "text", "text": "..."}]`). As a result the agents streamed an empty reply over `/chat/stream` and the websocket, and returned a stringified list over `/chat`.
 
Adds a shared `_extract_message_text` helper in the agent base module that flattens string or list-of-blocks content to plain text, and uses it in the tool calling and ReAct agents' streaming and single-shot paths. Non-text blocks (such as tool use and reasoning) are ignored.
 
 ### Testing
 - New unit tests cover string, list, mixed, empty, and non-text content shapes.
 - Existing `test_base`, `test_tool_calling`, and `test_react` suites pass.
 - Manually verified against `anthropic.claude` models on AWS Bedrock: `/chat`, `/chat/stream`, and the websocket all return text (previously empty/garbled).
 
## Release Notes

* **Refactor**
  * Improved normalization of agent message content so plain-text output is produced consistently across both standard and streaming workflows, including provider-specific list-style formats.

* **Tests**
  * Added unit and regression tests to verify message content extraction, correct streaming chunk reconstruction, and proper parsing of ReAct tool-call steps with list-style content.

### DIFF
--- a/packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/base.py
+++ b/packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/base.py
@@ -45,6 +45,31 @@ def _extract_reasoning_content(message: BaseMessage) -> str:
     return reasoning if isinstance(reasoning, str) else str(reasoning)
 
 
+def _extract_message_text(content: Any) -> str:
+    """Flatten LangChain message content to plain text.
+
+    Chat models return ``content`` either as a string or, for providers such as
+    Anthropic (including via AWS Bedrock ``ChatBedrockConverse``), as a list of
+    typed content blocks (for example ``{"type": "text", "text": "..."}``). The
+    agents consume message content as text, so this concatenates the text of any
+    text blocks and ignores non-text blocks such as tool-use or reasoning. A
+    plain string is returned unchanged and any other type yields an empty string.
+    """
+    if isinstance(content, str):
+        return content
+    if isinstance(content, list):
+        parts: list[str] = []
+        for block in content:
+            if isinstance(block, str):
+                parts.append(block)
+            elif isinstance(block, dict) and block.get("type", "text") == "text":
+                text = block.get("text")
+                if isinstance(text, str):
+                    parts.append(text)
+        return "".join(parts)
+    return ""
+
+
 def _format_agent_thoughts_for_log(message: BaseMessage) -> str:
     """Return concise text for detailed agent thought logs."""
     content = message.content
--- a/packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/react_agent/agent.py
+++ b/packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/react_agent/agent.py
@@ -43,6 +43,7 @@ from nat.plugins.langchain.agent.base import INPUT_SCHEMA_MESSAGE
 from nat.plugins.langchain.agent.base import NO_INPUT_ERROR_MESSAGE
 from nat.plugins.langchain.agent.base import TOOL_NOT_FOUND_ERROR_MESSAGE
 from nat.plugins.langchain.agent.base import AgentDecision
+from nat.plugins.langchain.agent.base import _extract_message_text
 from nat.plugins.langchain.agent.base import _format_agent_thoughts_for_log
 from nat.plugins.langchain.agent.dual_node import DualNodeAgent
 from nat.plugins.langchain.agent.react_agent.output_parser import ReActAgentParsingFailedError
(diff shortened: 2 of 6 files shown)


yczhang-nv:
@ctrlsam Thanks for jumping on this! I think this fixes the tool-calling agent paths, but `ReAct` still needs one more iteration.

`react_agent/register.py` now extracts text in the outer response/stream handlers, but `ReActAgentGraph.agent_node` still passes `output_message.content` directly into `ReActOutputParser().aparse(...)` before the register-layer `_response_fn` sees it.

On PR head `d075103ca`, list-style content still fails there:

```python
await ReActOutputParser().aparse([
    {"type": "text", "text": "Thought: done\nFinal Answer: block answer"}
])
```

This raises:
`TypeError: expected string or bytes-like object, got 'list'`

`_chunk_to_message(AIMessageChunk(content=[...]))` also preserves `.content` as a list, so this is reachable for Bedrock/Anthropic streamed chunks.

Could you please normalize ReAct `output_message.content` inside `ReActAgentGraph.agent_node` before `remove_r1_think_tags`, logging, parser invocation, retry handling, and final-answer assignment?

It would also be good to add a regression test at the ReAct graph or workflow level with list-style message content, not only base helper coverage.

yczhang-nv:
@ctrlsam Please feel free to let me know when you finished iterating on this. I'll kick-off CI for this PR then

yczhang-nv (changes requested):
Please resolve the comments above

ctrlsam:
Good catch @yczhang-nv, thanks. Pushed 65cbac24: agent_node now flattens `output_message.content `with `_extract_message_text` right after the LLM call (both branches), so `remove_r1_think_tags`, the output parser, final-answer assignment, and retry handling all see a string.

Added ReAct graph-level regression tests for list content on the final-answer and tool-call paths. Beyond unit coverage (full nvidia_nat_langchain suite: 680 passed), I ran `nat serve` from this branch against Claude Opus 4.8 on AWS Bedrock - `react_agent` and `tool_calling_agent` both return correct text over `/chat`, `/chat/stream`, and the websocket including a real tool-calling cycle, with clean logs and no TypeError.

yczhang-nv:
Thanks @ctrlsam for the quick update! PR merged

yczhang-nv (approved):

PR #2067, opened 2026-06-12
fix: harden Dynamo latency demo against SGlang version drift and degenerate trie predictions

bbednarski9:
## Description

Fixes two failures in `examples/dynamo_integration/latency_sensitivity_demo` (reported against NAT v1.8.0-rc5) when run on a source-built Dynamo ≥1.1.0 (observed on ai-dynamo 1.3.x). The SGLang worker exited before serving in Step 3, and Step 4 failed requests on certain prediction-trie values. Jira: NAT-391.

## Problem

**Step 3 — worker exits during argument parsing.**`dynamo_stack_sensitivity.sh` hardcoded `--schedule-low-priority-values-first`. `dynamo.sglang` forwards this straight to SGLang's argparse via `ServerArgs.add_cli_args`, so any flag the bundled SGLang doesn't define is rejected as an unrecognized argument and the worker dies
before registering on `/v1/models`. The flag's availability is SGLang-version dependent — confirmed against real installs: 0.4.9.post6 has neither priority flag, 0.5.4.post3 has both, and the source-built Dynamo in the report bundles an intermediate SGLang that has `--enable-priority-scheduling` but not  `--schedule-low-priority-values-first`.

**Step 4 — request fails on degenerate predictions.** Prediction-trie overrides in `dynamo_llm.py` truncate sub-1 means to 0 via `int()` (e.g. `remaining_calls.mean < 1`on a conversation's final call), which tripped the `>= 1` nvalidation guard and failed the request.

## Changes

- **`dynamo_stack_sensitivity.sh`**: probe the installed SGLang `ServerArgs` and pass only the priority flags it actually supports, instead of hardcoding them. Warn clearly when `--schedule-low-priority-values-first` (lower-value-first ordering, which the demo's priority semantics depend on) is unavailable rather than crashing.
- **`dynamo_llm.py`**: when a prediction override yields a value `< 1`, fall back to the Pydantic-validated static config value (`total_requests` / `osl` / `iat`) so a degenerate prediction no longer fails the request. The downstream validation guard is retained for direct/`latency_sensitivity` cases. - **`test_dynamo_llm.py`**: the three `test_transport_raises_when_*_zero` tests encoded the old raise-on-zero behavior; rewritten a ,`test_transport_falls_back_when_prediction_*_zero` to assert the fallback and that other prediction-derived fields pass through.

This supersedes the reported manual workaround (deleting the flag and patching the validation site): deleting the flag silently inverts priority scheduling on SGLang versions that *do* support it, whereas the probe preserves intended behavior where available and degrades gracefully where not.

## Testing

- `pytest packages/nvidia_nat_core/tests/nat/llm/` → 120 passed.
- `bash -n dynamo_stack_sensitivity.sh` passes; priority-flag selection verified against
  all four version cases (both flags / enable-only / neither / probe-failure).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved prediction handling by validating derived metrics and automatically falling back to static configuration values when predictions produce invalid or below-threshold results.
  * Enhanced worker initialization with improved runtime detection of priority-scheduling capabilities, replacing hardcoded assumptions and ensuring compatibility across different deployment configurations.

### DIFF
--- a/examples/dynamo_integration/latency_sensitivity_demo/src/latency_sensitivity_demo/scripts/dynamo_stack_sensitivity.sh
+++ b/examples/dynamo_integration/latency_sensitivity_demo/src/latency_sensitivity_demo/scripts/dynamo_stack_sensitivity.sh
@@ -151,6 +151,44 @@ validate_config
 curl -sf http://localhost:2379/health >/dev/null 2>&1 || { echo "etcd not running. See header comment."; exit 1; }
 curl -sf http://localhost:8222/healthz >/dev/null 2>&1 || { echo "NATS not running. See header comment."; exit 1; }
 
+# ── Priority scheduling flags (SGLang version-dependent) ───────────────────────
+# dynamo.sglang passes these through to SGLang's argparse, so an unsupported flag
+# makes the worker exit during argument parsing before the model registers on
+# /v1/models. The supported set varies with the SGLang version bundled in the
+# installed Dynamo, so probe ServerArgs and include only what is available.
+PRIORITY_ARGS=()
+SGLANG_PRIORITY_PROBE="$(python3 - <<'PY' 2>/dev/null || true
+try:
+    from sglang.srt.server_args import ServerArgs
+    fields = getattr(ServerArgs, "__dataclass_fields__", {})
+    print("PROBE_OK")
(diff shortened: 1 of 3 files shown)


dagardner-nv (approved):

bbednarski9:
Thanks @dagardner-nv

PR #2070, opened 2026-06-12
fix(middleware): wire is_final enforcement with build-time validation and runtime call_next suppression

ericevans-nv:
## Summary

- `validate_middleware` was defined but never called; wired it into `configure_middleware` so `is_final` rules are enforced at workflow build time
- Fixed a logic bug in `validate_middleware` where the \"multiple finals\" error path was unreachable; rewrote with a two-pass approach and error messages that include the violating middleware name
- Added runtime `is_final` short-circuit in `FunctionMiddleware.function_middleware_invoke` and `function_middleware_stream`: when `is_final=True`, `call_next` is not invoked by the base implementation — subclasses must explicitly call it
- Threaded `is_final` through `DynamicFunctionMiddleware.__init__` so the flag can be forwarded via the constructor chain
- Updated `GuardrailsMiddleware` to explicitly declare `is_final=False` by default with a docstring explaining the effect of setting it to `True`
- Updated `middleware.md` to reflect build-time enforcement and runtime short-circuit behavior
- Added unit tests covering: final-not-last raises `ValueError`, multiple-finals raises `ValueError`, base invoke does not call `call_next` when `is_final=True`, base stream yields nothing when `is_final=True`, and non-final correctly calls `call_next`

## Summary by CodeRabbit

* **New Features**
  * Middleware can be marked as "final" to terminate the middleware chain; only one final middleware is allowed and it must be last.

* **Bug Fixes**
  * Validation now rejects multiple finals and finals not positioned last; final middleware default behavior halts further invocation unless explicitly overridden.
  * Middleware configuration now validates middleware sequences before use.

* **Documentation**
  * Clarified final-middleware semantics and Guardrails guidance.

* **Tests**
  * Added tests for final/non-final behavior and validation rules.

* **Chores**
  * Updated markdown link-check ignores for external docs.


docs/source/build-workflows/advanced/middleware.md
```diff
@@ -451,6 +451,16 @@ Configured field paths are validated at function registration (the earliest poin
 
 NeMo Guardrails sequences its own rails internally (all input rails, then all output rails, within a single `LLMRails` instance). Attach a **single** `guardrails` middleware to a given function and list every rail it needs inside that instance. Do **not** chain multiple `guardrails` middleware onto the same function: the outer instance would re-evaluate the inner instance's already-blocked refusal string, wasting rail calls and obscuring which rail blocked.
 
+`GuardrailsMiddleware` can be marked as a *final* middleware so that when it runs, `call_next` invokes the intercepted function directly — not another middleware. The following shows how to mark it as final:
```
mnajafian-nv: This section looks inconsistent with the current implementation. It says `GuardrailsMiddleware` can be marked final, but the built-in constructor always passes `is_final=False`, and I don’t see an `is_final` field on `GuardrailsMiddlewareConfig` or `DynamicMiddlewareConfig`. As written, users cannot enable this behavior from YAML. Please either expose the flag intentionally or reword this as guidance for custom Guardrails subclasses only.

packages/nvidia_nat_core/tests/nat/middleware/test_middleware_components.py
```diff
@@ -364,6 +400,102 @@ async def process(value: int) -> int:
             with pytest.raises(ValueError, match="Middleware `nonexistent_middleware` not found"):
                 await builder.add_function("test_func", config.functions["test_func"])
 
+    def test_validate_middleware_final_not_last_raises_error(self):
```
mnajafian-nv, line 403: Non-blocking: the direct `validate_middleware()` tests cover the helper well. Adding one build-path regression test would make this even stronger by proving invalid final-middleware ordering fails through the user-facing `Function.configure_middleware()` / workflow build path, which is the behavior described in the PR summary.

mnajafian-nv (approved):
conditional approval upon reviewing inline suggestions and internal alignment

PR #2105, opened 2026-07-08
fix memory leak lru cache

### LINKED ISSUE #2104: [Bug]: Severe Memory Leak in core utilities via @lru_cache on instance methods
### Version

0.1.dev

Source

### Describe the bug.

In Python, using `functools.lru_cache` on a class instance method creates a permanent, strong reference to `self` in the cache keys of the class-level function object. Because the cache persists for the lifetime of the program, the class instances can **never** be garbage collected.

Running static analysis (`ruff check --select B`) on the `NeMo-Agent-Toolkit` codebase reveals **25 occurrences** of this `B019` anti-pattern across critical core utilities:
1. `FunctionBase` in `function_base.py` (9 occurrences)
2. `DecomposedType` in `type_utils.py` (11 occurrences)
3. `TypeIntrospectionMixin` in `type_introspection_mixin.py` (5 occurrences)

Because `FunctionBase` and `DecomposedType` wrapper instances are created constantly during type inspection (which is used heavily in validation, routing, and telemetry during server operation), these leaked instances accumulate permanently in long-running processes like `nat mcp serve` or `nat serve`. By default, `lru_cache` locks the 128 most recent instances in memory per decorated property (i.e. `25 * 128 = 3200` permanent objects locked in memory).

**Expected Behavior:**
Core utility instances should be garbage collected once they go out of scope. Property wrappers should use `functools.cached_property` (which correctly stores the cache in the instance `__dict__` so it is garbage collected along with the instance), or remove the cache if the operation is cheap.

```shell
import gc
from pydantic import BaseModel
from nat.builder.function_base import FunctionBase

class DummyFunction(FunctionBase[str, str, str]):
    def __init__(self):
        super().__init__()

# Create 10,000 instances and access a cached property
for _ in range(10000):
    obj = DummyFunction()
    _ = obj.input_type
    
del obj

# Force garbage collection
gc.collect()

# Count how many objects are permanently locked in memory
alive_objects = sum(1 for o in gc.get_objects() if isinstance(o, DummyFunction))
print(f"Leaked instances: {alive_objects}") # Will print exactly 128
```

### Relevant log output

uv run ruff check --select B019 packages/nvidia_nat_core/src/nat

packages/nvidia_nat_core/src/nat/builder/function_base.py:94:14: B019 Use of `functools.lru_cache` or `functools.cache` on methods can lead to memory leaks
packages/nvidia_nat_core/src/nat/builder/function_base.py:121:14: B019 Use of `functools.lru_cache` or `functools.cache` on methods can lead to memory leaks
packages/nvidia_nat_core/src/nat/builder/function_base.py:146:14: B019 Use of `functools.lru_cache` or `functools.cache` on methods can lead to memory leaks
... (25 total B019 warnings)

</pre></details>

### Other/Misc.

**Proposed Solution:**
I can raise a PR to safely refactor these occurrences:
1. Replace all `@property` + `@lru_cache` combinations with `@functools.cached_property`.
2. For any remaining `@lru_cache` instances on plain methods (e.g. `_extract_input_output_types` or args-taking methods like `_get_union_info`), switch to using an internal instance dictionary cache (e.g., `self._cache[arg]`) or drop the cache entirely if the operation is cheap enough to rerun safely.

### Code of Conduct

pranaysb:
## Description
This PR resolves 25 critical `B019` memory leaks across the core utilities by replacing improper `@lru_cache` usage on instance methods, which permanently locks class instances (and dynamically generated Pydantic models) in memory by creating strong references to `self` in the cache keys. 

These leaked instances accumulate over time, creating a runaway memory footprint in long-running processes like `nat mcp serve` or `nat serve`. 

**Fix Details:**
1. **Replaced `@property` + `@lru_cache` with `@cached_property`**:
   In `type_utils.py` and `function_base.py`, the `DecomposedType` and `FunctionBase` classes had 20 different property methods caching their results using `@lru_cache`. We swapped all of these to `functools.cached_property`.
   
2. **Replaced `TypeIntrospectionMixin` caches with internal instance dictionaries**:
   In `type_introspection_mixin.py`, we removed the `@lru_cache` wrapper on all five methods (such as `_get_input_validator` and `_extract_input_output_types`). To ensure performance is maintained without leaking memory, we replaced them with manual, per-method instance dictionary caches (e.g., `self._input_validator_cache` or `self._get_union_info_cache`). This avoids recreating expensive Pydantic models via `create_model` on the hottest observability paths, while ensuring the cache is safely garbage collected with the instance.

3. **Fixed `test_type_introspection_mixin.py` test failure**:
   One unit test was explicitly trying to call `.cache_clear()` on one of the methods modified. Since the method is no longer wrapped in `lru_cache`, this raised an `AttributeError`. I removed the now-inapplicable `.cache_clear()` call; this is perfectly safe because the instance's cache is empty at this point regardless of the caching mechanism, so it doesn't weaken the test coverage at all.

Closes #2104

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/docs/source/resources/contributing/index.md).
- [x] We require that all contributors "sign-off" on their commits. This certifies that the contribution is your original work, or you have rights to submit it under the same license, or a compatible license.
- [x] When the PR is ready for review, new or existing tests cover these changes.
- [x] When the PR is ready for review, the documentation is up to date with these changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of type handling and schema generation by switching computed type/introspection results to per-instance, read-only caching.
  * Reduced the risk of stale or shared cached results when extracting input/output types and building validators.
* **Tests**
  * Updated type introspection tests to reflect the new per-instance caching behavior.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/builder/function_base.py
+++ b/packages/nvidia_nat_core/src/nat/builder/function_base.py
@@ -23,13 +23,13 @@ import logging
 import typing
 from abc import ABC
 from collections.abc import Callable
-from functools import lru_cache
 from types import NoneType
 
 from pydantic import BaseModel
 
 from nat.utils.type_converter import TypeConverter
 from nat.utils.type_utils import DecomposedType
+from nat.utils.type_utils import read_only_cached_property
 
 InputT = typing.TypeVar("InputT")
 StreamingOutputT = typing.TypeVar("StreamingOutputT")
@@ -92,8 +92,7 @@ class FunctionBase(typing.Generic[InputT, StreamingOutputT, SingleOutputT], ABC)
 
         self._converter: TypeConverter = TypeConverter(converters)
 
-    @property
-    @lru_cache
+    @read_only_cached_property
     def input_type(self) -> type[InputT]:
         """
         Get the input type of the function. The input type is determined by the generic parameters of the class.
@@ -120,8 +119,7 @@ class FunctionBase(typing.Generic[InputT, StreamingOutputT, SingleOutputT], ABC)
 
         raise ValueError("Could not find input schema")
 
-    @property
-    @lru_cache
+    @read_only_cached_property
     def input_class(self) -> type:
         """
         Get the python class of the input type. This is the class that can be used to check if a value is an instance of
@@ -146,8 +144,7 @@ class FunctionBase(typing.Generic[InputT, StreamingOutputT, SingleOutputT], ABC)
 
         return input_origin
 
-    @property
-    @lru_cache
+    @read_only_cached_property
     def input_schema(self) -> type[BaseModel]:
         """
         Get the Pydantic model schema for validating inputs. The schema must be pydantic models. This allows for
@@ -179,8 +176,7 @@ class FunctionBase(typing.Generic[InputT, StreamingOutputT, SingleOutputT], ABC)
         """
         return self._converter_list
 
-    @property
-    @lru_cache
+    @read_only_cached_property
     def streaming_output_type(self) -> type[StreamingOutputT]:
         """
         Get the streaming output type of the function. The streaming output type is determined by the generic parameters
@@ -208,8 +204,7 @@ class FunctionBase(typing.Generic[InputT, StreamingOutputT, SingleOutputT], ABC)
 
         raise ValueError("Could not find output schema")
 
-    @property
-    @lru_cache
+    @read_only_cached_property
     def streaming_output_class(self) -> type:
         """
         Get the python class of the output type. This is the class that can be used to check if a value is an instance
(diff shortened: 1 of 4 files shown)


packages/nvidia_nat_core/src/nat/builder/function_base.py
```diff
@@ -23,7 +23,7 @@
 import typing
 from abc import ABC
 from collections.abc import Callable
-from functools import lru_cache
+from functools import cached_property, lru_cache
```
willkill07:
```suggestion
from functools import cached_property
from functools import lru_cache
```

willkill07 (changes requested):
Formatting will definitely fail. I triggered CI, too.

willkill07:
/ok to test [7ca8319](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2105/commits/7ca831995b2396896ed25eebca22f230b6635961)

willkill07 (approved):

PR #2114, opened 2026-07-14
feat(core): add opt-in provider hooks for generated ids and timestamps

DABH:
## Description

The runtime currently stamps workflow runs, intermediate steps, spans, and function invocations with `uuid.uuid4()` and `time.time()` values called inline at each site (`runtime/runner.py`, `builder/context.py`, `data_models/intermediate_step.py`, `data_models/span.py`). Two otherwise identical runs therefore never produce the same identifiers or timings. That gets in the way of several third-party and testing use cases:

- Record or replay style testing, where a test harness replays a captured run and asserts on stable step identifiers.
- Golden-file trace comparison and reproducible traces in CI.
- Integrations with runtimes that re-execute workflow code (for example plugins that relocate or instrument function execution) and need identifiers to remain stable across re-executions, so steps and spans correlate across systems.

This PR adds a small additive module, `nat.utils.providers`, with process-wide, opt-in hooks:

- `set_id_provider` / `get_id_provider` — a zero-argument callable returning a new identifier as a canonical UUID string (default: `uuid.uuid4`).
- `set_time_provider` / `get_time_provider` — a zero-argument callable returning the current time in fractional seconds (default: `time.time`).
- Derivation helpers used by the runtime: `generate_id`, `generate_trace_id` (128-bit), `generate_span_id` (64-bit), `current_time`, and `current_time_ns`. Integer identifiers are derived by parsing the id provider's UUID string, so the default distribution is identical to `uuid.uuid4().int`.

The existing call sites are routed through these hooks:

- `Runner` workflow run id, trace id, and workflow step id (single and streaming paths).
- `Context.push_active_function` invocation id.
- `IntermediateStepPayload` `UUID` and `event_timestamp` default factories.
- `Span`/`SpanContext`/`SpanEvent` trace ids, span ids, and start/end/event timestamps.

Pydantic `default_factory` sites reference the module-level functions, so the installed provider is resolved lazily at model construction time rather than captured at import.

**Zero behavior change by default:** when no provider is installed, the defaults are `uuid.uuid4` and `time.time`, exactly as before. Setters return the previously installed provider so callers can restore it. The hook is documented briefly in the observability guide, and the module is picked up by the auto-generated API reference.

## Testing

- New `packages/nvidia_nat_core/tests/nat/utils/test_providers.py` (8 tests): default provider behavior, setter install/restore semantics, UUID-string contract for integer id derivation, lazy resolution through Pydantic default factories, span model injection, `push_active_function` injection, and an end-to-end minimal `Runner` run (single and streaming) asserting deterministic run ids, trace ids, step UUIDs, and timestamps.
- Full core suite: `uv run pytest packages/nvidia_nat_core` — 2725 passed, 54 skipped (defaults unchanged).
- `pre-commit run yapf|ruff-check --files <touched files>`, `python ci/scripts/copyright.py --verify-apache-v2`, `vale` on the touched Markdown, and `python ci/scripts/path_checks.py` all pass.

## Summary by CodeRabbit

* **New Features**
  * Added process-wide, opt-in hooks to customize workflow/function identifiers and timestamps.
  * Workflow runs, spans, intermediate steps, and active function invocations now use the installed providers for deterministic observability values.
  * Added validation to reject provider outputs that produce invalid/zero-derived trace and span identifiers.

* **Documentation**
  * Documented how to configure, restore, and override default ID/timestamp providers, including deterministic examples and ID derivation behavior.

* **Tests**
  * Added deterministic test coverage for default behavior, provider installation/restoration, validation failures, and end-to-end runtime output.

### DIFF
--- a/docs/source/run-workflows/observe/observe.md
+++ b/docs/source/run-workflows/observe/observe.md
@@ -310,7 +310,7 @@ This is useful when you call a remote workflow and want its steps to appear unde
 
 ## Deterministic Identifiers and Timestamps
 
-The runtime stamps workflow runs, intermediate steps, spans, and function invocations with generated identifiers (`uuid.uuid4`) and wall-clock timestamps (`time.time`). For reproducible runs — for example, record/replay style tests, golden-file trace comparison, or runtimes that re-execute workflow code and need identifiers to remain stable across re-executions — install process-wide providers that the runtime uses instead:
+The runtime stamps workflow runs, intermediate steps, spans, and function invocations with generated identifiers (`uuid.uuid4`) and wall-clock timestamps (`time.time`). For reproducible runs — for example, record or replay style tests, golden-file trace comparison, or runtimes that re-execute workflow code and need identifiers to remain stable across re-executions — install process-wide providers that the runtime uses instead:
 
 ```python
 import itertools
--- a/docs/source/run-workflows/observe/observe.md
+++ b/docs/source/run-workflows/observe/observe.md
@@ -307,3 +307,21 @@ Context.get().intermediate_step_manager.push_intermediate_steps(remote_intermedi
 ```
 
 This is useful when you call a remote workflow and want its steps to appear under the trace of the current workflow in your observability backend, so you get one connected tree for the full request.
+
+## Deterministic Identifiers and Timestamps
+
+The runtime stamps workflow runs, intermediate steps, spans, and function invocations with generated identifiers (`uuid.uuid4`) and wall-clock timestamps (`time.time`). For reproducible runs — for example, record/replay style tests, golden-file trace comparison, or runtimes that re-execute workflow code and need identifiers to remain stable across re-executions — install process-wide providers that the runtime uses instead:
+
+```python
+import itertools
+import uuid
+
+from nat.utils.providers import set_id_provider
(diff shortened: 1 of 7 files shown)


willkill07:
@DABH coderabbit feedback seems reasonable here. No need to address yet.

Otherwise, I'll kick off CI to see how things go. We can defer updates based on the initial CI trigger.

willkill07:
/ok to test [f4fafaf](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2114/commits/f4fafaf3ed104b73f687a43a45a45d295a4702c0)

willkill07:
@DABH in the documentation can you rephrase "record/replay" to "record or replay"?

Also, i do think addressing the coderabbit feedback should be done.

DABH:
Thanks so much for taking a look!! Yes - will address that change and the coderabbit feedback shortly.  Thanks again!

DABH:
@willkill07 Should be all set - thanks again for taking a look!

willkill07:
/ok to test [7f819a4](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2114/commits/7f819a4307fd9d4805bbcbd6191f75166c7a523a)

willkill07 (approved):

PR #2115, opened 2026-07-14
fix(core): resolve string annotations against the defining module in FunctionInfo

DABH:
## Description

`FunctionInfo` breaks for components defined in modules that use `from __future__ import annotations`. With the future import, annotations are stored as strings, and `FunctionDescriptor.from_function()` / `FunctionInfo.from_fn()` read them raw off `inspect.signature(...)` and interpolate them into the annotations of wrappers synthesized inside `nat.builder.function_info` (for example the auto-generated streaming wrapper and the multi-argument input adapter). The strings are then resolved against the synthesizing module's globals instead of the defining module's, so plugin authors hit:

- `NameError: name '<UserType>' is not defined` from `typing.get_type_hints` on the synthesized wrapper for any user-defined input/output type.
- Generated input schemas containing unresolved `ForwardRef('<UserType>')` field annotations.
- `BaseModel` outputs wrapped in a spurious `OutputArgsSchema` value-wrapper instead of being used directly.
- Silent loss of `Streaming` metadata introspection on annotated async-generator returns (`typing.get_origin` of a string is `None`).

### Fix

Resolve string annotations at the inspection boundary with `inspect.signature(..., eval_str=True)`, which evaluates them against the defining function's own globals. Two call sites change: `FunctionDescriptor.from_function()` and the async-generator branch of `FunctionInfo.from_fn()`. Non-string annotations are passed through unchanged, so modules that do not use the future import are unaffected.

### Reproduction

The new regression test module `packages/nvidia_nat_core/tests/nat/builder/test_function_info_future_annotations.py` uses `from __future__ import annotations` itself. Before the fix, its four tests fail with exactly the symptoms above; after the fix they pass.

## Testing

- New regression tests: descriptor inspection, single-function path (including the auto-synthesized streaming wrapper), stream-function path, and multi-argument input schema generation — 4 passed (all fail before the fix).
- `uv run pytest packages/nvidia_nat_core` — 2721 passed, 54 skipped.
- `uv run pytest packages/nvidia_nat_langchain packages/nvidia_nat_security packages/nvidia_nat_a365` (packages containing registration modules that already use the future import) — 1035 passed, 43 skipped.
- `pre-commit run yapf|ruff-check --files <touched files>` and `python ci/scripts/copyright.py --verify-apache-v2` pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of functions using postponed or forward-referenced type annotations.
  - Correctly derives input/output types and schemas for regular, streaming, and multi-argument functions.
  - Preserves field-level annotation details in generated schemas.

- **Tests**
  - Added regression coverage for annotation resolution across supported function patterns.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/builder/function_info.py
+++ b/packages/nvidia_nat_core/src/nat/builder/function_info.py
@@ -214,7 +214,10 @@ class FunctionDescriptor:
 
         converters = []
 
-        sig = inspect.signature(func)
+        # Resolve string annotations (e.g. from modules using `from __future__ import annotations`) against the
+        # defining module's namespace. The inspected types are interpolated into the annotations of wrappers
+        # synthesized in this module, so unresolved strings would later be evaluated against the wrong globals.
+        sig = inspect.signature(func, eval_str=True)
 
         arg_count = len(sig.parameters)
 
@@ -582,7 +585,9 @@ class FunctionInfo:
         if (inspect.isasyncgenfunction(fn)):
             stream_fn = fn
 
-            sig = inspect.signature(fn)
+            # Resolve string annotations against the defining module's namespace so the return annotation can be
+            # introspected for `Streaming` metadata regardless of `from __future__ import annotations`.
+            sig = inspect.signature(fn, eval_str=True)
 
             output_origin = typing.get_origin(sig.return_annotation)
             output_args = typing.get_args(sig.return_annotation)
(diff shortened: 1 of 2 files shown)


willkill07:
/ok to test [60b33d7](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2115/commits/60b33d72fa0e9f2c1b6fd405461d398ad7e36e45)

willkill07 (approved):

PR #2118, opened 2026-07-14
fix(core): resolve lru_cache memory retention in DiscoveryMetadata

### LINKED ISSUE #2117: [Bug]: Severe Memory Leak in DiscoveryMetadata via @lru_cache on static methods
### Version

develop

Source

### Describe the bug.

There is a severe memory retention issue in `packages/nvidia_nat_core/src/nat/data_models/discovery_metadata.py` caused by a rogue `@lru_cache` decorator on the `get_distribution_name_from_config_type(config_type)` and `get_distribution_name_from_module(module)` static methods.
Because a bare `@lru_cache` defaults to `maxsize=128`, this doesn't leak *every* dynamic class ever created — but it does permanently pin the 128 most-recently-seen classes at any given time, evicting them only when 128 *newer* distinct classes come through. Since `inspect.getmodule()` is ~0.18µs, this caching buys negligible performance while needlessly extending the lifetime of dynamically generated classes in long-running processes like `nat mcp serve`.
Expected Behavior:
Classes should be gracefully garbage collected once they go out of scope. Any caches storing class/module references should use weak references (such as `weakref.WeakKeyDictionary`) rather than strong references so the garbage collector can properly clean them up.

```shell
import gc
from pydantic import BaseModel, create_model
from nat.data_models.discovery_metadata import DiscoveryMetadata
# Create 10,000 dynamic models and pass them through the cached method
for i in range(10000):
    DynamicModel = create_model(f"DynamicTestModel_{i}", field=(str, ...))
    _ = DiscoveryMetadata.get_distribution_name_from_config_type(DynamicModel)
del DynamicModel
# Force garbage collection
gc.collect()
# Count how many objects are permanently locked in memory
# The most recent 128 dynamic classes remain locked in memory forever.
alive_objects = sum(1 for o in gc.get_objects() if isinstance(o, type) and o.__name__.startswith("DynamicTestModel"))
print(f"Leaked instances: {alive_objects}") # Will print exactly 128
```

### Relevant log output

</pre></details>
Since this is a silent memory leak and doesn't throw a stack trace

### Other/Misc.

(Note: The newly enforced B019 ruff rule successfully catches lru_cache memory leaks on regular instance methods, but it explicitly does not flag @staticmethods. Because both of the leaking methods in DiscoveryMetadata were static methods, this leak slipped past both the original sweep and the new lint enforcement).

Proposed Solution: I can raise a PR to safely refactor these occurrences:

Replace the @lru_cache decorators with module-level weakref.WeakKeyDictionary caches. Because these dictionaries hold "weak" references to the dynamically generated Pydantic models (and modules), they allow the garbage collector to automatically clean up the classes when they are no longer needed while maintaining exact cache performance.

### Code of Conduct

pranaysb:
## Description

This PR resolves a memory retention issue in `DiscoveryMetadata` caused by bare `@lru_cache` decorators (which default to `maxsize=128`). While this doesn't cause an unbounded leak, it permanently pins the 128 most-recently-seen dynamic Pydantic models (and modules) in memory, evicting them only when 128 newer distinct classes come through. Since the cached operation (`inspect.getmodule()`) is extremely cheap (~0.18µs), this buys negligible performance while needlessly extending the lifetime of dynamically generated classes in long-running processes like `nat mcp serve`.

*(Note: The newly enforced `B019` ruff rule successfully catches `lru_cache` memory leaks on regular instance methods, but it explicitly does not flag `@staticmethod`s. Because both of the leaking methods in `DiscoveryMetadata` were static methods, this leak slipped past both the original sweep and the new lint enforcement).*

### Fix Details:

**Replaced `@lru_cache` with `WeakKeyDictionary` in `discovery_metadata.py`:**
To ensure the caching performance is maintained without leaking memory (following the same spirit as my PR #2105 — replacing unsafe `@lru_cache` with GC-safe caching — but using `WeakKeyDictionary` here since the keys are classes/modules, not `self`), we implemented a safe caching strategy for both `get_distribution_name_from_config_type` and `get_distribution_name_from_module`. We replaced the problematic decorators with module-level `weakref.WeakKeyDictionary` caches. Because these dictionaries hold "weak" references to the dynamically generated Pydantic models (and modules), they allow the garbage collector to automatically clean up the classes when they are no longer needed.

**Addressed CodeRabbit typing/docstring feedback in `type_utils.py`:**
Added proper Google-style docstrings and explicit type hints (`typing.Any`) to the `read_only_cached_property` descriptor to fully satisfy the public API contract requirements flagged during my PR #2105 review.

Closes #2117

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/docs/source/resources/contributing/index.md).
- [x] We require that all contributors "sign-off" on their commits. This certifies that the contribution is your original work, or you have rights to submit it under the same license, or a compatible license.
- [x] When the PR is ready for review, new or existing tests cover these changes.
- [x] When the PR is ready for review, the documentation is up to date with these changes.

## Summary by CodeRabbit

* **Performance**
  * Improved package metadata resolution by using a more memory-friendly caching approach for repeated distribution-name lookups.

* **Reliability**
  * Enhanced distribution-name resolution across both module and configuration types with safer handling of edge cases.

* **Documentation**
  * Added clearer documentation and stricter type annotations for the read-only cached property utility, improving maintainability.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/data_models/discovery_metadata.py
+++ b/packages/nvidia_nat_core/src/nat/data_models/discovery_metadata.py
@@ -135,7 +135,6 @@ class DiscoveryMetadata(BaseModel):
         return "nvidia-nat"
 
     @staticmethod
-    @lru_cache
     def get_distribution_name_from_config_type(config_type: type["TypedBaseModelT"]) -> str:
         """Get the distribution name from the config type using the mapping of module names to distro names.
 
--- a/packages/nvidia_nat_core/src/nat/data_models/discovery_metadata.py
+++ b/packages/nvidia_nat_core/src/nat/data_models/discovery_metadata.py
@@ -17,6 +17,7 @@ import importlib.metadata
 import inspect
 import logging
 import typing
+import weakref
 from enum import StrEnum
 from functools import lru_cache
 from types import ModuleType
@@ -34,6 +35,9 @@ if TYPE_CHECKING:
     from nat.data_models.common import TypedBaseModelT
 
 logger = logging.getLogger(__name__)
+
+_module_distro_cache = weakref.WeakKeyDictionary()
+_config_type_distro_cache = weakref.WeakKeyDictionary()
 
 
 class DiscoveryStatusEnum(StrEnum):
@@ -100,7 +104,6 @@ class DiscoveryMetadata(BaseModel):
         return distro_name if distro_name else root_package_name
 
     @staticmethod
-    @lru_cache
     def get_distribution_name_from_module(module: ModuleType | None) -> str:
         """Get the distribution name from the config type using the mapping of module names to distro names.
 
@@ -114,6 +117,12 @@ class DiscoveryMetadata(BaseModel):
 
         if module is None:
             return "nvidia-nat"
+
+        try:
+            if module in _module_distro_cache:
+                return _module_distro_cache[module]
+        except TypeError:
+            pass
 
         # Get the mapping of module names to distro names
         mapping = get_all_entrypoints_distro_mapping()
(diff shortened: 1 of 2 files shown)


willkill07:
@pranaysb please sign-off your commit(s)

willkill07:
/ok to test [ba5b856](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2118/commits/ba5b8563b13e2d52cf7532ba994e41457dfe7f61)

willkill07 (approved):

willkill07:
@pranaysb thank you! I've kicked off CI. Once it passes, I will merge :)

If it does happen to fail, please address the failures.

PR #2121, opened 2026-07-15
fix(mcp): recover the MCP client when the lifecycle worker dies

### LINKED ISSUE #2111: A single transient transport error permanently bricks a reused remote mcp_client (streamable-http) for the process
**Version:** develop @ e8692d00 (v1.9.0.dev)

**Describe the bug**

When a remote `mcp_client` (streamable-http) function group is reused across multiple tool calls in one long-lived process (for example an agent or served workflow that invokes an MCP-backed tool more than once), a single transient transport error permanently disables the shared client for the rest of the process. The trigger can be an `httpx.ConnectError`, a dropped keep-alive, or a server 5xx surfaced as a stream error, raised inside the MCP SDK's background POST task. After it fires, every subsequent tool call fails, and the process emits `RuntimeError: Attempted to exit cancel scope in a different task than it was entered in` at shutdown.

This is not a server concurrency issue, and it does not require concurrency to trigger. It reproduces with strictly sequential, single-task calls and a single injected failure.

**Root cause**

In `packages/nvidia_nat_mcp/src/nat/plugins/mcp/client/client_base.py`, the base client spawns a long-lived lifecycle worker task (call it L, `:208`) and, after connecting, parks it on `await self._lifecycle_commands.get()` (`:302`). Meanwhile the MCP SDK transport's `anyio` task group (from `streamable_http_client`) and its per-request child POST tasks remain live inside the suspended `connect_to_server` generator (`:702`).

When a transport background child task (`streamable_http.py` `tg.start_soon(handle_request_async)`, `:568-569`, unguarded) raises, anyio cancels the task group's cancel scope, which cancels the group's host task. That host task is L, wherever it is parked. L dies with `CancelledError` before any `reconnect` command can be serviced. From then on:

- `_run_lifecycle_command` sees `_lifecycle_task.done()` and raises `RuntimeError("MCPBaseClient not initialized...")` (`:289-290`), so reconnect can never succeed.
- `_connection_established` is left stale-True (the cancellation bypassed the state clearing in `_close_connection`), so the `is_connected` guard does not short-circuit. Each call proceeds into `call_tool`, the reconnect fails, and `MCPToolClient.acall` returns the plain string `"MCPToolClient tool call failed: ..."` (`:837`).
- At interpreter or async-gen shutdown the orphaned `connect_to_server` generator is finalized in a different task, exiting the transport and `ClientSession` cancel scopes off-task, which produces the "cancel scope in a different task" error.

**To Reproduce (deterministic, no concurrency)**

1. Build a `WorkflowBuilder` with a remote streamable-http `mcp_client` group (any reachable MCP server).
2. Make one successful tool call.
3. Inject a single `httpx.ConnectError` into the next POST (for example, monkeypatch the transport once), then restore the network.
4. Observe: the in-flight call is orphaned, `_lifecycle_task` is now cancelled, and all subsequent calls fail with `MCPBaseClient not initialized` even though the network is healthy again. The client never recovers within the process.

A natural, non-injected reproduction is to run several sequential MCP-backed tool calls through one persistent group over a few minutes. A real transient blip eventually triggers the same permanent brick.

**Expected behavior**

A transient transport error should, at worst, fail the in-flight call. The shared client should then recover (reconnect) for subsequent calls, or the lifecycle worker should not be cancellable by the transport's background task group.

**Generality (server- and transport-agnostic)**

- Not specific to any server. The defect is in NAT's server-agnostic `MCPBaseClient`. In the repro, a stateless remote server answered normally immediately before the injected client-side error.
- All three transports share the vulnerable machinery. `MCPStreamableHTTPClient` (`client_base.py:628`), `MCPSSEClient` (`:512`), and `MCPStdioClient` (`:561`) all subclass `MCPBaseClient` and override only `connect_to_server`. Each enters an MCP SDK transport that opens `anyio.create_task_group()` with `start_soon` background tasks (`streamable_http.py:647`, `sse.py:62`, `stdio/__init__.py:183`). A raise in any of those background children cancels the group scope and therefore the parked lifecycle worker. Streamable-http is confirmed by a live repro; SSE and stdio are confirmed structurally (same inherited lifecycle). The stdio case (subprocess crash or broken pipe) is the same class #1935 targeted, still uncovered for the parked-worker path.
- All session modes affected. The base client, `session_aware_tools` per-session, and `per_user_mcp_client` all spawn the same parked worker. Only `session_aware` mode has any within-process recovery: a bricked per-session client self-heals after `session_idle_timeout` (default 1h) evicts and lazily recreates it. The shared base client and per-user clients are bricked for the life of the process.

**Impact (High)**

- Hits NAT on its defaults: the default `streamable-http` transport, the default no-auth shared base client (which serves all requests and users in a `nat serve` deployment), and despite the default-on `reconnect_enabled`, which cannot recover it because reconnect routes through the same dead worker.
- The trigger is common, not an edge case. The transport keeps a long-lived idle background stream open. Servers, load balancers (for example AWS ALB idle around 60s), reverse proxies (nginx keepalive around 75s), and API gateways reap idle connections routinely. Any agent that spaces out tool calls (LLM thinking, multi-turn gaps, minutes-long tool jobs) leaves that stream idle past the shortest timeout in the path, so the next server-side reset triggers the brick. Over hours or days of uptime this is a when-not-if event.
- User experience: after one blip, every subsequent MCP tool call fails for the rest of the process with `MCPBaseClient not initialized`. ReAct agents retry-loop on it, and only a process restart recovers. In `nat serve`, MCP tools silently stop working for all users until an operator restarts.
- Unaffected: one-shot `nat run` (single call then exit), local stdio servers in practice (no HTTP idle-reap trigger, only a subprocess crash would hit it), and short remote flows with no idle gap. There is no data loss or corruption, which is why this is High and not Critical.

**Suggested fix directions** (in `client_base.py`)

1. Isolate the transport context in a dedicated task that catches child-task failures and converts them into a "connection lost" state (reset `_connection_established` and `_session`) instead of letting the cancel propagate into the parked command loop.
2. Make recovery possible: if `_lifecycle_task` is done or cancelled, restart a fresh lifecycle worker in `_reconnect` and `_run_lifecycle_command` rather than permanently raising `MCPBaseClient not initialized`, and reset `is_connected` when the worker dies.

**Note on related fix**

Commit `d2b17c9b` (#1935, "Fix MCP reconnect lifecycle after local server crash") addressed the caller-observed reconnect path. Its regression tests (`test_mcp_client_base.py`) simulate failures at the caller and `list_tools` site, not a failure originating in a transport background child task while the lifecycle worker is parked. This failure mode is uncovered.

Hokonoken:
## Description
Closes #2111

A transient transport failure permanently bricks a shared MCP client. The failure chain (confirmed by construction in [this comment](https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues/2111#issuecomment-4957979762)): the SDK's streamable-http transport starts `handle_request_async` unguarded on the transport task group (`streamable_http.py:569` in `mcp` 1.28.1), so a failing request-child cancels the lifecycle worker that hosts the transport's task group. That cancellation bypasses the worker's `except Exception` (`CancelledError` is not caught), so `_connection_established` stays stale-`True` while the worker is gone; every lifecycle command then raises `MCPBaseClient not initialized` and `_reconnect()` loops on the same error forever.

This implements the issue's **direction 2** (make the worker relaunchable) as the minimal user-facing fix:

- `is_connected` returns `False` once the worker task is done — no more stale-`True` after a transport-child cancellation.
- `_run_lifecycle_command` relaunches a dead worker via the new `_ensure_lifecycle_worker()`, which abandons the orphaned transport refs rather than closing them off-task (closing `_exit_stack` off-task raises "cancel scope in a different task").

The regression test's mock client mirrors the real transports' structure — a real `anyio.create_task_group()` with an unguarded background child — which is the path existing mocks don't exercise (why this was uncovered, as the issue notes).

**Verification (executed):**
- Regression test fails on unfixed source, passes with the fix (cross-checked by stashing only the source change).
- Full `test_mcp_client_base.py` suite: 33 passed, 8 skipped.
- `ruff` (0.15.0) and `yapf` (0.43.0) clean at the repo's pinned versions; copyright check clean (`ci/scripts/copyright.py --git-diff-commits`).

**Known limit (deliberate):** recovery abandons the dead worker's transport context instead of cleaning it up. Against the real streamable-http transport, that orphaned context can still surface the off-task cancel-scope complaint at teardown/GC (the regression test's mock does not reproduce that warning). Isolating the transport in a dedicated task — the issue's direction 1 — is the structural follow-up; happy to take that on separately if maintainers want it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated connection status to accurately report disconnected state when the underlying lifecycle worker is missing or stopped.
  - Improved resilience by recreating the worker and clearing stale session/tool state before executing new lifecycle commands.
  - Prevented stuck/hanging lifecycle commands; failures now raise a retryable connection error and the client can recover via reconnection.
- **Tests**
  - Added regression coverage for transport task failures, mid-command worker death, and successful recovery without “bricking” the client.

### DIFF
--- a/packages/nvidia_nat_mcp/src/nat/plugins/mcp/client/client_base.py
+++ b/packages/nvidia_nat_mcp/src/nat/plugins/mcp/client/client_base.py
@@ -291,7 +291,7 @@ class MCPBaseClient(ABC):
             if last_error:
                 raise last_error
 
-    def _ensure_lifecycle_worker(self) -> None:
+    def _ensure_lifecycle_worker(self) -> asyncio.Task:
         """Respawn the lifecycle worker if it died, so the client can recover (#2111).
 
         A failing transport background child task cancels the worker that hosts the
@@ -299,11 +299,13 @@ class MCPBaseClient(ABC):
         loop, so it never clears connection state and the task is left ``done``. Rather
         than bricking the client for the rest of the process, drop the orphaned transport
         references and start a fresh worker that the next command can drive to reconnect.
+
+        Returns the live worker task, so callers can watch it alongside their command.
         """
         if self._lifecycle_commands is None:
             raise RuntimeError("MCPBaseClient not initialized. Use async with to initialize.")
         if self._lifecycle_task is not None and not self._lifecycle_task.done():
-            return
+            return self._lifecycle_task
         # The previous worker's transport context was entered in that (now dead) task.
         # Do not close _exit_stack here: closing it off-task raises "cancel scope in a
         # different task". Drop the stale references and let a fresh connect start clean.
@@ -312,14 +314,26 @@ class MCPBaseClient(ABC):
         self._tools = None
         self._connection_established = False
         self._lifecycle_task = asyncio.create_task(self._lifecycle_worker(), name=f"mcp-client-{self.server_name}")
+        return self._lifecycle_task
 
     async def _run_lifecycle_command(self, command: str) -> None:
         """Run a connection lifecycle command in the task that owns the transport stack."""
-        self._ensure_lifecycle_worker()
+        worker = self._ensure_lifecycle_worker()
 
         future: asyncio.Future[None] = asyncio.get_running_loop().create_future()
         await self._lifecycle_commands.put((command, future))
-        await future
+        # The worker can die between the liveness check above and resolving the
+        # future (the same transport-child cancellation this class recovers from),
+        # which would leave the future unresolved forever. Watch both.
+        await asyncio.wait({future, worker}, return_when=asyncio.FIRST_COMPLETED)
+        if future.done():
+            future.result()
+            return
+        # The worker died before resolving our command. Abandon the queued command
+        # (the worker loop skips commands whose future is already done) and surface
+        # a retryable error so _reconnect / the next call can relaunch a worker.
+        future.cancel()
+        raise ConnectionError(f"MCP client lifecycle worker for {self.server_name} exited while running '{command}'")
 
     async def _lifecycle_worker(self) -> None:
         """Own MCP transport context entry and exit to keep AnyIO cancel scopes task-local."""
@@ -328,6 +342,11 @@ class MCPBaseClient(ABC):
 
         while True:
             command, future = await self._lifecycle_commands.get()
+
+            if future.done():
+                # The caller abandoned this command (its previous worker died before
+                # consuming it). Executing it now would replay a stale request.
+                continue
 
             if command == "close":
                 try:
--- a/packages/nvidia_nat_mcp/src/nat/plugins/mcp/client/client_base.py
+++ b/packages/nvidia_nat_mcp/src/nat/plugins/mcp/client/client_base.py
@@ -239,6 +239,13 @@ class MCPBaseClient(ABC):
     @property
     def is_connected(self) -> bool:
         """Whether the client has an active, initialized connection."""
+        # A transport background child task can cancel the lifecycle worker
+        # (which hosts the transport's task group) without running the worker's
+        # state-clearing, leaving _connection_established stale-True (#2111).
+        # Treat a dead worker as not connected so callers do not proceed into
+        # tool calls against a torn-down transport.
+        if self._lifecycle_task is None or self._lifecycle_task.done():
+            return False
         return self._exit_stack is not None and self._connection_established
 
     @property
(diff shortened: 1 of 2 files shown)


Hokonoken:
Both review findings addressed in 233e807:

- **Major (indefinite wait past the liveness check):** `_run_lifecycle_command` now watches the worker task alongside the command future (`asyncio.wait`, first-completed). If the worker dies first, the command fails with a retryable `ConnectionError` and the abandoned queue entry is skipped by the next worker instead of being replayed. New regression test `test_worker_death_mid_command_raises_instead_of_hanging` pins exactly that race (a mock whose connect parks lets the test cancel the worker while the future is pending) — verified both ways: without the guard the await hangs and the test fails via its bounded `fail_after` timeout; with it, the command raises and the client recovers.
- **Minor (fixed 50 ms sleep):** replaced with a bounded deterministic wait on `worker_before.done()` under `anyio.fail_after(5)`.

Full `test_mcp_client_base.py`: 34 passed, 8 skipped; ruff/yapf clean at the pinned versions.

willkill07:
/ok to test [233e807](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2121/commits/233e8070a4707d4351cd6e59d2816b8bf36b59af)

willkill07 (approved):
This robustness improvement is welcomed. Thank you!

PR #2128, opened 2026-07-16
fix(profiler): import the CrewAI profiler handler from its real module

### LINKED ISSUE #2127: [Bug]: CrewAI profiler never installs - framework_wrapper imports nat.plugins.crewai.callback_handler, module is crewai_callback_handler
### Version

1.6.0 (also present on `develop` as of 2026-07-15)

PyPi, Source

### Describe the bug.

`set_framework_profiler_handler` imports the CrewAI profiler from a module path that does not exist. `ModuleNotFoundError` subclasses `ImportError`, so the surrounding `except ImportError` swallows it and logs a message blaming missing extras. **CrewAI agents therefore get no instrumentation at all**, regardless of `framework_wrappers=[LLMFrameworkEnum.CREWAI]` — no LLM events, no tool events, no tokens.

Verified against `develop`:

- `packages/nvidia_nat_profiler/src/nat/plugins/profiler/decorators/framework_wrapper.py`
  → `from nat.plugins.crewai.callback_handler import CrewAIProfilerHandler`
- `packages/nvidia_nat_crewai/src/nat/plugins/crewai/callback_handler.py` → **does not exist (404)**
- `packages/nvidia_nat_crewai/src/nat/plugins/crewai/crewai_callback_handler.py` → exists, defines `CrewAIProfilerHandler`

```python
try:
    from nat.plugins.crewai.callback_handler import CrewAIProfilerHandler   # no such module
    ...
except ImportError:
    logger.warning("CrewAI profiler not available. Install NAT with CrewAI extras: "
                   'pip install "nvidia-nat[crewai]". Error: %s', e)
```

The `langchain`, `llama_index` and `semantic_kernel` branches all import `<plugin>.callback_handler` correctly — CrewAI is the only plugin whose module carries the extra prefix.

**Expected:** the CrewAI profiler installs and CrewAI agents report LLM/tool events.
**Actual:** it never installs; a warning claims the extras are missing when they are installed.

Also note `nat/plugins/crewai/register.py` does not import its callback handler, unlike the `semantic_kernel` plugin's `register.py`.

```shell
# With nvidia-nat[crewai] installed:

python -c "
import importlib
for m in ['nat.plugins.crewai.callback_handler', 'nat.plugins.crewai.crewai_callback_handler']:
    try:
        importlib.import_module(m); print('OK   ', m)
    except Exception as e:
        print('FAIL ', m, '->', type(e).__name__)
"
# FAIL  nat.plugins.crewai.callback_handler  -> ModuleNotFoundError
# OK    nat.plugins.crewai.crewai_callback_handler

# Then run any workflow whose function is registered with
# @register_function(config_type=..., framework_wrappers=[LLMFrameworkEnum.CREWAI])
# and subscribe to the IntermediateStep stream: zero CrewAI LLM/tool events arrive,
# and the warning below is logged.
```

### Relevant log output

WARNING nat.plugins.profiler.decorators.framework_wrapper: CrewAI profiler not available.
Install NAT with CrewAI extras: pip install "nvidia-nat[crewai]".
Error: No module named 'nat.plugins.crewai.callback_handler'

(nvidia-nat-crewai 1.6.0 and crewai 0.203.2 are both installed.)

</pre></details>

### Other/Misc.

**Impact.** The warning actively misdirects: it names a cause (missing extras) that is not the cause, so the natural response is to reinstall the extras and conclude the toolkit is fine. Meanwhile CrewAI reports zero tokens and zero tool calls, which in a profiler reads as *free* rather than *unmeasured*.

**Suggested fix.**

```python
from nat.plugins.crewai.crewai_callback_handler import CrewAIProfilerHandler
```

Or rename the module to `callback_handler.py` for consistency with the other three plugins. Catching `ModuleNotFoundError` separately from `ImportError` would also stop a genuine bug from being reported as a missing optional dependency.

**A second fault sits behind this one.** Importing the real module and calling `CrewAIProfilerHandler().instrument()` by hand does restore CrewAI's `LLM_START` and `TOOL_START` events — but the matching `*_END` events still never arrive. They appear to fire from CrewAI's worker thread while the step manager rides a `ContextVar` that doesn't follow it; the logs fill with `Step id ... not found in outstanding start steps. This may occur if the step was started in a different context or already completed.` Since `usage_info` only rides on `LLM_END`, token counts remain unavailable even with the import fixed. Flagging it here because fixing the import alone will not restore token capture — it may deserve its own issue.

This is unrelated to #1802 (`_llm_call_monkey_patch` breaking with crewai >= 1.1.0); this reproduces on crewai 0.203.2, where the handler itself is fine but is never loaded.

### Code of Conduct

- [X] I agree to follow the NeMo Agent Toolkit Code of Conduct
- [X] I have searched the [open bugs](https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

EnesYilmazcode:
`set_framework_profiler_handler` imported the CrewAI handler from a module that does not exist, so the `ModuleNotFoundError` was swallowed as a missing-extra warning and CrewAI agents got no instrumentation, even with the crewai extra installed.

## Description
The profiler's CrewAI branch imported `CrewAIProfilerHandler` from `nat.plugins.crewai.callback_handler`, but the crewai plugin ships the handler in `crewai_callback_handler.py`. The langchain, llama_index and semantic_kernel branches in the same file import `<plugin>.callback_handler`, which those plugins provide; crewai is the only one whose module carries a prefix. `ModuleNotFoundError` subclasses `ImportError`, so the surrounding `except ImportError` swallowed the failure and logged a warning to install the crewai extra while CrewAI agents received no LLM events, no tool events, and no tokens, even with `nvidia-nat[crewai]` installed.

Closes [#2127](https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues/2127)

## Changes
- Import `CrewAIProfilerHandler` from `nat.plugins.crewai.crewai_callback_handler` in `framework_wrapper.py`
- Add a regression test that registers a stand-in handler at the real module path and asserts the CrewAI branch loads and instruments it

## Validation
- [x] `pytest packages/nvidia_nat_profiler/tests/decorators/test_framework_wrapper.py` passes with the fix
- [x] Without the fix the test fails (`instrument_calls == []`) and logs the reported `CrewAI profiler not available ... No module named 'nat.plugins.crewai'` warning
- [x] `ruff check` clean on both files
- [x] No other reference to the old `nat.plugins.crewai.callback_handler` path remains in the repo

## Note
The reporter also flagged that CrewAI `*_END` events are dropped because the step manager's `ContextVar` does not follow CrewAI's worker thread, so token capture stays broken after the import is fixed. That is a separate defect and is left for its own issue, as they suggested.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CrewAI framework profiling to load the correct profiler handler, ensuring instrumentation is applied as expected during initialization.

* **Tests**
  * Added an asynchronous pytest that verifies the CrewAI profiling path selects the proper handler and that instrumentation becomes active when the wrapped async context runs.

### DIFF
--- a/packages/nvidia_nat_profiler/src/nat/plugins/profiler/decorators/framework_wrapper.py
+++ b/packages/nvidia_nat_profiler/src/nat/plugins/profiler/decorators/framework_wrapper.py
@@ -109,7 +109,7 @@ def set_framework_profiler_handler(
 
             if LLMFrameworkEnum.CREWAI in frameworks and not _library_instrumented["crewai"]:
                 try:
-                    from nat.plugins.crewai.callback_handler import CrewAIProfilerHandler
+                    from nat.plugins.crewai.crewai_callback_handler import CrewAIProfilerHandler
                     handler = CrewAIProfilerHandler()
                     handler.instrument()
                     _library_instrumented["crewai"] = True
--- a/packages/nvidia_nat_profiler/tests/decorators/test_framework_wrapper.py
+++ b/packages/nvidia_nat_profiler/tests/decorators/test_framework_wrapper.py
@@ -1,0 +1,48 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
+from types import ModuleType
+
+from nat.builder.framework_enum import LLMFrameworkEnum
+from nat.plugins.profiler.decorators import framework_wrapper
+
+
+async def test_crewai_profiler_loads_handler_from_real_module(monkeypatch):
+    """The CrewAI branch must load the profiler handler from crewai_callback_handler, not callback_handler."""
+    instrument_calls = []
+
+    class _StubHandler:
+
+        def instrument(self):
+            instrument_calls.append(True)
+
+    stub_module = ModuleType("nat.plugins.crewai.crewai_callback_handler")
+    stub_module.CrewAIProfilerHandler = _StubHandler
+    monkeypatch.setitem(sys.modules, "nat.plugins.crewai.crewai_callback_handler", stub_module)
+    monkeypatch.setitem(framework_wrapper._library_instrumented, "crewai", False)
+
+    @asynccontextmanager
+    async def build_fn(config, builder) -> AsyncIterator[str]:
+        yield "workflow"
+
+    wrapped = framework_wrapper.set_framework_profiler_handler(None, [LLMFrameworkEnum.CREWAI])(build_fn)
+    async with wrapped(None, None) as result:
+        assert result == "workflow"
+
+    assert instrument_calls == [True]
+    assert framework_wrapper._library_instrumented["crewai"] is True


willkill07 (approved):

willkill07:
/ok to test [0462aaf](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2128/commits/0462aaffd8d7c6310fe3260afff393d07a8adf29)

willkill07:
@EnesYilmazcode pre-commit is failing

willkill07 (approved):

PR #2132, opened 2026-07-19
fix(profiler): extract LlamaIndex OpenAI token usage

### LINKED ISSUE #2126: [Bug]: LlamaIndexProfilerHandler._extract_token_usage reports 0 tokens for OpenAI-compatible clients
### Version

1.6.0 (also present on `develop` as of 2026-07-15)

PyPi

### Describe the bug.

`LlamaIndexProfilerHandler._extract_token_usage` looks for a `"usage"` key inside `response.additional_kwargs` and reads `input_tokens`/`output_tokens` off it. For LlamaIndex's OpenAI-compatible clients neither assumption holds, so it returns zeros — without raising.

`packages/nvidia_nat_llama_index/src/nat/plugins/llama_index/callback_handler.py`:

```python
if response and response.additional_kwargs and "usage" in response.additional_kwargs:
    usage = response.additional_kwargs["usage"] if "usage" in response.additional_kwargs else {}
    token_usage.prompt_tokens = usage.input_tokens if hasattr(usage, "input_tokens") else 0
    token_usage.completion_tokens = usage.output_tokens if hasattr(usage, "output_tokens") else 0
```

What a non-streaming `OpenAILike.chat()` response against an OpenAI-compatible endpoint actually contains:

```
response.additional_kwargs keys : ['prompt_tokens', 'completion_tokens', 'total_tokens']
response.raw.usage              : CompletionUsage(prompt_tokens=18, completion_tokens=34, total_tokens=52)
```

Two independent problems, either of which alone zeroes the count:

1. **There is no** `"usage"` **key.** The counts sit at the top level of `additional_kwargs`, so the `if` is False and the block never executes.
2. `input_tokens`**/**`output_tokens` **is Responses-API naming.** The Chat Completions API returns `prompt_tokens`/`completion_tokens`, so even if the key existed the `hasattr` checks would fail and both fields fall through to `0`.

Nothing raises, so the surrounding `try/except` never fires. It returns `TokenUsageBaseModel()` — all zeros.

**Expected:** the token usage the endpoint reported.<br>**Actual:** 0 prompt tokens, 0 completion tokens, for every LlamaIndex agent on an OpenAI-compatible endpoint.

```shell
from llama_index.core.llms import ChatMessage
from llama_index.llms.openai_like import OpenAILike
from nat.plugins.llama_index.callback_handler import LlamaIndexProfilerHandler

llm = OpenAILike(model="<model>", api_base="http://localhost:11434/v1", api_key="not-needed",
                 is_chat_model=True, is_function_calling_model=True)
r = llm.chat([ChatMessage(role="user", content="say hi")])

print(list(r.additional_kwargs.keys()))
# ['prompt_tokens', 'completion_tokens', 'total_tokens']   <- no "usage" key

print(r.raw.usage)
# CompletionUsage(prompt_tokens=18, completion_tokens=34, total_tokens=52, ...)

u = LlamaIndexProfilerHandler._extract_token_usage(r)
print(u.prompt_tokens, u.completion_tokens)
# 0 0      <- expected 18 34
```

### Relevant log output

<details><summary>Click here to see error details</summary>

No error is produced. \_extract_token_usage returns TokenUsageBaseModel() with all fields<br>zero, and the try/except never fires because nothing raises.

</details>

### Other/Misc.

**Impact.** Any LlamaIndex agent on an OpenAI-compatible endpoint reports 0 tokens. In a profiler, 0 is not a neutral default — it reads as *free* rather than *unmeasured*. In a side-by-side framework comparison this is actively misleading.

**Suggested fix.** Read `response.raw.usage` (the provider's own object) and fall back to the flattened `additional_kwargs`, accepting both `prompt_tokens`/`completion_tokens` and `input_tokens`/`output_tokens` naming.

**Note for whoever fixes this: there is a second fault stacked behind this one.** Fixing `_extract_token_usage` alone will still report 0 for a streaming agent such as `FunctionAgent`, because an OpenAI-compatible endpoint sends no usage on a streamed response unless the caller sets `stream_options: {"include_usage": true}`. Probing inside a real `FunctionAgent` run, the handler receives:

```
response_type=ChatResponse  raw_type=ChatCompletionChunk  raw.usage=None  additional_kwargs=[]
```

There is nothing to extract. That part is arguably caller-side rather than a toolkit bug, but the two faults present as one symptom and each masks the other — worth knowing before concluding a fix didn't work. Setting `stream_options` **and** fixing the key lookup gives correct counts; either alone gives 0.

**Adjacent, possibly worth separate issues.** Both are caught and logged so they only lose cosmetic fields, but they fire on every run and send you hunting in the wrong place:

* `callback_handler.py` \~line 116 — `payload.get(EventPayload.SERIALIZED)['model']` raises `KeyError: 'model'` (loses the model name on the start event).
* `callback_handler.py` \~line 181 — `llm_text_output += block.text` over `response.message.blocks` raises `AttributeError: 'ThinkingBlock' object has no attribute 'text'` for models emitting reasoning blocks (loses the recorded output text).

This may be the same class of bug as NVIDIA/NeMo-Agent-Toolkit#2116 (an evaluator reporting 0.0 because it reads a key nothing writes) — a silent zero from a key mismatch.

### Code of Conduct

- [X] I agree to follow the NeMo Agent Toolkit Code of Conduct
- [X] I have searched the [open bugs](<https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues?q=is%3Aopen+is%3Aissue+label%3Abug>) and have found no duplicates for this bug report

jstar0:
## Summary

OpenAI-compatible LlamaIndex responses can expose usage through
`response.raw.usage` or as flattened `additional_kwargs` fields. The profiler
only inspected a nested `additional_kwargs["usage"]` object and only recognized
Responses API field names, so successful calls silently reported zero tokens.

## Changes

- Read provider usage first, then fall back through nested and flattened
  LlamaIndex metadata.
- Support both Chat Completions and Responses API token field names.
- Preserve total, cached, and reasoning token counts.
- Add regression coverage for raw, flattened, and existing nested response
  shapes.

The change is limited to non-streaming extraction. It does not alter streaming
request configuration or claim to address providers that do not return usage.

## Verification

```bash
uv run --extra llama-index pytest packages/nvidia_nat_llama_index/tests
uv run --extra llama-index pre-commit run --files packages/nvidia_nat_llama_index/src/nat/plugins/llama_index/callback_handler.py packages/nvidia_nat_llama_index/tests/test_llama_index_callback_handler.py
```

Results: `41 passed, 5 skipped`; changed-file YAPF and Ruff hooks passed.

Closes #2126

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token usage reporting across multiple response formats.
  * Added support for total, cached, and reasoning token counts when available.
  * Preserved compatibility with existing nested usage data and raw provider responses.
* **Tests**
  * Added coverage for OpenAI-style, flattened, and nested token usage formats.


willkill07:
/ok to test [c7f5bb6](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2132/commits/c7f5bb61ae8b8b8747c05cc918fb79d3edbb0854)

willkill07 (approved):

PR #2140, opened 2026-07-26
Fix avg_llm_latency ATIF evaluator always reporting 0.0

### LINKED ISSUE #2116: avg_llm_latency evaluator always reports 0.0 — ATIF evaluator reads a key the ATIF converter never writes
### Version

1.8.0

_No response_

### Describe the bug.

[nat_avg_llm_latency_issue.md](https://github.com/user-attachments/files/29998188/nat_avg_llm_latency_issue.md)

```shell

```

### Relevant log output

</pre></details>

### Other/Misc.

_No response_

### Code of Conduct

AnishPatel526:
# Fix avg_llm_latency ATIF evaluator always reporting 0.0

## Reference Issues/PRs
Closes #2116.

## What does this implement/fix? Explain your changes.
`AverageLLMLatencyAtifEvaluator` computed per-step LLM latency from
`step.extra["span_event_timestamp"]` (an ISO start paired with `step.timestamp`
as the end). That key never exists on a converted ATIF step: `Step.extra` is
built through `AtifStepExtra`, which is `ConfigDict(extra="forbid")` and defines
only `ancestry` / `invocation` / tool fields. `span_event_timestamp` is a field
on the pre-conversion `IntermediateStep.payload`, not on the post-conversion
`Step.extra`. So the evaluator was reading a pre-conversion field name off a
post-conversion object and always found nothing, leaving `latencies` empty and
the score `0.0`.

The converter folds that pre-conversion field into the invocation timing:
`nat/utils/atif_converter.py::_atif_invocation_from_ist` sets
`invocation.start_timestamp = ist.payload.span_event_timestamp` (epoch seconds),
and the emitted step carries `metrics=pending.metrics`, so it clears the
evaluator's `if step.source != "agent" or not step.metrics: continue` guard.
That is the path this PR fixes.

### Fix
Read the timing the converter actually writes. A new helper `_step_llm_latency`
prefers `step.extra["invocation"]` (`end_timestamp - start_timestamp`), rejects
non-finite / overflowing values, and falls back to the legacy
`span_event_timestamp` + `step.timestamp` pair so any producer emitting that
shape (and the existing tests for it) keeps working. Steps with neither source
are skipped rather than scored as a spurious zero.

### Before / after
For a real converted ATIF agent step:
```python
step.extra = {"invocation": {"start_timestamp": 1000.0, "end_timestamp": 1003.5}}
```
- Before: `span_event_timestamp` absent (and forbidden by the schema) ->
  `latencies == []` -> `score == 0.0`.
- After: `invocation` used -> `score == 3.5`.

The legacy `span_event_timestamp` path is unchanged (still `5.0` for the
existing 5-second fixture).

### Known follow-up (out of scope here)
The ATOF-script converter `atof_to_atif_converter.py` emits `invocation` timing
but sets no `metrics`, so its steps are dropped by the evaluator's `not
step.metrics` guard and would still score `0.0` after this change, for a
separate reason. Worth a follow-up issue; this PR does not attempt to change
that producer.

## Testing
Added tests in
`packages/nvidia_nat_profiler/tests/profiler/test_runtime_evaluator_atif.py`:
- `test_evaluate_atif_item_uses_invocation_timing` uses the converter's real
  `extra["invocation"]` shape (with a conflicting `span_event_timestamp` present
  to prove `invocation` takes precedence) and asserts a non-zero average. It
  fails on `develop` (score `0.0`) and passes with this change.
- `test_evaluate_atif_item_skips_non_finite_invocation_timing` covers NaN, inf,
  and an integer too large to convert to float.

The existing `span_event_timestamp` tests continue to pass via the fallback.
`ruff check` and `ruff format` are clean on both changed files.

## Checklist
- [x] "closes #2116" is in the PR body.
- [x] New/changed code is tested (regression test that fails before the fix).
- [ ] Documentation updated where applicable (N/A, behavior fix, no API change).
- [x] Commits are signed off (DCO).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LLM latency measurements by supporting invocation timing data when available.
  * Added validation to ignore invalid, non-finite, or unusable timestamps.
  * Preserved fallback timing behavior when invocation data is unavailable.

* **Tests**
  * Added coverage for preferred invocation timings and invalid timestamp handling.


AnishPatel526:
This is a non-breaking bug fix (`avg_llm_latency` always returned 0.0 because the evaluator read `span_event_timestamp`, which the ATOF-to-ATIF converter never writes). Could a maintainer apply the `bug` and `non-breaking` labels so the Label Checker can pass? Happy to adjust if you'd categorize it differently.

AnishPatel526:
Hi! Just wanted to follow up on this contribution!

chuenchen309:
I checked this independently against `develop` (`2618705`). The diagnosis holds, and there's a stronger form of the argument plus one open question about which producer actually benefits.

**The key the evaluator reads cannot legally exist.** `AverageLLMLatencyAtifEvaluator` reads `step.extra.get("span_event_timestamp")` (`atif_evaluate.py:53`). Step extras are built through `AtifStepExtra`, which declares `model_config = ConfigDict(extra="forbid")` and defines only `ancestry`, `invocation`, `tool_ancestry`, `tool_invocations`. So `span_event_timestamp` isn't merely absent from the current producers — the typed contract for `Step.extra` forbids it. Across the repo, `extra={"span_event_timestamp": ...}` appears **only** in test fixtures (`test_runtime_evaluator_atif.py`, and one dict in `examples/dynamo_integration/.../test_tsq_formula.py`); no production path writes it.

**Both converters emit the `invocation` shape,** so the fix is broader than the PR description claims:

- `atof_to_atif_converter.py` writes `extra = {"ancestry": …, "invocation": _build_invocation_info(…)}`, with `start_timestamp` / `end_timestamp` as epoch seconds (`round(micros / 1_000_000, 3)`).
- `nat/utils/atif_converter.py` writes `extra=AtifStepExtra(…, invocation=pending.invocation, …).model_dump(exclude_none=True)`, where `_atif_invocation_from_ist` sets `start_timestamp=ist.payload.span_event_timestamp`.

That second one is the crisp statement of the bug: `span_event_timestamp` is a field on the **pre-conversion** `IntermediateStep.payload`, and the converter's job is to fold it into `invocation.start_timestamp`. The evaluator was reading the pre-conversion field name off a post-conversion object, so it always missed.

**The open question — which producer does this actually unblock?** The loop skips steps before it ever looks at timing:

```python
if step.source != "agent" or not step.metrics:
    continue
```

`nat/utils/atif_converter.py` sets `metrics=pending.metrics` on the very step whose `extra` carries `invocation`, so those steps clear the guard and this fix does change their score. But `atof_to_atif_converter.py` — the producer the PR description cites — contains no `metrics` key at all: `grep -n metrics` over that file is empty, so every step it emits is dropped by `not step.metrics` regardless of this change. Unless something downstream in the ATOF pipeline attaches metrics (I didn't find it, but I may have missed a stage), ATOF-converted trajectories would still score `0.0` after this PR, for a second and separate reason.

None of that argues against merging — the timing lookup is wrong either way and the fallback keeps the legacy shape working. It does suggest the description would be more accurate pointing at `nat/utils/atif_converter.py` as the path this actually fixes, and that ATOF-script output may need a follow-up.

One small note on the fallback: it's keyed on `isinstance(start, (int, float))`, which is right for both converters since `AtifInvocationInfo.start_timestamp` is `float | None` in epoch seconds, while the legacy fixtures use ISO strings and fall through to `_iso_to_epoch`. The two shapes differ in both key and type, so they can't collide.

*AI-assisted review. The contract, converter and evaluator code above were read on `develop` at `2618705`; I did not run the test suite locally, so the pass/fail claims in the PR description are not independently confirmed here.*

AnishPatel526:
Updated the description to point at nat/utils/atif_converter.py as the path this fixes, and added a note that the ATOF-script producer (atof_to_atif_converter.py) emits no metrics so its steps stay at 0.0, worth a separate follow-up. Thanks again for the diagnosis!

AnishPatel526:
Hi. Sorry about that I originally had it in ruff instead of yapf. Fixed it.

willkill07:
/ok to test [8d98276](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2140/commits/8d982768f8ea3662186d445bf1c640ffc29ea052)

willkill07 (approved):

PR #2036, opened 2026-06-09
Honor ChatRequest.model in LangChain react_agent workflows

marcusds:
## Summary

- `ChatRequest.model` was accepted by `/v1/chat/completions` but silently ignored. This PR wires it up for LangChain-backed react_agent (and per_user_react_agent) workflows.
- Each LangChain LLM client is wrapped with `configurable_fields` before retry/thinking patches are applied, so per-request model overrides don't shed patch behavior when LangChain's `_prepare()` re-instantiates the underlying model.
- The correct Pydantic field name is detected per backend (`model_name` → `model` → `model_id`) via introspection; Azure and local HuggingFace pipelines are excluded since they don't support runtime model switching.
- Invalid model name format and unsupported backends return a 422 with a descriptive message rather than silently no-oping or producing an opaque 500.

## Out of scope (follow-up PRs)

- `tool_calling_agent` parity — plan doc scopes this to react_agent first
- `use_native_tool_calling=True`: `bind_tools` severs the configurable pathway; override is silently ignored for native tool calling graphs
- Test coverage for the `ChatRequest.model` → `configurable["model_name"]` path
- OpenAI-spec error envelope (`{"error": {...}}`) — gap exists across the whole codebase, not specific to this change

## Test plan

- [ ] Deploy react_agent with a NIM or OpenAI backend; send `model: "some-other-model"` in request body; confirm inference uses the override
- [ ] Send `model: "bad##name"` — confirm 422 with `Invalid model name` message
- [ ] Deploy with Azure backend; send `model: "gpt-4o"` — confirm 422 with unsupported backend message
- [ ] Send request without `model` field — confirm unchanged behavior

## Summary by CodeRabbit

* **New Features**
  * Configurable per-request model selection for ReAct agent workflows.
  * Automatic detection and support for runtime model overrides in LangChain clients.
  * Model-name validation and improved invocation configuration, including adaptive recursion limits based on tool-call settings.

PR #2141, opened 2026-07-28
fix: do not forward the unknown-model sentinel as a per-request model override

### LINKED ISSUE #2110: nat run with react_agent sends the ChatRequest sentinel "unknown-model" as a per-request model override, breaking every NIM call
**Version:** develop @ e8692d00 (v1.9.0.dev, regression introduced by #2036 on 2026-07-08)

**Describe the bug**
Since #2036 ("Honor ChatRequest.model in LangChain react_agent workflows"), running any `react_agent` workflow through `nat run` (or `nat.utils.run_workflow`) fails at the first agent LLM call. The failure surfaces as:

```
[AGENT] Failed to call agent_node: [###] Unknown Error
age not found
```

which is a masked `404 page not found` from `https://integrate.api.nvidia.com/v1/chat/completions`.

**Root cause**
- String inputs are converted with `ChatRequest.from_string(data, model="unknown-model")` (`packages/nvidia_nat_core/src/nat/data_models/api_server.py:936`; the `ChatRequest.model` field also defaults to `"unknown-model"` at line 375).
- `react_agent`'s `_build_lc_config` (`packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/react_agent/register.py:47`) treats any non-`None` `message.model` as a deliberate per-request override and sets `configurable: {"model_name": message.model}`.
- The `RunnableConfigurableFields` wrapper added in `_patch_llm_based_on_config` then reconstructs the client with `model="unknown-model"`, so the request body carries `"model": "unknown-model"` and the NIM gateway returns `404 page not found` (it routes by model name).

The sentinel default therefore silently clobbers the `model_name` configured in the workflow YAML on every CLI run. CI does not catch this because the Test stage never passes `--run_integration`, so no live agent/LLM call is exercised.

**To Reproduce**
1. `export NVIDIA_API_KEY=<valid key>`
2. Run any react_agent example, e.g.:
   `nat run --config_file examples/getting_started/simple_calculator/configs/config.yml --input "What is 2 * 4?"`
3. Observe `[AGENT] Failed to call agent_node: [###] Unknown Error` (masked 404). Instrumenting aiohttp shows the request body contains `"model": "unknown-model"` despite the config specifying a valid model.

**Expected behavior**
`nat run` uses the model configured in the workflow YAML. The `"unknown-model"` sentinel should not be treated as a user-requested override.

**Suggested fix**
Ignore the sentinel in `_build_lc_config` (or don't apply an override when the request model equals the `ChatRequest` default):

```python
if model == "unknown-model":
    model = None
```

Verified locally: with this one-line guard, react_agent workflows run correctly end-to-end again, and genuine per-request model overrides continue to work. (The same `message.model` forwarding exists in `tool_calling_agent`; it likely needs the same guard.)

knQzx:
## Description

Closes #2110

when a model can't be resolved, the ReAct agent builder receives the sentinel `model="unknown-model"` and forwards it straight through as a per-request `configurable.model_name` override, so every downstream LLM/NIM call is told to use a model literally named `unknown-model` and the provider returns 404. this is a regression from #2036, which introduced the sentinel but didn't guard against forwarding it

the fix treats the `unknown-model` sentinel the same as an unset model - the per-request model override is only applied when a real model name is present

the guard sits inside `_build_lc_config` rather than at its call sites, so it also covers `register_per_user_agent.py`, which imports the same helper and passes `message.model` the same way - `per_user_react_agent` has the identical defect on develop

the sentinel value itself now lives in `nat/data_models/api_server.py` as `UNKNOWN_MODEL_SENTINEL` and every reference points at it (the five defaults/fallbacks there, the `_string_to_nat_chat_request` converter, this guard, and the literal in `tool_calling_agent/register.py`), so the value can't drift apart between packages

one behaviour note: with the sentinel short-circuiting the block it no longer reaches the `supports_override` check, so a backend without per-request override support accepts plain string input instead of raising. that is intended - the user never asked for a model - and the test pins it

reproduce: build a ReAct agent whose model resolves to the `unknown-model` sentinel (an unresolved model reference), then run it. on develop the request carries `configurable: {model_name: "unknown-model"}` and the call 404s; with this change no model override is attached and the configured default is used. the added `test_build_lc_config_ignores_unknown_model_sentinel` asserts the sentinel is not forwarded

- I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/docs/source/resources/contributing/index.md)
- We require that all contributors "sign-off" on their commits - this commit is signed off
- New tests cover these changes
- The documentation is up to date with these changes

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented unspecified or unknown model values from being incorrectly applied as per-request model overrides.
  - Preserved validation and support checks when a valid model override is provided.
  - Standardized handling of unknown model values across chat responses and streaming responses.
- **Tests**
  - Added regression coverage for missing, unknown, and valid model overrides.


chuenchen309:
I verified this independently against `develop` (`2618705`). The bug is real and the fix is in the right place; two notes below, one of which I think is worth acting on.

**The chain, link by link.** `_string_to_nat_chat_request` is registered on `GlobalTypeConverter` and hardcodes the sentinel:

```python
def _string_to_nat_chat_request(data: str) -> ChatRequest:
    return ChatRequest.from_string(data, model="unknown-model")   # api_server.py:954-955
```

`_response_fn` converts through exactly that path (`register.py:169`) and hands `message.model` to `_build_lc_config` (`register.py:184`). The remaining question was whether the sentinel survives validation, and it does — `_MODEL_NAME_RE` accepts it:

```
>>> _MODEL_NAME_RE.match("unknown-model")
<re.Match object; span=(0, 13), match='unknown-model'>
```

So on `develop`, any plain-string input to a ReAct agent attaches `configurable: {"model_name": "unknown-model"}`. Confirmed.

**Fixing inside `_build_lc_config` rather than at the call sites matters more than it looks.** There is a third caller outside this file: `register_per_user_agent.py:89` imports `_build_lc_config` and passes `message.model` the same way. That import and call arrived in #2036 alongside the sentinel, so `per_user_react_agent` has the identical defect. A guard placed at the two call sites in `register.py` would have left it broken; this patch covers all three. Might be worth a line in the description, since the diff touching only `react_agent/register.py` makes it look narrower than it is.

**The one thing I'd change: the sentinel now has two independent definitions.** `_UNKNOWN_MODEL_SENTINEL` is declared in `react_agent/register.py`, but the value originates in `api_server.py`, where it appears as a bare literal in six places (lines 375, 401, 430, 455, 489, 955). If any of those is ever changed, this guard keeps comparing against the old string and silently stops matching — the failure mode is the original 404, with no test catching it, because the test derives its input from the converter and would follow the new value.

This repo already has the pattern for the fix: `STDOUT_ENDPOINT_SENTINEL` in `nat/utils/telemetry/config.py:82`, referenced from `handler.py:230`. Exporting something like `UNKNOWN_MODEL_SENTINEL` from `api_server.py` and importing it here would make the coupling explicit and let the six literals collapse onto it later.

**Minor, and I think intentional:** with the sentinel short-circuiting the whole block, it no longer reaches the `supports_override` check either. That is the correct behaviour — the user did not request a model, so raising "the configured inference backend does not support per-request model selection" at them would be wrong — but it does mean a backend without override support now silently accepts string input that previously raised. Worth stating explicitly rather than leaving as a side effect of the guard's placement.

*AI-assisted review. The code paths above were read on `develop` at `2618705` and the regex match was executed; I did not build the package or run the test suite, so the pass/fail claims in the PR description are not independently confirmed here.*

willkill07:
/ok to test [7f41f9c](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2141/commits/7f41f9c8a92bff1b1832d1965ff2f630d779e198)

packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/react_agent/register.py
willkill07, line 1: You need to update other references with the source of truth.

I recommend holding the constant in `packages/nvidia_nat_core/src/nat/data_models/api_server.py` rather than in the langchain package.

Reference:
https://github.com/search?q=repo%3ANVIDIA%2FNeMo-Agent-Toolkit+%22unknown-model%22&type=code

willkill07 (changes requested):

packages/nvidia_nat_langchain/src/nat/plugins/langchain/agent/react_agent/register.py
knQzx: done - the constant now lives in `nat/data_models/api_server.py` as `UNKNOWN_MODEL_SENTINEL` and every reference points at it: the five defaults/fallbacks in that file, the `_string_to_nat_chat_request` converter, this guard, and the hardcoded literal in `tool_calling_agent/register.py`. no bare `"unknown-model"` left outside the definition

also dropped the misleading comment - `ChatRequest.model` defaults to `None`, only the plain-string converter injects the sentinel

the test now asserts the converter output equals the constant, and covers `supports_override=False` so the sentinel is ignored before the backend capability check and can never raise the unsupported-override error

willkill07 (approved):

PR #2147, opened 2026-07-28
feat(core): expose read-only views of function group filters

DABH:
## Description

`FunctionGroup` is part of the public plugin-authoring contract: third-party plugin packages import it from the stable `nat.plugin_api` facade (per `docs/source/extend/third-party-plugins.md`, symbols exported from `nat.plugin_api` are the public contract for external plugin packages) and build provider tool groups through `register_function_group`. A group's composition can be shaped by two kinds of filter callbacks: a group-level `filter_fn` (constructor argument, or `set_filter_fn`) and per-function filter callbacks (`add_function(..., filter_fn=...)`, or `set_per_function_filter_fn`). Today both are write-only from the outside: a plugin package that wants to validate, document, or instrument how a group is composed (for example, asserting in its own tests that the expected filters were wired up, or logging which functions are dynamically gated) has to reach into the private `_filter_fn` and `_per_function_filter_fn` attributes, which is exactly the kind of implementation-module reliance the third-party plugin guide tells authors to avoid.

This change adds two minimal read-only properties to `FunctionGroup`, mirroring the existing `instance_name` and `middleware` properties:

- `filter_fn` returns the configured group-level filter callback, or `None` when no group-level filter has been set.
- `per_function_filter_fns` returns an immutable `types.MappingProxyType` view of the per-function filter callbacks keyed by function name. The view rejects mutation with `TypeError` and stays live: filter callbacks registered later (through `add_function` or `set_per_function_filter_fn`) appear in a previously obtained view.

The change is purely additive: no existing behavior changes, and no new module-level symbols are added to `nat.plugin_api` (the properties ride along on the already-exported `FunctionGroup` class, so the pinned export test and the plugin API surface documentation are unaffected).

## Testing

- `uv run pytest packages/nvidia_nat_core/tests/nat/builder/test_function_group.py` — 28 passed (includes two new tests covering groups constructed with and without filters, filters set after construction, read-only enforcement on the mapping view, and the view reflecting later additions).
- `uv run pytest packages/nvidia_nat_core/tests/nat/builder` — 319 passed.
- `uv run pre-commit run yapf --files packages/nvidia_nat_core/src/nat/builder/function.py packages/nvidia_nat_core/tests/nat/builder/test_function_group.py` — passed.
- `uv run pre-commit run ruff-check --files packages/nvidia_nat_core/src/nat/builder/function.py packages/nvidia_nat_core/tests/nat/builder/test_function_group.py` — passed.
- `uv run python ci/scripts/copyright.py --verify-apache-v2` — passed.

## Summary by CodeRabbit

* **New Features**
  * Function groups now expose configured filtering behavior.
  * Added access to per-function filters through a read-only mapping that reflects updates automatically.

* **Bug Fixes**
  * Improved visibility and consistency of function filtering configuration while preventing accidental changes through the exposed mapping.


willkill07:
> a plugin package that wants to validate, document, or instrument how a group is composed (for example, asserting in its own tests that the expected filters were wired up, or logging which functions are dynamically gated) has to reach into the private _filter_fn and _per_function_filter_fn attributes, which is exactly the kind of implementation-module reliance the third-party plugin guide tells authors to avoid.

I get your point, but at the same time, we intentionally don't want to expose `filter_fn` or `per_function_filter_fn`, even as a read-only property. Since it's not necessary for the runtime behavior and is only for **testing**, I don't see why manual inspection is inappropriate.

DABH:
Fair - agreed the filter callables themselves shouldn't be public, and manual inspection works for tests.

The narrower need I was reaching for is that at runtime, a plugin that validates configuration at startup wants to know whether a group's membership is dynamic (filtered), because dynamic membership changes what can be verified up front. It never needs the filters themselves though. Would a minimal `dynamic_membership: bool` read-only property (no callables exposed) be acceptable instead? If that still feels like surface you'd rather not carry, happy to close this and rely on manual inspection per your comment.  Thanks either way!

willkill07:
@DABH `FunctionGroup.get_config` has `include` and `exclude` members which can be used to tell if things are dynamically filtered. In short: if `include` is specified, then it's explicit.

In the general case, custom filter functions can be passed when retrieving functions, so it's not quite a simple yes/no case. Rather, it is dependent on explicit API usage.

Does this help?

willkill07 (changes requested):
I believe the existing public API offers equivalent semantics.

DABH:
It does help, thanks. Verified the semantics on a current nightly too: an explicit `include` pins the membership universe (unknown names fail at build via `get_included_functions`), `exclude` is validated static subtraction, and retrieval-time filter args mean a group-level flag couldn't be authoritative anyway — as you said, it depends on explicit API usage. That dissolves the need for new surface: configuration-time checks can read `get_config()` for the static case, and anything deeper stays test-territory manual inspection per your earlier comment. Closing — thanks for talking it through.

PR #2157, opened 2026-08-12
fix(langchain): set a User-Agent for the wiki_search tool

### LINKED ISSUE #2156: [BUG] wiki_search tool fails on every run — Wikipedia API rejects requests without a User-Agent
The built-in `wiki_search` tool (`nat.plugins.langchain.tools.wikipedia_search`) fails on every invocation, including the top-level README's Hello World example and `examples/notebooks/hello_world.ipynb`.

**Repro:** follow the README Hello World verbatim, run:
```
nat run --config_file workflow.yml --input "List five subspecies of Aardvarks"
```

**Symptom:** the tool call fails with:
```
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```
The agent silently falls back to its own parametric knowledge, so the workflow appears to "succeed" while the tool is actually non-functional.

**Root cause:**
```
curl 'https://en.wikipedia.org/w/api.php?action=query&list=search&srsearch=test&format=json'
-> 403, text/plain: "Please set a user-agent and respect our robot policy."
```
Wikipedia now enforces a User-Agent requirement on API requests. The `wikipedia` PyPI package used internally by `WikipediaLoader` (via `langchain_community`) doesn't set one, so every request 403s and the response body isn't valid JSON.

I have a fix ready — sets a descriptive User-Agent via `wikipedia.set_user_agent()` once at tool construction, matching the Wikimedia Foundation's User-Agent policy — plus unit tests covering it. Will open a PR against this issue.

min9pm8:
## Description

Closes #2156

Wikipedia's API now rejects unauthenticated requests with a 403 and a plain-text body ("Please set a user-agent and respect our robot policy"), which the `wikipedia` package's JSON parsing surfaces as a generic `JSONDecodeError`. This breaks the `wiki_search` tool on every call, including the top-level README's Hello World example and `examples/notebooks/hello_world.ipynb` — the agent silently falls back to its own parametric knowledge instead of retrieving Wikipedia content, so the failure is easy to miss.

This sets a descriptive User-Agent via `wikipedia.set_user_agent()` once when the tool is constructed, matching the [Wikimedia Foundation's User-Agent policy](https://foundation.wikimedia.org/wiki/Policy:User-Agent_policy).

## Testing

- Added `packages/nvidia_nat_langchain/tests/test_wikipedia_search.py`: verifies the tool sets a User-Agent on construction, and verifies search-result formatting against a mocked `WikipediaLoader`.
- Ran the full `nvidia_nat_langchain` test suite: 620 passed, 2 new tests passing, 13 skipped, 69 pre-existing failures — all `ModuleNotFoundError` for optional provider extras (`langchain_exa`, OpenAI, OCI, Bedrock, Azure, LiteLLM) not installed in this scoped dev environment, unrelated to this change.
- `ruff check` and `yapf --diff` both clean on the changed files.
- Manually reproduced the original failure and confirmed the fix resolves it end-to-end by running the Hello World workflow against a real NIM endpoint before and after the patch.

- New tests cover these changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Wikipedia search reliability by configuring the required User-Agent before searches are performed.
  * Search results now consistently include source URLs and page content when loaded asynchronously.

* **Tests**
  * Added coverage for User-Agent configuration, loader behavior, and formatted Wikipedia search results.

### DIFF
--- a/packages/nvidia_nat_langchain/src/nat/plugins/langchain/tools/wikipedia_search.py
+++ b/packages/nvidia_nat_langchain/src/nat/plugins/langchain/tools/wikipedia_search.py
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from importlib.metadata import version
+
 from nat.builder.builder import Builder
 from nat.builder.framework_enum import LLMFrameworkEnum
 from nat.builder.function_info import FunctionInfo
@@ -37,7 +39,8 @@ async def wiki_search(tool_config: WikiSearchToolConfig, builder: Builder):
     # Wikipedia's API rejects requests without a User-Agent header (returns a non-JSON 403),
     # and the `wikipedia` package (used internally by WikipediaLoader) doesn't set one by
     # default. See https://foundation.wikimedia.org/wiki/Policy:User-Agent_policy
-    wikipedia.set_user_agent("NeMoAgentToolkit/1.0 (https://github.com/NVIDIA/NeMo-Agent-Toolkit)")
+    wikipedia.set_user_agent(
+        f"NeMoAgentToolkit/{version('nvidia-nat-langchain')} (https://github.com/NVIDIA/NeMo-Agent-Toolkit)")
 
     async def _wiki_search(question: str) -> str:
         # Search the web and get the requested amount of results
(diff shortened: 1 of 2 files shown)


min9pm8:
Addressed all three points from the CodeRabbit review in 24de94f: User-Agent version now comes from `importlib.metadata.version()` instead of a hardcoded string, added type annotations on the test fixture, renamed it to match this repo's `*_fixture` convention, and strengthened both assertions to check actual content instead of truthy/substring checks.

willkill07 (approved):

PR #2160, opened 2026-08-17
fix(core): prevent duplicate middleware registrations in function chains

### LINKED ISSUE #2093: Middleware should not be registered twice for the same function
## Description

Dynamic function middleware can register itself against a function automatically, while explicit function middleware configuration can also register that same middleware against the same function.

When both paths are used for the same middleware/function pair, NAT can build a middleware chain where the same middleware instance appears twice for a single function.

## Observed behavior

```text
Building middleware for function 'example_function' in order of:
(<ExampleMiddleware object at ...>, <ExampleMiddleware object at ...>)
```

This causes the same middleware to execute twice for one function invocation.

## Problem

NAT should not allow the same middleware to be registered more than once for the same function boundary. Even if there are multiple valid ways to attach middleware, the final middleware chain for a function should not contain duplicate registrations of the same middleware instance.

## Why this matters

Duplicate registration can cause repeated pre/post invocation handling, duplicate logs, unnecessary latency, repeated input/output mutation, and incorrect behavior for middleware that is not idempotent.

## Suggested direction

NAT should detect this duplicate registration case and handle it explicitly. The exact behavior could be either a validation error or a safe deduplication step, but duplicate middleware entries should not silently make it into the function middleware chain.

## Additional context

This came up when using middleware that can be attached dynamically to workflow functions while also being listed explicitly on the function. The resulting middleware chain showed the same middleware instance twice for the same function.

alwaysprince05:
## Problem

A middleware instance could appear more than once in a function's middleware chain when it was both explicitly configured for the function and later auto-registered by a DynamicFunctionMiddleware. The middleware then executed multiple times per invocation — causing repeated pre/post handling, duplicate logs, extra latency, and incorrect behavior for non-idempotent middleware.

**Reproduced on develop @ ad7f4a47:**

```text
after explicit config    : [<DynamicMiddleware object at 0x...>]
after runtime discovery  : [<DynamicMiddleware object at 0x...>, <DynamicMiddleware object at 0x...>]  # same instance twice
middleware executions for one call: 2
```

## Root cause

All middleware configuration paths (explicit per-function config, function-group config, and dynamic auto-registration) flow through `validate_middleware` in `packages/nvidia_nat_core/src/nat/middleware/function_middleware.py`, which validated types and final-middleware ordering but never checked for duplicate instances. `DynamicFunctionMiddleware._register_function` appends itself to the function's existing chain without a membership check, so a middleware already present in the chain via explicit configuration was added a second time.

## Solution

Deduplicate middleware instances by identity in `validate_middleware` (the single choke point for every configuration path). The first occurrence keeps its position, so an explicitly configured middleware wins over a later auto-registered duplicate. Repeated registrations log a warning instead of silently duplicating chain entries. The final-middleware ordering guarantees are now evaluated on the deduplicated list. No public API changes.

## Tests

- `test_register_function_does_not_duplicate_middleware_already_in_chain` — regression test reproducing the exact collision (explicit config + dynamic registration); asserts the chain contains the instance once and it executes exactly once per invocation. **Failed before the fix, passes after.**
- `test_validate_middleware_deduplicates_repeated_instances`
- `test_validate_middleware_keeps_distinct_instances_of_same_class`
- `test_validate_middleware_deduplicated_final_is_accepted`

## Validation

- `pytest packages/nvidia_nat_core` → **2842 passed, 54 skipped, 0 failed** (Python 3.13)
- pre-commit (yapf, ruff, uv-lock, etc.) → all pass on changed files
- `ci/scripts/copyright.py --verify-apache-v2`, `documentation_checks.sh` (nbconvert + vale: 0 errors), `path_checks.sh` → all pass

## Limitations

- Integration tests not run (change is not covered by them); CI will cover Python 3.11–3.13.

Closes #2093

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate middleware from running more than once during a function invocation.
  - Preserved the order of configured middleware while removing repeated references.
  - Correctly handles duplicated final middleware after deduplication.
  - Keeps separate middleware instances independent, even when they share the same type.
  - Added logging when duplicate middleware entries are ignored.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/middleware/function_middleware.py
+++ b/packages/nvidia_nat_core/src/nat/middleware/function_middleware.py
@@ -30,6 +30,7 @@ the next callable.
 
 from __future__ import annotations
 
+import logging
 from collections.abc import AsyncIterator
 from collections.abc import Sequence
 from typing import Any
@@ -40,6 +41,8 @@ from nat.middleware.middleware import FunctionMiddlewareContext
 from nat.middleware.middleware import InvocationAction
 from nat.middleware.middleware import InvocationContext
 from nat.middleware.middleware import Middleware
+
+logger = logging.getLogger(__name__)
 
 
 class FunctionMiddleware(Middleware):
@@ -362,7 +365,12 @@ class FunctionMiddlewareChain:
 
 
 def validate_middleware(middleware: Sequence[Middleware] | None) -> tuple[Middleware, ...]:
-    """Validate a sequence of middleware, enforcing ordering guarantees."""
+    """Validate a sequence of middleware, enforcing ordering guarantees.
+
+    Repeated instances of the same middleware are dropped (the first occurrence keeps its
+    position so that an explicitly configured middleware wins over a later auto-registered
+    one), since the same instance must not execute more than once per invocation.
+    """
 
     if not middleware:
         return tuple()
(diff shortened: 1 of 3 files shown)


alwaysprince05:
Hi — thank you for the quick review feedback; both nits are addressed in ec14a469 (type hints + docstring code identifiers). The branch is also synced with develop.

Could a vetter/maintainer add the `bug` and `non-breaking` labels (I don't have permission to label this repo, and the required Label Checker needs them) so the pipeline can run? Thanks!

willkill07 (approved):

willkill07:
/ok to test [ec14a46](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2160/commits/ec14a46991f49ed696d931cdf67249cdd372a82e)

Issue #1755, opened 2026-03-06
Display simple agent thought outputs as alternative to intermediate steps


thepatrickchin:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### How would you describe the priority of this feature request

Medium

### Please provide a clear description of problem this feature solves

Non-technical users find the Intermediate Steps displayed in agent responses to be too complicated and business users don't want to see the "technical mumbo-jumbo". However if we disable the intermediate steps, there is no sense of progression aside from the tiny spinner and "Stop Generating" button. Some tools in our workflow take a minute to run so it feels like the agent is not responding without the intermediate steps being displayed.

 

### Describe your ideal solution

I propose adding a simple thought output as a user-friendly alternative to the intermediate steps UI. It would display LLM thoughts, show tool/function calls, and allow agent developers to emit custom process updates.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct
- [x] I have searched the [open feature requests](https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22%2Cimprovement%2Cenhancement) and have found no duplicates for this feature request

AnuradhaKaruppiah:
@ericevans-nv pls review

GraceJiang0312:
Hi, I'd like to work on this. I plan to split it into a few smaller PRs to keep review scope tight:

1. Static thought_description support — let function/tool configs declare a custom label that's shown instead of raw intermediate steps (no LLM output parsing involved)
2. Auto-extracted LLM thought text for ReAct agents
3. Streaming custom thought emission from within function code

Starting with 1. Will open a draft PR once I have something to show.

GraceJiang0312:
Opened #2163. This is the first PR from the breakdown above (static `thought_description` support). It's out of draft and ready for review. It's currently blocked on code owner review from `@nvidia/nat-developers`.

@thepatrickchin tagging you since this is your issue. I also saw your earlier attempt at #1756 and used it as a reference for the overall shape. Thanks for that groundwork. @willkill07 tagging you as well since you've been actively reviewing in this area. I would appreciate a look when either of you has a chance.

PR #2163, opened 2026-08-18
feat(core): add thought_description for simplified progress display

GraceJiang0312:
## Description

Part of a smaller-PR breakdown of #1755 (see [comment](https://github.com/NVIDIA/NeMo-Agent-Toolkit/issues/1755#issuecomment-5330187956)). This PR adds an optional `thought_description` config field on `FunctionBaseConfig`. When set, the step adaptor surfaces it as `thought_text` on `FUNCTION_START`/`FUNCTION_END` steps instead of the default "Running function: `<name>`" text — giving a friendlier progress label without needing to parse LLM output.

Deliberately out of scope for this PR (planned as follow-ups):
- Automatic thought extraction from ReAct-style LLM output
- Tool-call thought labels (`TOOL_START`/`TOOL_END` events are emitted by each framework's own callback handler, not `nvidia_nat_core`, so that needs per-framework wiring)
- Streaming custom thought emission from within function code

Opened as draft — happy to adjust based on feedback before marking ready for review.

## Summary by CodeRabbit

* **New Features**
  * Added optional thought descriptions for functions.
  * Intermediate execution steps now display configured thought text with running and completed statuses.
  * Thought descriptions are supported for both standard and streaming execution.
  * Thought text is omitted for top-level workflows and functions without configured descriptions.
* **Bug Fixes**
  * Improved propagation and fallback handling for thought descriptions across execution updates.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/data_models/api_server.py
+++ b/packages/nvidia_nat_core/src/nat/data_models/api_server.py
@@ -524,6 +524,7 @@ class ResponseIntermediateStep(ResponseBaseModelIntermediate):
     type: str = "markdown"
     name: str
     payload: str
+    thought_text: str | None = None
 
 
 class ResponseObservabilityTrace(BaseModel, ResponseSerializable):
--- a/packages/nvidia_nat_core/src/nat/builder/function.py
+++ b/packages/nvidia_nat_core/src/nat/builder/function.py
@@ -76,6 +76,7 @@ class Function(FunctionBase[InputT, StreamingOutputT, SingleOutputT], ABC):
         self.description = description
         self.instance_name = instance_name or config.type
         self.display_name = config.name or self.instance_name
+        self.thought_description = config.thought_description
         self._context = Context.get()
         self._configured_middleware: tuple[Middleware, ...] = tuple()
         self._middlewared_single: _InvokeFnT | None = None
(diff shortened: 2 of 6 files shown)


packages/nvidia_nat_core/src/nat/data_models/api_server.py
```diff
@@ -524,6 +524,7 @@ class ResponseIntermediateStep(ResponseBaseModelIntermediate):
     type: str = "markdown"
     name: str
     payload: str
+    thought_text: str | None = None
```
willkill07, line 527: Why is this called `thought_text` and not consistent with `thought_description` ?

GraceJiang0312: `thought_description` is the raw value configured on `FunctionBaseConfig`, essentially the "input". `thought_text` on `ResponseIntermediateStep` is the rendered string actually shown in the UI, it's derived from `thought_description` via `StepAdaptor._get_thought_description()`: falls back to a default like "Running function: <name>" when `thought_description` isn't set, and always has a status suffix appended ("..." while running, "... completed" once finished). So it's never a direct copy of the config value, it seemed clearer to give it a distinct name (`thought_text`) rather than reusing `thought_description` for a value that's been transformed.

willkill07: Right, but this now increases cognitive load. It's still a description. And now folks have to manually tie `description` -> `text`.

Reduced orthogonality is a good thing.

willkill07:
/ok to test [9b85466](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2163/commits/9b85466abd644e366c253d2264a87a066e25e636)

packages/nvidia_nat_core/src/nat/data_models/api_server.py
willkill07: @GraceJiang0312 to be clear, I'm waiting for an update that addresses this. I do not believe we should have two different names where one is *effectively* the same as the other. Both are `thought_descriptions`. There is no reason to introduce another name.

alwaysprince05:
Good point on the naming consistency, @willkill07.

I agree that having both `thought_description` and `thought_text` increases cognitive load unnecessarily. 

Suggestion: Rename `thought_text` to `thought_description` throughout for consistency. The rendered value with status suffixes is an implementation detail that doesn't need a separate field name.

LGTM on the feature itself — the thought_description metadata propagation looks clean.

PR #2165, opened 2026-08-19
fix(auth): isolate OAuth2 client per flow in console/MCP auth handler

### LINKED ISSUE #2164: OAuth2 console/MCP auth flow handler shares one OAuth client across concurrent flows, corrupting token exchange
### Version

1.8.0 (develop, HEAD at c933737c)

Source

### Describe the bug.

`ConsoleAuthenticationFlowHandler` (and `MCPAuthenticationFlowHandler`, which subclasses it) supports multiple concurrent OAuth2 Authorization-Code flows on a single handler instance — `self._flows` is a `dict[str, _FlowState]` keyed by the OAuth `state` parameter, and `self._active_flows` is a counter that keeps the shared redirect server alive until it drops to zero.

However, the OAuth2 client used to redeem an authorization code is **not** part of that per-flow state. `construct_oauth_client()` stores every newly built client into a single shared instance attribute, `self._oauth_client`, overwriting whatever was there before:

https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/packages/nvidia_nat_core/src/nat/front_ends/console/authentication_flow_handler.py#L91-L106

The `/auth/redirect` callback built once in `_build_redirect_app()` always exchanges the code using `self._oauth_client` — i.e. whichever client happens to be the most recently constructed one — rather than the client that generated the specific flow's authorization URL:

https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/packages/nvidia_nat_core/src/nat/front_ends/console/authentication_flow_handler.py#L220-L252

If two OAuth2 flows are in flight at the same time on the same handler instance (e.g. a workflow with two tools backed by two different OAuth2 providers that both need a fresh login), the second flow's `construct_oauth_client()` call silently overwrites `self._oauth_client` before the first flow's redirect has been completed by the user. When the first flow's redirect lands, the handler correctly looks up that flow's `_FlowState` (correct `token_url`, PKCE verifier, etc.) but redeems the code using the **second** flow's client — wrong `client_id`/`client_secret`/`token_endpoint_auth_method`. The token exchange then fails (`invalid_grant`/`invalid_client`) and the first flow's `authenticate()` call raises or crashes.

Nothing serializes concurrent flows against each other: `construct_oauth_client()` is called before the `self._server_lock` is even acquired, and that lock is released well before the (up to 5 minute) wait on the user completing the browser login, so the window during which a second flow can corrupt the first is effectively the whole login duration.

The sibling implementations used by the FastAPI HTTP/WebSocket front ends already avoid this by storing the OAuth client per-flow (`flow_state.client`) instead of on a shared instance attribute:
- https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/packages/nvidia_nat_core/src/nat/front_ends/fastapi/auth_flow_handlers/http_flow_handler.py#L175
- https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/packages/nvidia_nat_core/src/nat/front_ends/fastapi/auth_flow_handlers/websocket_flow_handler.py#L133

`ConsoleAuthenticationFlowHandler` (and by inheritance `MCPAuthenticationFlowHandler`) never received the equivalent fix.

```python
handler = ConsoleAuthenticationFlowHandler()

# Flow A: authenticate against provider A
task_a = asyncio.create_task(handler.authenticate(cfg_provider_a, AuthFlowType.OAUTH2_AUTHORIZATION_CODE))

# Before the user finishes logging in for A, flow B starts against a different provider
task_b = asyncio.create_task(handler.authenticate(cfg_provider_b, AuthFlowType.OAUTH2_AUTHORIZATION_CODE))

# construct_oauth_client(cfg_provider_b) overwrites handler._oauth_client before A's
# browser redirect lands, so A's token exchange is attempted with provider B's
# client_id/client_secret/token endpoint.
```

### Relevant log output

Authentication failed: Authorization server rejected the request. You may close this tab.
# or, depending on the provider, a KeyError/RuntimeError when parsing the
# malformed token response returned for the mismatched client.

</pre></details>

### Other/Misc.

I have a fix and a regression test ready and will open a PR shortly: adding a `client` field to the handler's per-flow `_FlowState`, populating it in `_handle_oauth2_auth_code_flow`, and using `flow_state.client` instead of `self._oauth_client` in the redirect callback — mirroring the pattern already used by `http_flow_handler.py` / `websocket_flow_handler.py`.

pranaysb:
## Description

Closes #2164

`ConsoleAuthenticationFlowHandler` supports multiple concurrent OAuth2 authorization-code flows on a single instance -- flows are tracked in `self._flows: dict[str, _FlowState]` keyed by the OAuth `state`, and `self._active_flows` keeps the shared redirect server alive until it drops to zero. However, the OAuth2 client used to redeem an authorization code was kept in a single shared `self._oauth_client` attribute rather than being part of that per-flow state.

`construct_oauth_client()` overwrites `self._oauth_client` on every call, and the `/auth/redirect` callback always reads the *current* `self._oauth_client` instead of the client that was built for the flow being completed. If a second flow starts (e.g. a different OAuth2 provider) before an earlier flow's browser login finishes, the second flow's client silently overwrites the first's, and the first flow's token exchange is attempted with the wrong `client_id`/`client_secret`/`token_endpoint_auth_method` -- causing it to fail or, if the mismatch happens to pass the authorization server's checks, exchange the code under the wrong client. Nothing serializes this: `construct_oauth_client()` runs before `self._server_lock` is even acquired, and the lock is released well before the (up to 5 minute) wait on the browser redirect, so the corruption window spans the whole login duration.

This mirrors a bug class already fixed for the FastAPI HTTP/WebSocket auth flow handlers, which store the client per-flow (`flow_state.client`) instead of on a shared attribute. This PR applies the same fix to the console/MCP handler:
- Add a `client` field to `_FlowState`.
- Populate it in `_handle_oauth2_auth_code_flow` right after `construct_oauth_client()`.
- Use `flow_state.client` in the redirect callback instead of `self._oauth_client`.

`MCPAuthenticationFlowHandler` inherits the fix automatically since it reuses the base class's `_handle_oauth2_auth_code_flow` and `_build_redirect_app`.

Added `test_console_oauth2_concurrent_flows_use_isolated_clients`, which drives two OAuth2 flows concurrently on one handler instance against two different registered clients and asserts each flow's token exchange used its own client. Verified this test fails against the pre-fix code (the corrupted flow's token exchange returns `{"detail": "invalid_grant"}` and the handler raises trying to build the `Bearer` header) and passes after the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OAuth sign-in reliability when multiple authorization flows run at the same time.
  * Ensured each sign-in flow uses its own authorization session, preventing token mix-ups.

* **Tests**
  * Added coverage for concurrent OAuth sign-in flows to verify that each receives the correct token.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/front_ends/console/authentication_flow_handler.py
+++ b/packages/nvidia_nat_core/src/nat/front_ends/console/authentication_flow_handler.py
@@ -48,6 +48,7 @@ class _FlowState:
     verifier: str | None = None
     token_url: str | None = None
     use_pkce: bool | None = None
+    client: AsyncOAuth2Client | None = None
 
 
 # --------------------------------------------------------------------------- #
@@ -165,6 +166,7 @@ class ConsoleAuthenticationFlowHandler(FlowHandlerBase):
 
         flow_state.token_url = cfg.token_url
         flow_state.use_pkce = cfg.use_pkce
+        flow_state.client = client
 
         # PKCE bits
         if cfg.use_pkce:
@@ -232,7 +234,7 @@ class ConsoleAuthenticationFlowHandler(FlowHandlerBase):
                 return "Invalid state; restart authentication."
             flow_state = self._flows[state]
             try:
-                token = await self._oauth_client.fetch_token(  # type: ignore[arg-type]
+                token = await flow_state.client.fetch_token(  # type: ignore[union-attr]
                     url=flow_state.token_url,
                     authorization_response=str(request.url),
                     code_verifier=flow_state.verifier if flow_state.use_pkce else None,
--- a/packages/nvidia_nat_core/tests/nat/front_ends/auth_flow_handlers/test_console_flow_handler.py
+++ b/packages/nvidia_nat_core/tests/nat/front_ends/auth_flow_handlers/test_console_flow_handler.py
@@ -21,6 +21,7 @@ from unittest.mock import MagicMock
 
 import httpx
 import pytest
+from authlib.integrations.httpx_client import AsyncOAuth2Client
 from httpx import ASGITransport
 from mock_oauth2_server import MockOAuth2Server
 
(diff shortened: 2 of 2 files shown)


willkill07 (approved):

willkill07:
/ok to test [4a96ba6](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2165/commits/4a96ba6b8ef08e884bb2a9ee23670f2d0e8b0c05)

PR #2166, opened 2026-08-20
fix(adk): resolve PEP 563 string annotations in tool wrapper

### LINKED ISSUE #2161: ADK tool wrapper raises KeyError when the tool's module uses `from __future__ import annotations
### Version

1.8.0 (also reproduced on 1.5.0)

PyPi

### Describe the bug.

`google_adk_tool_wrapper` builds the wrapped callable's `__signature__` from
`input_schema.__annotations__`. For any input schema whose class was declared in a module
using `from __future__ import annotations` (PEP 563), those annotations are **strings**
(`'str'`) rather than type objects.

Google ADK then treats a string annotation as a *deferred* annotation and tries to resolve it
against the wrapped function's own module namespace. The wrapped callable is defined inside
`nat/plugins/adk/tool_wrapper.py`, where the parameter name is not defined, so it raises:
KeyError: '<parameter name>'

This happens in `google/adk/tools/_function_parameter_parse_util.py::_handle_params_as_deferred_annotations`,
reached from `FunctionTool._get_declaration()` ← `llm_request.append_tools()` ←
`base_llm_flow._preprocess_async`. That is during **agent preprocessing**, so the failure
occurs before any LLM call and affects every turn — the ADK agent cannot use the tool at all.

Expected: a NAT tool can be handed to an ADK agent regardless of whether the module that
defines it uses PEP 563.

Two notes on scope:

- A tool whose module does **not** use `from __future__ import annotations` is unaffected.
  This is why the existing unit tests in `packages/nvidia_nat_adk/tests/test_adk_tool_wrapper.py`
  pass: their fixture models are declared without PEP 563.
- The `KeyError` names the *parameter*, with nothing pointing at annotations, which makes it
  hard to diagnose from the traceback alone.

Pydantic already resolves these annotations, so `input_schema.model_fields[...].annotation`
returns the real type in both cases and is a drop-in source for the signature. Happy to open a
PR with that change plus a regression test.

```shell
from __future__ import annotations  # <-- the trigger (PEP 563)

import asyncio

from nat.builder.framework_enum import LLMFrameworkEnum
from nat.cli.register_workflow import register_function
from nat.data_models.function import FunctionBaseConfig
from nat.runtime.loader import PluginTypes
from nat.runtime.loader import discover_and_register_plugins


class EchoConfig(FunctionBaseConfig, name="echo_tool"):
    pass


@register_function(config_type=EchoConfig)
async def echo_tool(config: EchoConfig, builder: Builder):

    async def echo(text: str) -> str:
        """Echo the supplied text."""
        return text

    yield FunctionInfo.from_fn(echo, description="Echo the supplied text.")


async def main():
    discover_and_register_plugins(PluginTypes.ALL)
        await builder.add_function("echo", EchoConfig())
        tool = await builder.get_tool("echo", wrapper_type=LLMFrameworkEnum.ADK)

        print("signature:", tool.func.__signature__)

        # ADK calls this while preparing the request (llm_request.append_tools).
        tool._get_declaration()


asyncio.run(main())
```

### Relevant log output

signature: (text: 'str')

Traceback (most recent call last):
```
  File ".../google/adk/tools/function_tool.py", line 89, in _get_declaration
    build_function_declaration(
  File ".../google/adk/tools/_automatic_function_calling_util.py", line 332, in from_function_with_options
    param = _function_parameter_parse_util._handle_params_as_deferred_annotations(
  File ".../google/adk/tools/_function_parameter_parse_util.py", line 57, in _handle_params_as_deferred_annotations
    param = param.replace(annotation=annotation_under_future[name])
                                     ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
KeyError: 'text'


### Other/Misc.

Removing the from __future__ import annotations line makes the same script print signature: (text: str) and succeed.
nvidia-nat-core 1.8.0
nvidia-nat-adk  1.8.0
google-adk      1.38.0
python          3.11.9
OS              Windows 11
Also reproduced with nvidia-nat-adk 1.5.0 / google-adk 1.25.1. nat/plugins/adk/tool_wrapper.py is byte-identical across 1.5.0, 1.6.0, 1.7.0 and 1.8.0 (sha256 12aefab31718…), so 1.6.0 and 1.7.0 are expected to behave the same — I did not run them.

### Code of Conduct

Christian-Sidak:
## Summary

- Replace `__annotations__` lookup with `model_fields[name].annotation` in the ADK tool wrapper's signature-building logic
- When the input schema class is defined in a module using `from __future__ import annotations` (PEP 563), `__annotations__` holds strings (`'str'`) rather than type objects; Pydantic's `model_fields` always exposes the resolved type
- Falls back to `__annotations__` for non-Pydantic input schemas to preserve existing behaviour
- Adds a regression test that recreates PEP 563 string annotation conditions and verifies the generated `__signature__` carries real type objects

## Root cause

`google_adk_tool_wrapper` called `getattr(input_schema, '__annotations__', {})` to build the wrapped callable's signature. With PEP 563, those annotations are strings. Google ADK's `_handle_params_as_deferred_annotations` then tried to resolve them against the wrapper module's namespace, which does not declare those names, producing `KeyError: '<param_name>'` before any LLM call.

## Test plan

- [ ] All existing tests in `packages/nvidia_nat_adk/tests/test_adk_tool_wrapper.py` still pass
- [ ] New test `test_google_adk_tool_wrapper_pep563_annotations` passes and would have failed before this fix

Fixes #2161

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved generated tool signatures to preserve concrete Python type annotations from Pydantic input models.
  * Added compatibility for schemas that use alternate field definitions.
  * Signature generation continues to be skipped when no input schema is provided.

* **Tests**
  * Added regression coverage for postponed, string-based, and resolved annotations.
  * Verified that wrapped tools expose accurate parameter types.


willkill07:
/ok to test [20dad8c](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2166/commits/20dad8c1333ccc0bd4b05af65a2fb6ef11d8d0d6)

willkill07 (approved):

willkill07:
@Christian-Sidak pre-commit is failing:
https://github.com/NVIDIA/NeMo-Agent-Toolkit/actions/runs/32575091567/job/97036157260?pr=2166#step:5:383

Christian-Sidak:
Thanks for the approval and the heads-up @willkill07. Applied the yapf/ruff changes the hooks reported (dropped the now-unused `inspect` import and reformatted the assert) in b3081b0. CI should be green now.

willkill07:
/ok to test [b3081b0](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2166/commits/b3081b09cddfde95d8e0a3390241ddfdc80c8ba3)

Issue #427, opened 2025-07-13
[BUG]: Failed to work on Windows 11


victortong-git:
### Version

1.2.0 dev

### Which installation method(s) does this occur on?

Source

### Describe the bug.

Dear all,

I configured NeMo Agent Toolkit and trying to create workflow. It has the following error in Win11. Linux is working good.

(.venv) C:\code\xxx>aiq workflow create --workflow-dir my-agents internet_search
2025-07-11 20:59:13,423 - aiq.cli.commands.workflow.workflow_commands - ERROR - Package root does not exist: C:code\xxx\aiqtoolkit
2025-07-11 20:59:13,423 - aiq.cli.commands.workflow.workflow_commands - ERROR - Package root does not exist: C:code\xxx\aiqtoolkit
Installing workflow ‘internet_search’…
2025-07-11 20:59:31,868 - aiq.cli.commands.workflow.workflow_commands - ERROR - An error occurred while creating the workflow: Command ‘[‘pip’, ‘install’, ‘-e’, ‘C:\code\xxx\my-agents\internet_search’]’ returned non-zero exit status 1.
Traceback (most recent call last):
File “C:\code\xxx\aiqtoolkit\src\aiq\cli\commands\workflow\workflow_commands.py”, line 219, in create_command
result = subprocess.run(install_cmd, capture_output=True, text=True, check=True)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File “C:\Users\victo\AppData\Local\Programs\Python\Python312\Lib\subprocess.py”, line 571, in run
raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command ‘[‘pip’, ‘install’, ‘-e’, ‘C:\code\xxx\my-agents\internet_search’]’ returned non-zero exit status 1.
An error occurred while creating the workflow: Command ‘[‘pip’, ‘install’, ‘-e’, ‘C:\code\xxx\my-agents\internet_search’]’ returned non-zero exit status 1.

### Minimum reproducible example

```shell
use the same instruction to install aiqtoolkit and check if aiq --version can work. then, use aiq to create custom workflow. You can see the error message. aiq info having error too.

I followed instruction to sync all groups and all extras.
```

### Relevant log output

<details><summary>Click here to see error details</summary><pre>

 [Paste the error here, it will be hidden by default]

[...]

PR #2169, opened 2026-08-21
docs(get-started): mark native Windows x86_64 unsupported

### LINKED ISSUE #2168: docs: platform table lists native Windows x86_64 as "Should Work" but nat workflow create fails there
### How would you describe the priority of this documentation request

Medium

### Please provide a link or source to the relevant docs

https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/docs/source/get-started/installation.md#supported-platforms

### Describe the problems in the documentation

The Supported Platforms table lists native Windows x86_64 as `❓ Untested, Should Work`. It does not work, and the rest of the page already says so.

The cell is marked "Untested", so here is a measurement. `nat workflow create` — the command used in the first tutorial, `docs/source/get-started/tutorials/create-a-new-workflow.md` — calls `os.symlink()` unconditionally at `packages/nvidia_nat_core/src/nat/cli/commands/workflow/workflow_commands.py:308-309`, with no Windows fallback. On Windows, creating a symlink requires Developer Mode or an elevated process, so on a default install it fails:

```
os.symlink FAILED
  type              : OSError
  winerror          : 1314
  errno             : 22
  is PermissionError: False
  str               : [WinError 1314] A required privilege is not held by the client
```

Measured on Windows 11 x86_64, Python 3.13.13, default (non-elevated, Developer Mode off) shell.

Two details worth noting:

- `WinError 1314` surfaces as a bare `OSError`, **not** `PermissionError`, so an `except PermissionError` guard would not catch it. The call is currently covered only by the broad `except Exception` at the end of `create_command`, which logs and prints — but by then the workflow directory and rendered files already exist, so the user is left with a partially created workflow.
- CI cannot catch this: the GitHub workflows run `ubuntu-latest` and `linux-amd64-cpu4` only, with no Windows job.

This also matches the existing prose on the same page, which the table currently contradicts:

- `installation.md` (Install From Source): "For Windows users, it is recommended to run NeMo Agent Toolkit inside WSL2."
- `installation.md` (Windows (WSL2)): "NeMo Agent Toolkit is developed and tested on Linux and macOS. On Windows, the recommended path is to run the toolkit inside the Windows Subsystem for Linux 2 (WSL2)."

And with #427, where the same command was reported failing on Windows 11 and was closed with "Please use WSL2 for testing on windows. Windows is not natively supported."

### (Optional) Propose a correction

Change the native Windows x86_64 row from `❓ Untested, Should Work` to `❌ Unsupported`, matching the `Windows | aarch64` row directly below it. The two `Windows (WSL2)` rows immediately above already give readers the supported path.

I have opened a PR for exactly that one-cell change.

If you would rather keep the row and make it true, the alternative is a Windows fallback in `workflow_commands.py` — a directory junction or a plain copy instead of a symlink. I did not take that route because #427 indicates native Windows is not a platform you intend to support, and that would be a behaviour change rather than a documentation fix. Happy to be told otherwise.

### Code of Conduct

I agree to follow this project's Code of Conduct. I searched the open documentation issues before filing — there is one open `doc`-labelled issue (#1899, unrelated) and no duplicate of this.

AI assistance was used to investigate and draft this report. I ran the reproduction above myself on my own machine and verified every claim in it.

MohammedAlkindi:
## Description

The Supported Platforms table listed native Windows x86_64 as `❓ Untested, Should Work`. That cell is marked "Untested", so I tested it: it does not work.

`nat workflow create` — the command in the first tutorial — calls `os.symlink()` unconditionally at `packages/nvidia_nat_core/src/nat/cli/commands/workflow/workflow_commands.py:308-309`. On a default Windows install (Developer Mode off, non-elevated) that raises `OSError [WinError 1314] A required privilege is not held by the client`. Full reproduction and measurements are in #2168.

This is a one-cell change to `❌ Unsupported`, matching the `Windows | aarch64` row directly below. It brings the table in line with what the same page already says in prose — "For Windows users, it is recommended to run NeMo Agent Toolkit inside WSL2" and "NeMo Agent Toolkit is developed and tested on Linux and macOS" — and with #427, closed with "Please use WSL2 for testing on windows. Windows is not natively supported." The two `Windows (WSL2)` rows immediately above already point readers at the supported path.

I did not change `workflow_commands.py`. Adding a Windows fallback there would be a behaviour change on a platform #427 says is not supported, and that is your call rather than mine — noted as an alternative in the issue if you would prefer the row kept and made true.

Closes #2168

### Verification

- `pre-commit run markdown-link-check --files docs/source/get-started/installation.md` — Passed.
- Rendered the table to confirm column alignment and that only the one cell changed; the diff is a single line.
- I did not run the full test suite or a docs build. This change touches one table cell in one Markdown file and no code, and I did not have the project installed.

### AI assistance

AI assistance was used to investigate this and draft the change. I reviewed the diff and ran the verification above myself, and the WinError 1314 output in #2168 is from my own machine.

I have not added a `Co-authored-by:` commit trailer — my attribution policy is that commits carry only my identity and AI involvement is disclosed in prose, as here. Say the word if you would rather it were in the commit metadata.

- The commit is signed off (DCO).
- No new or existing tests are affected — this is a documentation-only change.
- The documentation is up to date with these changes.

## Summary by CodeRabbit

* **Documentation**
  * Updated the supported-platforms table to indicate that native Windows x86_64 is unsupported.

### DIFF
--- a/docs/source/get-started/installation.md
+++ b/docs/source/get-started/installation.md
@@ -80,7 +80,7 @@ To install these first-party plugin libraries, you can use the full distribution
 | macOS | aarch64 | 3.11, 3.12, 3.13 | ✅ Tested |
 | [Windows (WSL2)](#windows-wsl2) | x86_64 | 3.11, 3.12, 3.13 | ✅ Tested |
 | [Windows (WSL2)](#windows-wsl2) | aarch64 | 3.11, 3.12, 3.13 | ❓ Untested, Should Work |
-| Windows | x86_64 | 3.11, 3.12, 3.13 | ❓ Untested, Should Work |
+| Windows | x86_64 | 3.11, 3.12, 3.13 | ❌ Unsupported |
 | Windows | aarch64 | 3.11, 3.12, 3.13 | ❌ Unsupported |
 
 ## Software Prerequisites


willkill07:
/ok to test [fc22aa1](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2169/commits/fc22aa1649bce0700ee0f81d6dda07574c2d5bdd)

willkill07 (approved):

Issue #2162, opened 2026-08-18
NAT A2A & MCP per_user workflow not starting


danielphilipp:
### Version

v1.8.0

### Describe the bug.

Hi,

yesterday we wanted to start agentic workflows using Nemo Agent Toolkit v1.8 but it fails to start an workflow and raises an error for a specific combination of configurations. We believe it is a bug.

Our current setup:

We have an orchestrator agent workflow running and want to connect an sub-agent using A2A. The sub agent has a connect to an authenticated MCP server. As stated in the docs both, the orchestrator and the sub-agent are "per_user_react_agent" type and the MCP connect in the sub-agent is "per_user_mcp_client" and has an "mcp_oauth2" authentication method. However the sub-agent workflow fails to start with the following error:

 [Paste the error here,2026-08-17 13:56:56 - ERROR    - nat.cli.commands.start:243 - Failed to initialize workflow
Error: Must set a workflow before building

This is the whole traceback. We managed to narrow down the problem: The workflow launch fails when the sub-agent is "per_user_react_agent" and the mcp server is "per_user_mcp_client" type. When using the non per_user types the workflow starts but fails with an 401 Unauthorized for the MCP initialization, which is expected since we have an authenicated MCP server. 

Are there problems with an A2A setup and per_user workflows in combination with MCP?

Please find attached our configs and startup commands. We are using Docker Containers as we are running in K8s. The images were built from the Dockerfile in the v1.8.0 tagged git release. The orchestrator is working as expected, only the sub-agent with the MCP connection raises the problems.

[orchestrator_start_server.py](https://github.com/user-attachments/files/31181829/orchestrator_start_server.py)
[orchestrator_agent_config.yml](https://github.com/user-attachments/files/31181830/orchestrator_agent_config.yml)

[subagent_config.yml](https://github.com/user-attachments/files/31181852/subagent_config.yml)
[subagent_start_server.py](https://github.com/user-attachments/files/31181853/subagent_start_server.py)

We need your help to solve this.

Regards
Daniel

### Minimum reproducible example

```shell
See attachments
```

### Relevant log output

 [Paste the error here,2026-08-17 13:56:56 - ERROR    - nat.cli.commands.start:243 - Failed to initialize workflow
Error: Must set a workflow before building

### Code of Conduct

- [x] I agree to follow the NeMo Agent Toolkit Code of Conduct
[...]

PR #2170, opened 2026-08-22
fix(mcp,fastmcp): serve per-user workflows from MCP front ends

BarneyChambers:
## Description

The MCP and FastMCP front ends called `builder.build()` unconditionally at startup. The shared builder deliberately leaves a per-user workflow unset, because `SessionManager` builds one per user on demand, so every per-user config died with `Must set a workflow before building` before the server came up.

That configuration is what [MCP authentication](https://github.com/NVIDIA/NeMo-Agent-Toolkit/blob/develop/docs/source/components/auth/mcp-auth/index.md) prescribes: pair `per_user_mcp_client` with a per-user workflow such as `per_user_react_agent`.

Reproduced on `develop` at `c933737c`, doing exactly what the front end does:

```
shared workflow      OK, workflow built
per_user workflow    ValueError: Must set a workflow before building
```

Related to #2162. The A2A portion of that report is in #2167.

### Startup

Build session managers with `SessionManager.create`, which is what the other front ends use. It builds the shared workflow, or leaves it unset and starts the per-user builder reaper. Workers track managers and call `shutdown()` on exit.

### Tool registration and runtime

`register_function_with_mcp` no longer reads `session_manager.workflow` for per-user configs. It uses `get_workflow_input_schema()` instead. Tool wrappers call `session(user_id=...)` for per-user runs, resolving the user from runtime context or the Bearer token on the MCP request.

Serving a per-user workflow requires `server_auth`. Without it there is no user to build for, and `session()` raises a clear error naming the missing user ID.

## Testing

- `pytest packages/nvidia_nat_mcp/tests/server/test_per_user_workflow.py`: **6 passed**
- `pytest packages/nvidia_nat_fastmcp/tests/server/test_per_user_workflow.py`: **6 passed**
- `pytest packages/nvidia_nat_mcp/tests/`: **392 passed**, 8 skipped
- `pytest packages/nvidia_nat_fastmcp/tests/`: **33 passed**
- `pre-commit run --from-ref origin/develop --to-ref HEAD`: yapf, ruff, markdown link check passed
- `ci/scripts/copyright.py --verify-apache-v2`: passed

Not covered: a live MCP tool call against a real OAuth2 issuer.

- New or existing tests cover these changes.
- Documentation is up to date with these changes (`docs/source/run-workflows/mcp-server.md`, `docs/source/run-workflows/fastmcp-server.md`).

## Summary by CodeRabbit

* **New Features**
  * Added support for serving per-user workflows through MCP and FastMCP.
  * Per-user requests resolve identity from authenticated bearer tokens or runtime connection context.
  * Workflow-specific tools include tailored names, schemas, and descriptions.

* **Bug Fixes**
  * Improved cleanup of workflow sessions during normal shutdown and interruptions.
  * Debug tool listings now omit per-user workflows.

* **Documentation**
  * Clarified authentication requirements, transport limitations, and tool-inspection guidance for per-user workflows.

### DIFF
--- a/docs/source/run-workflows/fastmcp-server.md
+++ b/docs/source/run-workflows/fastmcp-server.md
@@ -154,9 +154,9 @@ With this configuration, the MCP server is accessible at `http://localhost:9902/
 
 A per-user workflow can be served over FastMCP. Each user gets their own workflow instance, built on first use. [MCP authentication](../components/auth/mcp-auth/index.md) prescribes pairing `per_user_mcp_client` with a per-user workflow, such as `per_user_react_agent`.
 
-The user is taken from the Bearer token on the MCP request, or from runtime context when it is already set. Configure [`server_auth`](../components/auth/mcp-auth/index.md) when serving a per-user workflow over FastMCP. Without it there is no user to build for, and tool calls fail with an error saying the user ID could not be determined.
+The user is taken from the Bearer token on the MCP request when [`server_auth`](#authentication) is configured, or from runtime context when it is already set. Per-user workflows require authenticated streamable-http requests so the server can derive a user ID. Without that, tool calls fail with an error saying the user ID could not be determined.
 
-The `/debug/tools/list` route for a per-user workflow shows the registered workflow tool and its input schema, but the functions inside the workflow belong to each user's instance rather than to a shared one.
+The `/debug/tools/list` route returns an empty tool list for a per-user workflow because no shared workflow instance exists at startup. Use `nat mcp client tool list` to inspect the registered MCP tool and its input schema.
 
 ## Inspecting and Running MCP Tools Published by a FastMCP Server
 
--- a/docs/source/run-workflows/fastmcp-server.md
+++ b/docs/source/run-workflows/fastmcp-server.md
@@ -150,6 +150,14 @@ general:
 
 With this configuration, the MCP server is accessible at `http://localhost:9902/api/v1/mcp`.
 
+### Per-User Workflows
+
+A per-user workflow can be served over FastMCP. Each user gets their own workflow instance, built on first use. [MCP authentication](../components/auth/mcp-auth/index.md) prescribes pairing `per_user_mcp_client` with a per-user workflow, such as `per_user_react_agent`.
+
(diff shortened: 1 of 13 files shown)


willkill07:
@BarneyChambers the CodeRabbit feedback seems addressable/valid, especially the cleanup one.

BarneyChambers:
Pushed a few fixes from the review and a local CodeRabbit run.

**Context on tool wrappers**: Per-user wrappers now take a `Context`-typed `ctx` param so FastMCP can inject the MCP request and we can pull the user ID from the Bearer token. `ctx` is still kept out of the client-facing schema via `__signature__`.

**Docs**: Fixed the per-user `/debug/tools/list` bit. That route comes back empty at startup because there is no shared workflow instance yet. `nat mcp client tool list` is the right way to see the registered tool and schema.

**Cleanup**: `cleanup()` shuts down every tracked `SessionManager` even if one fails, clears the list either way, and re-raises the first error.

**Auth docs**: Updated `mcp-server.md` to describe `server_auth` on streamable-http instead of the old "no built-in auth" line.

Thanks @willkill07 for taking a look at this on the weekend, much appreciated.

BarneyChambers:
### Per-user MCP/FastMCP follow-up

After Will's review and a local CodeRabbit pass, I pushed two rounds of fixes on top of the original startup work.
### Worker cleanup
If one `SessionManager` failed to shut down, we stopped and could leave the others running. Now we try all of them, clear the tracked list either way, log what failed, and re-raise the first error.
### Docs
The per-user sections said `/debug/tools/list` would show the workflow tool but it currently doesn't. That route comes back empty because there's no shared workflow at startup. The docs now say that and tell people to use `nat mcp client tool list` instead.

I also fixed `mcp-server.md`, which still claimed MCP serve had no built-in auth. Streamable-http `server_auth` is documented now.

### Request context for per-user tool calls
Per-user workflows need the MCP request to work out who the user is from the Bearer token. Our first pass tried to pull `ctx` out of `**kwargs`, which FastMCP doesn't inject like that.

We switched to `fastmcp.server.context.Context` and put `ctx` on the wrapper signature so FastMCP can inject it, without exposing it in the client tool schema. There's a test that calls the tool with a fake HTTP request and checks the user ID comes from the Bearer token.

### Tests
- Cleanup when one manager fails
- Context injection on the wrapper
- Bearer token user ID resolution

The rest of what CodeRabbit flagged is mostly type hints in tests. I am including these fixes in the below commit

BarneyChambers:
> @BarneyChambers the CodeRabbit feedback seems addressable/valid, especially the cleanup one.

Everything has been addressed, thank you!

PR #2174, opened 2026-08-23
fix(optimizer): reject empty search spaces

### LINKED ISSUE #2172: Empty SearchSpace() passes validation and fails inside Optuna instead
### Version

`develop` at 15a28d6

Source

### Describe the bug.

`SearchSpace.validate_search_space_parameters` rejects every malformed combination except
one: an entirely empty `SearchSpace()`. No `values`, no `low`, no `high`. That object is
accepted at config-parse time and only fails once the optimizer is already running, inside
Optuna, as `TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'`.

`suggest` at `packages/nvidia_nat_core/src/nat/data_models/optimizable.py:100` falls through
the `values` branch and the `isinstance(self.low, int)` branch, and hands `None` to
`trial.suggest_float` as both bounds:

```python
if self.values is not None:
    return trial.suggest_categorical(name, self.values)
if isinstance(self.low, int):
    return trial.suggest_int(name, self.low, cast(int, self.high), ...)
return trial.suggest_float(name, cast(float, self.low), cast(float, self.high), ...)
```

The `cast()` calls say these are floats. At runtime they are `None`.

The same object handled by `to_grid_values` on line 125 gives a clear config error instead:

```
ValueError: Grid search requires either 'values' or both 'low' and 'high' to be defined
```

So the file already treats an empty search space as invalid. The disagreement is only about
where the user finds out: grid search says so immediately, Optuna says so mid-run with an
error that names neither the parameter nor the config.

The range-based section of the validator is where this belongs, next to the XOR check that
already catches `low` without `high`.

```python
from nat.data_models.optimizable import SearchSpace
import optuna

space = SearchSpace()          # accepted, no error
print(space.model_dump(exclude_none=True))
# {'log': False, 'is_prompt': False}

optuna.create_study().optimize(lambda t: space.suggest(t, "lr"), n_trials=1)
# TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'

space.to_grid_values()
# ValueError: Grid search requires either 'values' or both 'low' and 'high' to be defined
```

For contrast, every neighbouring mistake is caught at construction:

```python
SearchSpace(low=1)                          # rejected: needs both low and high
SearchSpace(high=2)                         # rejected: needs both low and high
SearchSpace(values=[1], low=1, high=2)      # rejected: mutually exclusive
SearchSpace(values=[])                      # rejected: must not be empty
SearchSpace(low=5, high=1)                  # rejected: low must be less than high
SearchSpace(is_prompt=True, low=1, high=2)  # rejected: prompt spaces take no bounds
SearchSpace()                               # accepted
```

### Relevant log output

```shell
Trial 0 failed with parameters: {} because of the following error:
TypeError("'>' not supported between instances of 'NoneType' and 'NoneType'")
  File ".../nat/data_models/optimizable.py", line 103, in suggest
    return trial.suggest_float(name, cast(float, self.low), cast(float, self.high), log=self.log, step=self.step)
  File ".../optuna/distributions.py", line 145, in __init__
    if low > high:
TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'
```

### Other/Misc.

Reproduced against `optimizable.py` on develop with pydantic 2.13.4 and optuna 4.9.0.

I am happy to fix this and add the validator case plus a test, if you want it. I have not
opened a PR and will wait to be assigned first. If you would rather `suggest` raise its own
error than have the validator reject the object, say so and I will do it that way instead,
though the validator seems the better place given `to_grid_values` already rejects it.

primorLee:
## Description

Reject a non-prompt SearchSpace that defines neither categorical values nor numeric bounds during Pydantic configuration validation. This prevents an empty configuration from reaching Optuna as two None bounds and failing mid-run with an unrelated TypeError.

The existing grid-search runtime test is moved to the shared construction-time validation contract, and the complete SearchSpace test module remains green.

Closes #2172

- I am familiar with the Contributing Guidelines.
- The commit is signed off under the Developer Certificate of Origin.
- Existing and new tests cover these changes.
- No documentation update is required because this only enforces the already documented search-space contract.

## Validation
- python -m pytest packages/nvidia_nat_core/tests/nat/data_models/test_optimizable.py -q (45 passed)
- python -m ruff check on both changed files
- YAPF 0.43.0 diff check on both changed files

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for search space configurations.
  * Configurations must now provide either explicit values or both lower and upper bounds.
  * Invalid configurations are rejected during creation with clearer behavior.

### DIFF
--- a/packages/nvidia_nat_config_optimizer/src/nat/plugins/config_optimizer/optimizable_utils.py
+++ b/packages/nvidia_nat_config_optimizer/src/nat/plugins/config_optimizer/optimizable_utils.py
@@ -83,7 +83,9 @@ def walk_optimizables(obj: BaseModel, path: str = "") -> dict[str, SearchSpace]:
                 _, val_t = get_args(ann) or (None, None)
                 if isinstance(val_t, type) and issubclass(val_t, BaseModel):
                     if allowed_params is None or name in allowed_params:
-                        spaces[f"{full}.*"] = SearchSpace(low=None, high=None)  # sentinel
+                        # This is an internal wildcard marker, not a runnable search space.
+                        # Construct it without applying the public configuration validator.
+                        spaces[f"{full}.*"] = SearchSpace.model_construct()
 
     if allowed_params is None and has_optimizable_flag:
         logger.warning(
--- a/packages/nvidia_nat_core/src/nat/data_models/optimizable.py
+++ b/packages/nvidia_nat_core/src/nat/data_models/optimizable.py
@@ -86,6 +86,8 @@ class SearchSpace(BaseModel, Generic[T]):
         # 3. Range-based validation
         if (self.low is None) != (self.high is None):  # XOR using !=
             raise ValueError(f"SearchSpace range requires both 'low' and 'high'; got low={self.low}, high={self.high}")
+        if self.low is None and self.high is None:
+            raise ValueError("SearchSpace requires either 'values' or both 'low' and 'high' to be defined")
         if self.low is not None and self.high is not None and self.low >= self.high:
             raise ValueError(f"SearchSpace 'low' must be less than 'high'; got low={self.low}, high={self.high}")
 
--- a/packages/nvidia_nat_core/tests/nat/data_models/test_optimizable.py
+++ b/packages/nvidia_nat_core/tests/nat/data_models/test_optimizable.py
@@ -229,11 +229,6 @@ class TestSearchSpaceToGridValues:
         with pytest.raises(ValueError, match="Log scale is not yet supported for grid search"):
             space.to_grid_values()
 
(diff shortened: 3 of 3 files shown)


willkill07:
/ok to test [4128a62](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2174/commits/4128a621d62d6b2f75a71ba0aff26ef73e36749b)

willkill07 (approved):

willkill07:
/ok to test [484ae94](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2174/commits/484ae943b2e2003f4ccf811cb4573c960424710d)

PR #2186, opened 2026-08-28
fix: Preserve optional defaults in ADK tool signatures

### LINKED ISSUE #2183: ADK tool wrapper drops optional-parameter defaults, so every field is declared required
### Version

PyPi

### Describe the bug.

`google_adk_tool_wrapper` builds the wrapped callable's `__signature__` with
`inspect.Parameter(name, POSITIONAL_OR_KEYWORD, annotation=...)` and no `default`. Every
input-schema field therefore becomes a required parameter, even when the schema says it is
optional with a default.

The code, `packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py` on `develop` @ 2d2e342:
```
inspect.Parameter(
    param_name,
    inspect.Parameter.POSITIONAL_OR_KEYWORD,
    annotation=resolve_type(param_annotation),
)
```

Google ADK derives its function declaration from that signature, so it tells the model that
optional parameters are required. The model must then invent a value for each one on every
call, and the tool's own defaults never apply. For a tool like
`search(query, max_results=5, region="global")` the model is forced to supply `max_results`
and `region` it was never asked about - quietly changing the tool's behaviour rather than
failing loudly.

Expected: a field that is optional in the input schema is optional in the ADK declaration.

The information needed is already in hand at that point. #2166 changed this same loop to
iterate Pydantic's `model_fields`, and the `FieldInfo` it now yields carries both `default`
and `is_required()` - so the fix reads two more attributes off the object the loop already
has.

One ordering detail for whoever fixes it: a parameter with a default cannot precede one
without, and Pydantic preserves declaration order, which may interleave required and optional
fields. The parameters need sorting (required first) or `inspect.Signature` raises
`ValueError`. NAT invokes the wrapped function with keyword arguments, so reordering is safe.

Related to #2161 / #2166 - same function, but a separate defect. That crash masked this one;
with #2166 merged it is now directly observable on `develop`.

Suggested fix, against `develop` @ 2d2e342. A helper that reads the default off the
`FieldInfo` the loop already has:

```python
def _field_default(field_info: Any) -> Any:
    """Return a Pydantic field's default, or ``inspect.Parameter.empty`` if required."""
    if field_info.is_required():
        return inspect.Parameter.empty
    if field_info.default_factory is not None:
        try:
            return field_info.default_factory()
        except TypeError:
            # Factories that need validated data cannot be evaluated here.
            return None
    return field_info.default
```

Then carry the default through into the signature:

```python
params: list[inspect.Parameter] = []
if input_schema is not None:
    model_fields = getattr(input_schema, "model_fields", None)
    if model_fields is not None:
        field_items = ((n, f.annotation, _field_default(f)) for n, f in model_fields.items())
    else:
        field_items = ((n, a, getattr(input_schema, n, inspect.Parameter.empty))
                       for n, a in getattr(input_schema, "__annotations__", {}).items())

    for param_name, param_annotation, param_default in field_items:
        params.append(
            inspect.Parameter(
                param_name,
                inspect.Parameter.POSITIONAL_OR_KEYWORD,
                annotation=resolve_type(param_annotation),
                default=param_default,
            ))

    # A parameter with a default cannot precede one without, and Pydantic
    # preserves declaration order, which may interleave them. NAT invokes the
    # wrapped function with keyword arguments, so reordering is safe.
    params.sort(key=lambda p: p.default is not inspect.Parameter.empty)

setattr(func_to_wrap, "__signature__", inspect.Signature(parameters=params))
```

`default_factory` is evaluated where it can be, and falls back to `None` for factories that
need validated data, rather than leaking a factory object into the signature. The non-Pydantic
branch keeps working - a class attribute is used as the default if one exists, otherwise the
field stays required, which matches today's behaviour. The `sort` is what stops
`inspect.Signature` raising `ValueError` when a schema interleaves required and optional
fields.

With this applied the reproducer above prints:

```
signature      : (query: str, max_results: int = 5, region: str = 'global')
ADK required   : ['query']
```

```shell
import asyncio

from nat.builder.framework_enum import LLMFrameworkEnum
from nat.cli.register_workflow import register_function
from nat.data_models.function import FunctionBaseConfig
from nat.runtime.loader import PluginTypes
from nat.runtime.loader import discover_and_register_plugins


class SearchConfig(FunctionBaseConfig, name="search_tool"):
    pass


@register_function(config_type=SearchConfig)
async def search_tool(config: SearchConfig, builder: Builder):

    async def search(query: str, max_results: int = 5, region: str = "global") -> str:
        """Search. query is required; max_results and region are optional."""
        return f"{query}|{max_results}|{region}"

    yield FunctionInfo.from_fn(search, description="Search with optional filters.")


async def main():
    discover_and_register_plugins(PluginTypes.ALL)
        await builder.add_function("search", SearchConfig())

        schema = (await builder.get_function("search")).input_schema
        print("schema required:", sorted(n for n, f in schema.model_fields.items() if f.is_required()))

        tool = await builder.get_tool("search", wrapper_type=LLMFrameworkEnum.ADK)
        print("signature      :", tool.func.__signature__)
        print("ADK required   :", sorted(tool._get_declaration().parameters.required or []))


asyncio.run(main())
```

### Relevant log output

```
schema required: ['query']
signature      : (query: str, max_results: int, region: str)
ADK required   : ['max_results', 'query', 'region']

```

### Other/Misc.

max_results and region are optional in the schema and required in the declaration. With the defaults preserved the same script prints:

```
signature      : (query: str, max_results: int = 5, region: str = 'global')
ADK required   : ['query']
```

Run it against develop (or any build including #2166). On the released 1.8.0 wheel the KeyError from #2161 fires first and masks this.

Environment:

```
nvidia-nat-core 1.8.0
nvidia-nat-adk  1.8.0
google-adk      1.38.0
python          3.11.9
OS              Windows 11
```

### Code of Conduct

mikemikimike:
## Description

Closes #2183.

`google_adk_tool_wrapper` rebuilt wrapped call signatures without carrying defaults from the Pydantic input model. Google ADK therefore advertised optional tool fields as required.

This change preserves Pydantic field defaults, including default factories, in the wrapped signature and orders required parameters before optional parameters so the resulting Python signature remains valid.

## Compatibility

Required fields remain required. Fields with defaults are now optional in the Google ADK tool declaration. The change does not require an external LLM or API service.

## Tests

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH=test_stubs;packages/nvidia_nat_adk/src;packages/nvidia_nat_core/src uv run --offline --python 3.12 --no-project --with "pydantic>=2,<3" --with pytest --with pytest-asyncio python -m pytest -p pytest_asyncio.plugin --asyncio-mode=auto packages/nvidia_nat_adk/tests/test_adk_tool_wrapper.py -q` — 9 passed, covering Pydantic v1/v2 metadata, annotation-only defaults, parameter ordering, and data-aware `default_factory` handling.
- `uv run --offline --no-project --with "pydantic>=2,<3" --with mypy python -m mypy --ignore-missing-imports packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py packages/nvidia_nat_adk/tests/test_adk_tool_wrapper.py` — passed.
- `python -m ruff check packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py packages/nvidia_nat_adk/tests/test_adk_tool_wrapper.py` — passed.
- `python -m compileall -q packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py packages/nvidia_nat_adk/tests/test_adk_tool_wrapper.py` — passed.
- `python -m ruff format --check packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py packages/nvidia_nat_adk/tests/test_adk_tool_wrapper.py` — reports both files would be reformatted; no formatter writes were kept because the same result is present before this change.

## Limitations

The full repository checkout and full CI suite were not run because GitHub Git transport and codeload downloads terminated with TLS EOF errors in the local environment. The focused test was run against a REST-reconstructed source subset with test-only import stubs; the stubs are not part of this PR. Running the other ADK package tests in that subset stops during collection because the subset does not contain the core `nat.data_models` and `nat.llm` modules those tests import.

- I am familiar with the Contributing Guidelines and have followed them.
- I have signed off on my commit as required by the Developer Certificate of Origin.
- When the PR is ready for review, the documentation is up to date or not required for this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool parameter handling so required fields and configured defaults are preserved correctly.
  * Added support for literal defaults, generated list defaults, and data-aware default values.
  * Improved compatibility with legacy and annotation-only schema formats.
  * Prevented errors when processing default factories that require validated input data.

* **Tests**
  * Added regression coverage for field metadata, required-field ordering, and supported default types.

### DIFF
--- a/packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py
+++ b/packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Tool Wrapper file"""
-
 import logging
 import types
 from collections.abc import AsyncIterator
@@ -56,7 +55,7 @@ def resolve_type(t: Any) -> Any:
 def google_adk_tool_wrapper(
     name: str,
     fn: Function,
-    _builder: Builder,  # pylint: disable=W0613
+    _builder: Builder  # pylint: disable=W0613
 ) -> Any:  # Changed from Callable[..., Any] to Any to allow FunctionTool return
     """Wrap a NAT `Function` as a Google ADK `FunctionTool`.
 
@@ -152,9 +151,8 @@ def google_adk_tool_wrapper(
                 if model_fields is not None:
                     field_items = ((n, f.annotation, _field_default(f)) for n, f in model_fields.items())
                 else:
-                    field_items = (
-                        (n, a, inspect.Parameter.empty) for n, a in getattr(input_schema, "__annotations__", {}).items()
-                    )
+                    field_items = ((n, a, inspect.Parameter.empty)
+                                   for n, a in getattr(input_schema, "__annotations__", {}).items())
                 for param_name, param_annotation, default in field_items:
                     params.append(
                         inspect.Parameter(
@@ -162,8 +160,7 @@ def google_adk_tool_wrapper(
                             inspect.Parameter.POSITIONAL_OR_KEYWORD,
                             annotation=resolve_type(param_annotation),
                             default=default,
-                        )
-                    )
+                        ))
             params.sort(key=lambda param: param.default is not inspect.Parameter.empty)
             setattr(func_to_wrap, "__signature__", inspect.Signature(parameters=params))
 
--- a/packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py
+++ b/packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py
@@ -77,12 +77,7 @@ def google_adk_tool_wrapper(
             return inspect.Parameter.empty
         default_factory = getattr(field, "default_factory", None)
         if default_factory is not None:
-            if getattr(field, "default_factory_takes_validated_data", False):
-                return None
-            try:
-                return default_factory()
-            except TypeError:
-                return None
+            return None
         return getattr(field, "default", inspect.Parameter.empty)
 
     async def callable_ainvoke(*args: Any, **kwargs: Any) -> Any:
--- a/packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py
+++ b/packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py
@@ -77,7 +77,10 @@ def google_adk_tool_wrapper(
             return inspect.Parameter.empty
         default_factory = getattr(field, "default_factory", None)
         if default_factory is not None:
-            return default_factory()
+            try:
+                return default_factory()
+            except TypeError:
+                return None
         return getattr(field, "default", inspect.Parameter.empty)
 
     async def callable_ainvoke(*args: Any, **kwargs: Any) -> Any:
--- a/packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py
+++ b/packages/nvidia_nat_adk/src/nat/plugins/adk/tool_wrapper.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Tool Wrapper file"""
+
 import logging
 import types
 from collections.abc import AsyncIterator
@@ -55,7 +56,7 @@ def resolve_type(t: Any) -> Any:
 def google_adk_tool_wrapper(
     name: str,
     fn: Function,
-    _builder: Builder  # pylint: disable=W0613
+    _builder: Builder,  # pylint: disable=W0613
 ) -> Any:  # Changed from Callable[..., Any] to Any to allow FunctionTool return
     """Wrap a NAT `Function` as a Google ADK `FunctionTool`.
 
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
Could we avoid executing default_factory() while constructing the tool signature?

Pydantic default factories can be dynamic or stateful, for example generating a UUID or timestamp, and may also have side effects. Calling the factory here executes user code during tool registration and freezes one generated value into the ADK signature rather than preserving the factory's per-validation semantics.

If the main requirement is to mark the parameter as optional, it would be safer to derive that without invoking the factory and leave actual factory execution to Pydantic when the input model is instantiated.

mikemikimike:
I rechecked this PR for issue #2183. The current review history has no changes-requested review, and there are no failing checks reported for the current head. Could a maintainer please review whether this change fully resolves the issue and approve it, or advise on any remaining work?

sylvesterkaczmarek (approved):
Rechecked the current head. The signature builder now marks default-factory fields optional without invoking `default_factory`; the regression verifies both zero registration-time calls and independent factory execution during Pydantic validation. My earlier concern is resolved.

willkill07 (approved):

PR #2191, opened 2026-08-29
fix(core): support zero-argument functions

### LINKED ISSUE #2184: A NAT function cannot take zero arguments: FunctionInfo.from_fn rejects no-input tools
### Version

PyPi

### Describe the bug.

`FunctionInfo.from_fn` cannot register a function that takes no parameters. A zero-argument
function is recognised when the descriptor is built, but never wrapped, so it reaches a
validator that requires exactly one parameter and raises.

`FunctionDescriptor.from_function` handles the zero case explicitly
(`packages/nvidia_nat_core/src/nat/builder/function_info.py`, `develop` @ 2d2e342):

```python
arg_count = len(sig.parameters)

if (arg_count == 0):
    input_type = NoneType
    is_input_typed = False
    input_schema = NoneType
elif (arg_count == 1):
    ...
```

But `FunctionInfo.create` only wraps the multi-argument case:

```python
if (final_single_fn_desc.arg_count > 1):
    if (input_schema is not None):
        logger.warning("Using provided input_schema for multi-argument function")
    else:
        input_schema = final_single_fn_desc.get_base_model_function_input()
    ...
    final_single_fn = _convert_input_pydantic
```

Functions with more than one parameter are therefore collapsed into a single-parameter
function whose input is a generated Pydantic model. There is no corresponding
`arg_count == 0` branch, so a zero-argument function passes through unchanged and hits
`_validate_single_fn`:

```python
if len(sig.parameters) != 1:
    raise ValueError("single_fn must have exactly one parameter")
```

The result is that 1 and 3 parameters both work while 0 does not, which reads as an
oversight rather than an intentional constraint - the zero case is recognised at the
descriptor level and then not carried through.

No-input tools are common - fetch the current policy, get the current time, list everything,
health check. The workaround is to add a dummy parameter the tool ignores. That parameter is
then advertised to the model, which has to invent a value for it, so the workaround is
visible in the model's tool schema rather than hidden in the adapter.

Expected: a zero-argument function registers, with an empty input schema.

Suggested approach: mirror the existing `arg_count > 1` handling with an `arg_count == 0`
branch that wraps the function in a single-parameter shim taking an empty Pydantic model, so
the rest of the pipeline continues to see exactly one input. The decision worth making
explicitly is what that empty schema should look like on the wire for the framework adapters
that read it - which is why this is filed as an issue rather than a PR.

Sketch of that branch, mirroring the existing multi-argument one in `FunctionInfo.create`
(`develop` @ 2d2e342). Not tested - the empty-schema question below should be settled first:

```python
if (final_single_fn_desc.arg_count == 0):
    if (input_schema is None):
        # An input model with no fields, so the rest of the pipeline still
        # sees exactly one parameter.
        input_schema = create_model(f"{final_single_fn.__name__}Input")

    saved_final_single_fn = final_single_fn

    async def _discard_empty_input(value: input_schema) -> final_single_fn_desc.output_type:
        # Nothing to unpack - the function takes no arguments.
        return await saved_final_single_fn()

    final_single_fn = _discard_empty_input

    # Reset the descriptor
    final_single_fn_desc = FunctionDescriptor.from_function(final_single_fn)

elif (final_single_fn_desc.arg_count > 1):
    ...  # unchanged
```

The same treatment would be needed on the streaming path, which has the matching
`final_stream_fn_desc.arg_count > 1` branch and no zero case either.

The part that is genuinely a maintainer decision, and the reason this is an issue rather than
a PR: what an empty input schema should look like on the wire. A field-less Pydantic model
serialises to `{"type": "object", "properties": {}}`, which the framework adapters then hand
to their own tool-declaration builders. Whether every adapter is happy with an empty
properties object - or whether a no-input tool should be represented some other way - is a
call about NAT's public contract, not something worth guessing at from outside.

```shell
import asyncio
import inspect

async def no_args() -> str:
    """A tool that needs no input, e.g. fetch the current policy."""
    return "policy text"


async def one_arg(text: str) -> str:
    """One parameter."""
    return text


async def three_args(query: str, limit: int = 5, region: str = "global") -> str:
    """Three parameters, two optional."""
    return f"{query}{limit}{region}"


async def main():
    for fn in (no_args, one_arg, three_args):
        count = len(inspect.signature(fn).parameters)
        try:
            FunctionInfo.from_fn(fn, description=fn.__doc__)
            print(f"{fn.__name__:<12} ({count} params): OK")
        except Exception as exc:
            print(f"{fn.__name__:<12} ({count} params): {type(exc).__name__}: {exc}")


asyncio.run(main())
```

### Relevant log output

```
no_args      (0 params): ValueError: single_fn must have exactly one parameter
one_arg      (1 params): OK
three_args   (3 params): OK
```

### Other/Misc.

Traceback for the failing case:

```
  File ".../nat/builder/function_info.py", line 621, in from_fn
    return FunctionInfo.create(single_fn=single_fn, ...)
  File ".../nat/builder/function_info.py", line 543, in create
    return FunctionInfo(single_fn=final_single_fn, ...)
  File ".../nat/builder/function_info.py", line 311, in __init__
    single_input_type, single_output_type = _validate_single_fn(single_fn)
  File ".../nat/builder/function_info.py", line 61, in _validate_single_fn
    raise ValueError("single_fn must have exactly one parameter")
ValueError: single_fn must have exactly one parameter
```

Environment:

```
nvidia-nat-core 1.8.0
python          3.11.9
OS              Windows 11
```

### Code of Conduct

mikemikimike:
## Summary

`FunctionInfo.from_fn` rejected valid zero-argument coroutine and async-generator functions because the function descriptor recognized the zero-argument case but `FunctionInfo.create` only adapted functions with multiple arguments.

This change adapts zero-argument functions to the existing single-input pipeline with an empty Pydantic input model. The original callable is invoked without arguments, while framework consumers continue to receive one validated input object. Both single and streaming function paths are covered, including explicit conversion helpers and `Streaming` metadata; existing one- and multi-argument behavior is unchanged.

## Validation

- `PYTHONPATH=packages/nvidia_nat_core/src python -m pytest packages/nvidia_nat_core/tests/nat/builder/test_function_info.py -q --no-cov` — 128 passed
- `ruff check packages/nvidia_nat_core/src/nat/builder/function_info.py packages/nvidia_nat_core/tests/nat/builder/test_function_info.py` — passed
- `python -m compileall -q packages/nvidia_nat_core/src/nat/builder/function_info.py packages/nvidia_nat_core/tests/nat/builder/test_function_info.py` — passed
- `git diff --check` — passed

The full core-package test collection was attempted locally but is limited by missing optional dependencies and Windows-only incompatibilities; the focused builder suite runs successfully.

Closes #2184.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved support for zero-argument functions that produce single results or streams.
  - Fixed conversions between single-result and streaming outputs for functions without input parameters.
  - Preserved streamed values when converting results, including annotated streaming functions.
  - Added empty input schemas for zero-argument functions, ensuring these functions execute without requiring input messages.
  - Improved handling of streaming functions that convert multiple emitted values into a single result.

- **Tests**
  - Added coverage for zero-argument streaming functions and single/stream result conversions.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/builder/function_info.py
+++ b/packages/nvidia_nat_core/src/nat/builder/function_info.py
@@ -674,7 +674,7 @@ class FunctionInfo:
                         single_fn = _stream_to_single_output_empty
 
                     else:
-                        single_input_type = sig.parameters[list(sig.parameters.keys())[0]].annotation
+                        single_input_type = next(iter(sig.parameters.values())).annotation
 
                         async def _stream_to_single_output(message: single_input_type) -> single_output_type:
                             values = []
--- a/packages/nvidia_nat_core/src/nat/builder/function_info.py
+++ b/packages/nvidia_nat_core/src/nat/builder/function_info.py
@@ -453,7 +453,22 @@ class FunctionInfo:
 
             final_single_fn_desc = FunctionDescriptor.from_function(final_single_fn)
 
-            if (final_single_fn_desc.arg_count > 1):
+            if (final_single_fn_desc.arg_count == 0):
+                input_schema = input_schema or create_model("EmptyInputSchema")
+
+                saved_final_single_fn = final_single_fn
+
+                async def _convert_input_pydantic_empty(
+                        value: input_schema) -> final_single_fn_desc.output_type:
+                    no_args_fn = typing.cast(Callable[[], Awaitable[typing.Any]], saved_final_single_fn)
+                    return await no_args_fn()
+
+                final_single_fn = _convert_input_pydantic_empty
+
+                # Reset the descriptor
+                final_single_fn_desc = FunctionDescriptor.from_function(final_single_fn)
+
+            elif (final_single_fn_desc.arg_count > 1):
                 if (input_schema is not None):
                     logger.warning("Using provided input_schema for multi-argument function")
                 else:
@@ -495,7 +510,23 @@ class FunctionInfo:
 
             final_stream_fn_desc = FunctionDescriptor.from_function(final_stream_fn)
 
-            if (final_stream_fn_desc.arg_count > 1):
+            if (final_stream_fn_desc.arg_count == 0):
+                input_schema = input_schema or create_model("EmptyInputSchema")
+
+                saved_final_stream_fn = final_stream_fn
+
+                async def _convert_input_pydantic_empty_stream(
+                        value: input_schema) -> AsyncGenerator[final_stream_fn_desc.output_type]:
+                    no_args_fn = typing.cast(Callable[[], AsyncGenerator[typing.Any]], saved_final_stream_fn)
+                    async for m in no_args_fn():
+                        yield m
+
+                final_stream_fn = _convert_input_pydantic_empty_stream
+
+                # Reset the descriptor
+                final_stream_fn_desc = FunctionDescriptor.from_function(final_stream_fn)
+
+            elif (final_stream_fn_desc.arg_count > 1):
                 if (input_schema is not None):
                     logger.warning("Using provided input_schema for multi-argument function")
                 else:
--- a/packages/nvidia_nat_core/src/nat/builder/function_info.py
+++ b/packages/nvidia_nat_core/src/nat/builder/function_info.py
@@ -411,14 +411,27 @@ class FunctionInfo:
             if (single_fn_desc.output_type != single_to_stream_fn_desc.input_type):
                 raise ValueError("single_to_stream_fn must have the same input type as the output from single_fn")
 
-            async def _converted_stream_fn(
-                    message: single_fn_desc.input_type) -> AsyncGenerator[single_to_stream_fn_desc.output_type]:
-                value = await single_fn(message)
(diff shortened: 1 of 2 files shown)


willkill07:
@mikemikimike CI job is failing: https://github.com/NVIDIA/NeMo-Agent-Toolkit/actions/runs/33283874107/job/99183560414?pr=2191

mikemikimike:
Verified the latest head 5fc01cae: the explicit single/stream conversion paths and implicit Streaming conversion now invoke zero-argument callables without passing a message, and the requested Sequence/docstring/parameter-lookup cleanup is included. Focused builder suite passes (128 tests). Please re-review.

willkill07:
/ok to test [5fc01ca](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2191/commits/5fc01cae8da841cfa0934cc7f733434e8cb15669)

willkill07 (approved):

PR #2201, opened 2026-09-01
Add framework-parity harness + CrewAI example

ayantikanandi18:
This project says it works with many different AI agent frameworks (CrewAI, Semantic Kernel, ADK, and others), but there was no automated way to check that those examples still actually work over time. This adds a nightly check that installs each framework's example fresh and runs it, so we'll know right away if one breaks. It also adds a missing CrewAI example, built to match the existing Semantic Kernel one so they're easy to compare side by side. More details are in scripts/framework_parity/README.md.

## Summary by CodeRabbit

* **New Features**
  * Added a CrewAI-based travel-planning example with itinerary, budgeting, and summary generation.
  * Added hotel pricing and local events tools, including configurable sample data.
  * Added setup documentation covering installation, API keys, execution, workflow details, and expected output.
  * Added the example to the frameworks overview.

* **Tests**
  * Added an integration test for generating a Tokyo travel plan.


.github/workflows/framework_parity.yml
willkill07, line 1: We already have an internal nightly CI that runs all examples. This is superfluous and must be removed.

willkill07 (changes requested):
Please update your PR to remove the entire framework parity harness and CI changes.

examples/frameworks/README.md
willkill07, line 1: examples/ already has a README that lists all of the framework examples. I'd prefer the addition of the new demo to go there.

ayantikanandi18:
Thanks for the review. Both points make sense, and I have pushed the changes.

Removed the framework parity harness and the CI workflow entirely. I did not know about the internal nightly CI that already runs the examples, so the harness was duplicating work that is already covered. I also reverted the .gitignore line it needed.

Removed the new examples/frameworks/README.md and added crewai_demo to the existing examples/README.md instead, in alphabetical order and in the same format as the other entries.

I also took the framework parity wording out of the demo's own README and its config docstring, since the harness it referred to is gone. That was text only, nothing in the workflow, tools, or config changed.

The PR is now just the CrewAI example.

ayantikanandi18:
Branch is now updated with develop, so it is no longer out of date.

One open question I would like your call on. Every other framework example is registered in the root pyproject.toml, both in the examples list and in tool.uv.sources. I tried to add nat_crewai_demo there too, but it breaks dependency resolution. The examples extra would then pull in the crewai extra, and crewai is already declared as conflicting with openpipe-art because of incompatible litellm versions. Making that work looks like it needs a change to the conflicts list, which felt like your decision rather than mine, so I left it out.

So right now the example is not registered in the root pyproject. Happy to add it whichever way you prefer, or to leave it as is if the internal nightly CI picks up examples a different way.

The only failing check is the Label Checker, which needs the category and breaking labels. I do not have permission to add labels to this repo.

willkill07:
/ok to test [4eb43b9](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2201/commits/4eb43b9420ade92ab0c829a270ac1757236a7e39)

PR #2221, opened 2026-09-12
Fix truncate_string exceeding max_length for small limits

### LINKED ISSUE #2220: truncate_string returns a string longer than max_length for small limits
### Describe the bug

`nat.utils.string_utils.truncate_string(text, max_length)` is documented to
"truncate a string to a maximum length," but for `max_length < 3` it returns a
string **longer** than `max_length`. When the input needs truncating it does:

```python
return text[:max_length - 3] + "..."
```

For `max_length < 3`, `max_length - 3` is negative, so `text[:max_length - 3]`
slices from the *end* of the string instead of the start, and the `"..."` is then
appended — producing output that is both longer than `max_length` and cut from
the wrong end.

### Reproduction

```python
from nat.utils.string_utils import truncate_string

truncate_string("abcdefghij", 2)   # -> 'abcdefghi...'  (12 chars for max_length=2)
truncate_string("abcdefghij", 1)   # -> 'abcdefgh...'   (11 chars for max_length=1)
```

Expected: the result never exceeds `max_length` (e.g. `".."`, `"."` — as much of
the ellipsis as fits).

For `max_length >= 3` the function already behaves correctly
(`truncate_string("abcdefghij", 5) == "ab..."`); only small limits are affected.

### Proposed fix

Guard the small-`max_length` case so the ellipsis is truncated to fit instead of
using a negative slice, keeping the existing behavior for `max_length >= 3`.
Happy to send a small PR with tests.

AnishPatel526:
## Description

`nat.utils.string_utils.truncate_string` is documented to "truncate a string to a
maximum length," but for `max_length < 3` it returned a string **longer** than
`max_length`. On the truncating path it did:

```python
return text[:max_length - 3] + "..."
```

When `max_length < 3`, `max_length - 3` is negative, so `text[:max_length - 3]`
slices from the *end* of the input instead of the start, and `"..."` is then
appended — producing output longer than `max_length` and cut from the wrong end:

```python
truncate_string("abcdefghij", 2)   # was 'abcdefghi...' (12 chars for max_length=2)
```

## Fix

Guard the case where there is no room for text plus the ellipsis and return only
as much of the ellipsis as fits (`".."`, `"."`, or `""`), instead of using the
negative slice. Behavior for `max_length >= 3` is unchanged
(`truncate_string("abcdefghij", 5) == "ab..."`), and the returned string now
never exceeds `max_length`.

## Tests

Added `tests/nat/utils/test_string_utils.py` covering `None`, no-truncation,
exact-length, normal truncation, and — as a regression guard — that the result
never exceeds `max_length` for a range of limits including `0`, `1`, `2`. The
regression assertions fail on the current code and pass with this change; `yapf`
and `ruff check` are clean on both files.

- I am familiar with the Contributing Guidelines.
- The commit is signed off under the Developer Certificate of Origin (`git commit -s`).
- Existing and new tests cover these changes.
- No documentation update is required; this is an internal correctness fix with no
  change to any public API or configuration surface.

Closes #2220.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text truncation to ensure results never exceed the specified maximum length.
  * Added safe handling for very small length limits while preserving expected behavior for empty and short inputs.

* **Tests**
  * Added coverage for unchanged inputs, ellipsis truncation, maximum-length enforcement, and small-limit edge cases.

### DIFF
--- a/packages/nvidia_nat_core/src/nat/utils/string_utils.py
+++ b/packages/nvidia_nat_core/src/nat/utils/string_utils.py
@@ -38,6 +38,9 @@ def convert_to_str(value: Any) -> str:
         raise ValueError(f"Unsupported type for conversion to string: {type(value)}")
 
 
+_ELLIPSIS = "..."
+
+
 def truncate_string(text: str | None, max_length: int = 100) -> str | None:
     """
     Truncate a string to a maximum length, adding ellipsis if truncated.
@@ -47,8 +50,16 @@ def truncate_string(text: str | None, max_length: int = 100) -> str | None:
         max_length: Maximum allowed length (default: 100)
 
     Returns:
-        The truncated text with ellipsis if needed, or None if input was None
+        The truncated text with ellipsis if needed, or None if input was None. The
+        returned string never exceeds ``max_length`` characters.
     """
     if not text or len(text) <= max_length:
         return text
-    return text[:max_length - 3] + "..."
+    # When there is no room for text plus the ellipsis, return as much of the
+    # ellipsis as fits. Guarding here avoids the negative-index slice
+    # ``text[:max_length - 3]`` would produce for ``max_length < 3`` (which cut
+    # characters off the end of ``text`` and returned a string longer than
+    # ``max_length``).
+    if max_length < len(_ELLIPSIS):
+        return _ELLIPSIS[:max(max_length, 0)]
+    return text[:max_length - len(_ELLIPSIS)] + _ELLIPSIS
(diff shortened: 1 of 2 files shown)


AnishPatel526:
Hello! Would love to get this reviewed. This is a non-breaking bug fix. Could a maintainer add the bug and non-breaking labels so the Label Checker passes? Thank you!

willkill07:
/ok to test [e031c41](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2221/commits/e031c41b399d6579388f61669c916032e01b36d2)

willkill07 (approved):

willkill07 (approved):

PR #2006, opened 2026-06-05
Add NeMo Guardrails policy middleware

ericevans-nv:
## Description

This PR adds first-class **NeMo Guardrails** support to the NeMo Agent Toolkit and
consolidates all safety/security middleware into the `nvidia-nat-security` plugin.

### New: NeMo Guardrails middleware (`_type: guardrails`)
- Adds `GuardrailsMiddleware`, which applies NeMo Guardrails rails (content safety,
  jailbreak detection, PII masking, output verification, etc.) at a function's
  **input and output boundaries** via the dynamic middleware system.
- **Schema-driven field selection:** rather than JSONPath text extraction, operators
  declare which string fields to guard using dotted attribute paths against the
  function's Pydantic input/output schema (`GuardrailFunctionFields`,
  e.g. `reviews.review`). List paths fan out and are concatenated into a single rail
  call; a field with no subpath guards the whole value. When no selection is given
  (e.g. the `<workflow>` boundary, whose value is a plain string), the entire value
  is guarded.
- Supports **pass / modify / block** outcomes driven by the Colang policy: modified
  text is written back to the input/output, blocks short-circuit the call and return
  the rail's message (`on_post_invoke_blocked` / `PostInvokeBlockedError` are
  overridable).

### Refactor / packaging
- **Moves the defense middleware** (`pii_defense`, `content_safety_guard`,
  `output_verifier`, `pre_tool_verifier`) and the **red-teaming middleware** out of
  `nvidia-nat-core` and into the `nvidia-nat-security` plugin (renames preserve history).
- Reworks extras accordingly:
  - Removes the `pii-defense` extra from `nvidia-nat-core`.
  - Adds `defense` and `guardrails` extras to `nvidia-nat-security` (and surfaces them
    on the umbrella `nvidia-nat` package). `guardrails` pulls in `nemoguardrails`.
- Updates lockfiles across the repo and `ci/.nim_models_used.json` to reflect the
  extras restructuring and the new example's NIM models.

> **Breaking change (migration):** Python imports of the defense/red-teaming
> middleware move from `nat.middleware.*` to
> `nat.plugins.security.middleware.*`, and `nvidia-nat[pii-defense]` is replaced by
> `nvidia-nat[defense]` (provided by `nvidia-nat-security`). The registered `_type`
> values are unchanged. See the updated migration guide.

### Config loading
- YAML config loading now uses `${VAR}`-style environment interpolation
  (`surrounded_vars_only=True`) and interpolates recursively after parse, so
[...]

PR #2242, opened 2026-09-20
fix(core): do not expand ${VAR} references a second time inside env values

### LINKED ISSUE #2241: yaml_loads expands ${VAR} references a second time inside environment variable values
### Version

develop @ 013255a4 (also 1.x releases since #2006)

Source

### Describe the bug.

`yaml_loads` expands `${VAR}` references twice: once on the raw YAML text (`expand(config, surrounded_vars_only=True)`) and again on every parsed string value (`_interpolate_config`). Both passes were added in #2006. The second pass cannot find anything the first one missed, so its only effect is to expand `${...}` that was *inside an environment variable's value*. An env var such as `PROMPT_TEMPLATE='Hello ${name}'`, a password containing `${`, or any template string is therefore rewritten (usually to an empty string, since the inner name is not set) instead of being used verbatim.

```python
import os
from pathlib import Path
from nat.utils.io.yaml_tools import yaml_loads

os.environ["NAT_TEST_TEMPLATE"] = "Hello ${NAME}, token ${TOKEN:-none}"
os.environ["NAME"] = "world"
print(yaml_loads("prompt: ${NAT_TEST_TEMPLATE}\n", Path("."))["prompt"])
```

### Relevant log output

```
Hello world, token none
```

Expected: `Hello ${NAME}, token ${TOKEN:-none}` (the env var's literal value).

### Full env printout

n/a

### Other/Misc.

Fix: drop the second, per-value pass and keep the text-level expansion (which is what existed before #2006). PR incoming.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

anishmehta24:
## Description

`yaml_loads` expanded `${VAR}` references twice: on the raw YAML text (`expand(config, surrounded_vars_only=True)`) and then again on every parsed string value (`_interpolate_config`), both added in #2006. The text-level pass already resolves every reference written in the config, so the only thing the second pass could act on was `${...}` that came from an environment variable's own value: a prompt template like `Hello ${name}`, a secret containing `${`, and so on were rewritten (usually to `Hello `, since the inner name is unset) instead of being used verbatim.

This removes the per-value pass and keeps the text-level expansion, which is the behaviour before #2006. `_resolve_file_references` still runs on the parsed tree as before.

Test: `test_yaml_loads_does_not_expand_variables_inside_env_values` sets env values containing `${...}` and `$$` and checks they come through unchanged while ordinary references still expand. It fails on `develop` (`'Hello world, token none' == 'Hello ${NAT_TEST_NAME}, ...'`). `test_yaml_tools.py`: 35 passed. yapf and ruff clean.

Closes #2241

## Summary by CodeRabbit

* **Bug Fixes**
  * YAML configuration variables are now expanded only once before parsing.
  * Nested `${...}` expressions and dollar signs within environment variable values are preserved correctly.

* **Tests**
  * Added regression coverage for single-pass environment variable expansion.


willkill07:
/ok to test [a0fb28c](https://github.com/NVIDIA/NeMo-Agent-Toolkit/pull/2242/commits/a0fb28cd5e183d6bbc29f366cb33291f65adc10b)

willkill07 (approved):

